PR #5, opened 2023-01-04
Update provider.tf

syedzubeen:


ansel1:
Thanks for the PR.  I'm not merging this because these files are mirrored from another, internal repo, but I have made this change to our internal repo.  It should be published here shortly.

PR #49, opened 2026-03-12
TFIN-86: Fix multi-node cluster creation

praveenkumarchaubey:
- Each joining node now uses its own address in CSR and join requests, not the provider's address
- The original node's host/port is now correctly passed to all join requests
- Added retries when signing a node's CSR fails due to temporary consensus unavailability after a previous node joins

### DIFF
--- a/internal/provider/cm/resource_cm_cluster.go
+++ b/internal/provider/cm/resource_cm_cluster.go
@@ -4,7 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"regexp"
+	"net"
+	"net/url"
 	"strings"
 	"time"
 
@@ -123,7 +124,6 @@ func (r *resourceCMCluster) Create(ctx context.Context, req resource.CreateReque
 
 	// Retrieve values from plan
 	var plan CMClusterTFSDK
-	regexURL := regexp.MustCompile(`https://([a-zA-Z0-9.\-]+)`)
 
 	diags := req.Plan.Get(ctx, &plan)
 	resp.Diagnostics.Append(diags...)
@@ -136,9 +136,11 @@ func (r *resourceCMCluster) Create(ctx context.Context, req resource.CreateReque
 	var originalNodePort int64
 	for _, n := range plan.Nodes {
 		if n.Original.ValueBool() {
-			urlOriginal := regexURL.FindStringSubmatch(n.Host.ValueString())
-			if len(urlOriginal) > 1 {
-				originalNodeHost = urlOriginal[1]
+			var err error
+			originalNodeHost, err = extractHost(n.Host.ValueString())
+			if err != nil {
+				resp.Diagnostics.AddError("Invalid host for original node", err.Error())
+				return
 			}
 			originalNodePort = n.Port.ValueInt64()
 			break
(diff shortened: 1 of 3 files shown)


internal/provider/cm/resource_cm_cluster.go
```diff
@@ -262,13 +300,15 @@ func (r *resourceCMCluster) Create(ctx context.Context, req resource.CreateReque
...
 			payloadJoinNode.LocalNodeHost = urlJoinNode[1]
-			payloadJoinNode.MemberNodeHost = urlMemberNodePub[1]
+			payloadJoinNode.MemberNodeHost = originalNodeHost
+			payloadJoinNode.MemberNodePort = originalNodePort
 
 			if node.Port.ValueInt64() != types.Int64Null().ValueInt64() {
 				payloadJoinNode.LocalNodePort = node.Port.ValueInt64()
 			}
+			urlLocalNodePub := regexURL.FindStringSubmatch(node.PublicAddress.ValueString())
+			payloadJoinNode.LocalNodePublicAddress = urlLocalNodePub[1]
```
michli: it'll be nice to have array size checking for urlLocalNodePub.

michli (approved):
LGTM overall.

internal/provider/cm/resource_cm_cluster.go
```diff
@@ -206,7 +222,7 @@ func (r *resourceCMCluster) Create(ctx context.Context, req resource.CreateReque
 			urlLocalNode := regexURL.FindStringSubmatch(node.Host.ValueString())
 			payloadCSR.LocalNodeHost = urlLocalNode[1]
 
-			urlLocalNodePubAddress := regexURL.FindStringSubmatch(r.client.CipherTrustURL)
+			urlLocalNodePubAddress := regexURL.FindStringSubmatch(node.PublicAddress.ValueString())
```
emanuelsteen: `node.PublicAddress` isn't a URL, right? Do we need the regexURL match?

emanuelsteen: I find it quite confusing, our example doesn't have the `host` not the `public_address` as a URL:

https://github.com/ThalesGroup/terraform-provider-ciphertrust/blob/main/examples/resources/ciphertrust_cluster/resource.tf

The regexURL is set as:
```
regexURL := regexp.MustCompile(`https://([a-zA-Z0-9.\-]+)`)
```

This regexp essentially extracts the hostname from a URL, so the old line such as:
```
urlLocalNodePubAddress := regexURL.FindStringSubmatch(r.client.CipherTrustURL)
```

Is quite confusing as it doesn't really return a URL but a hostname from a URL and the variable shouldn't be named `url*`.

Also, given this regexp it looks like it would fail to extract the hostname if the value isn't an URL, which our example isn't.

internal/provider/cm/resource_cm_cluster.go
praveenkumarchaubey: This has been fixed now.

internal/provider/cm/resource_cm_cluster.go
praveenkumarchaubey: This has been fixed now.

PR #75, opened 2026-03-27
removed the hardcoded key

michli:


nsvora (approved):
Test change looks good to me.
@michli one question - similar change would be required in TestSchedulerDataSource too right? Can you please check why TestSchedulerDataSource test is passing without your change?

PR #77, opened 2026-04-01
TFIN-112: Add ciphertrust_cluster_node resource

praveenkumarchaubey:
Introduces a new resource to add a single node to an existing CM cluster using the three-step CSR workflow
(generate CSR → sign on member → join).


examples/resources/ciphertrust_cluster_node/resource.tf
```diff
@@ -0,0 +1,77 @@
...
+  public_address = "10.10.10.13"
+
+  # member_host is the hostname or FQDN of the existing cluster member (the
+  # provider's configured node). When omitted, the provider's address is used.
+  member_host = "10.10.10.11"
+
+  # member_port is the port on the existing cluster member. Defaults to 5432.
+  member_port = 5432
+
+  credentials = {
```
nsvora, line 34: change credentials to node_provider
add new param - "url"

docs/resources/cluster.md
```diff
@@ -3,130 +3,121 @@
...
 
-Optional:
+## Notes
 
-- `auth_domain` (String) CipherTrust authentication domain of the user. This is the domain where the user was created.
-- `domain` (String) CipherTrust domain to log in to. Default is the empty string (root domain).
-- `no_ssl_verify` (Boolean) Set to false to verify the server's certificate chain and host name.
-- `password` (String) Password for the node
-- `username` (String) Username for the node
+- **Sequential Node Addition**: Additional nodes should be added using `ciphertrust_cluster_node` resources with `depends_on = [ciphertrust_cluster.main]`
```
nsvora: User can provide multiple nodes in configuration but internally terraform provider implementation would add nodes in sequence. Is that right? If so, can we refer to sequential addition in description instead of title (Sequential Node Addition) ?

docs/resources/cluster_node.md
```diff
@@ -0,0 +1,203 @@
...
+1. `DELETE /v1/nodes/{id}` on the cluster member (remove from cluster)
+2. `DELETE /v1/cluster` on the removed node (clear cluster config)
+
+## Import
+
+Cluster node resources cannot be imported.
+
+## Notes
+
+- **Sequential Joins**: The provider uses an internal mutex to ensure only one node joins at a time, even with `for_each` parallel resource creation. Join operations use `blocking=false` with status polling until the node reaches ready state, preventing consensus issues and avoiding timeouts from long-running join operations
```
nsvora: If i recall correctly, you had mentioned there was some reason not to use blocking parameter is that right? If user cannot use blocking parameter, I think we can remove it from here to simplify user experience.

docs/resources/cluster.md
```diff
@@ -3,130 +3,121 @@
...
+  depends_on = [ciphertrust_cluster.main]
+  
+  host           = each.value.host
+  port           = 5432
+  public_address = each.value.host
+  
+  member_host = "10.10.10.11"
+  member_port = 5432
+  
+  credentials = {
```
nsvora, line 65: As mentioned here, wouldn't renaming to node_provider be better?
https://github.com/ThalesGroup/terraform-provider-ciphertrust/pull/77#discussion_r3053760052

internal/provider/cm/resource_cm_cluster.go
```diff
@@ -41,77 +41,50 @@ func (r *resourceCMCluster) Metadata(_ context.Context, req resource.MetadataReq
...
-									Description: "Set to false to verify the server's certificate chain and host name.",
-								},
-							},
-						},
-					},
+			"local_node_host": schema.StringAttribute{
+				Required:    true,
+				Description: "The hostname or IP of this node. Must be reachable by all nodes in the cluster, including this one.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
```
nsvora, line 57: Marking as requiresreplace would mean update to local_node_host would delete and recreate node resource. I thought we cannot edit local_node_host. In that case we should not need requiresReplace here... right?

internal/provider/cm/resource_cm_cluster.go
```diff
@@ -41,77 +41,50 @@ func (r *resourceCMCluster) Metadata(_ context.Context, req resource.MetadataReq
...
+			"public_address": schema.StringAttribute{
+				Optional:    true,
+				Description: "The fully qualified domain name (FQDN) or public IP of this node. This attribute is used by CipherTrust Manager connectors to learn how to access this particular node of the cluster remotely. Can be updated.",
 			},
 			"node_id": schema.StringAttribute{
 				Computed:    true,
-				Description: "This CipherTrust manager node ID",
+				Description: "CipherTrust Manager node ID assigned to this cluster node.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
```
nsvora, line 74: UseStateForUnknown would mean use prior state value - not sure i understand the usage in current context. can you please elaborate?

internal/provider/cm/resource_cm_cluster.go
```diff
@@ -404,14 +241,14 @@ func (r *resourceCMCluster) Delete(ctx context.Context, req resource.DeleteReque
...
-	tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_cluster.go -> Delete]["+state.ID.ValueString()+"]["+output+"]")
+	// DELETE /v1/cluster
+	// Note: This only works if this is the last/only node in the cluster
+	output, err := r.client.DeleteByURL(ctx, state.NodeId.ValueString(), common.URL_CLUSTER_INFO)
+	tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_cm_cluster.go -> Delete]["+state.ID.ValueString()+"]["+output+"]")
 	if err != nil {
 		resp.Diagnostics.AddError(
 			"Error deleting cluster",
-			"Could not cluster, unexpected error: "+err.Error(),
+			"Could not delete cluster. Ensure all additional nodes have been removed first and this is the last node in the cluster. Error: "+err.Error(),
```
nsvora: Isn't this coming from API ? In that case can we avoid hardcoding here?

internal/provider/cm/resource_cm_cluster_node.go
```diff
@@ -0,0 +1,788 @@
...
+	}
+	if err != nil {
+		resp.Diagnostics.AddError("Error signing CSR on the existing cluster member after retries: ", err.Error())
+		return
+	}
+
+	// Step 3: Send the join request to the new node with the signed certificate and CA chain.
+	// Use the explicit member_host if provided, otherwise fall back to the provider's address.
+	memberHostInput := r.client.CipherTrustURL
+	if !plan.MemberHost.IsNull() && !plan.MemberHost.IsUnknown() && plan.MemberHost.ValueString() != "" {
```
nsvora: in which condition can plan.MemberHost.IsUnknown() be true?

internal/provider/cm/resource_cm_cluster_node.go
```diff
@@ -0,0 +1,788 @@
...
+	if state.Creds != nil && !state.Creds.Address.IsNull() && !state.Creds.Address.IsUnknown() {
+		if addr := state.Creds.Address.ValueString(); addr != "" {
+			nodeConnAddr = addr
+		}
+	}
+	nodeURL := nodeConnAddr
+	if !strings.Contains(nodeURL, "://") {
+		nodeURL = "https://" + nodeURL
+	}
+	// The node's auth service can be briefly unavailable after a cluster join even when status="r",
```
nsvora, line 514: I think we discussed about this in standup today . Please check if 30 minutes retry is required for status='r' ?

internal/provider/cm/resource_cm_cluster_node.go
```diff
@@ -0,0 +1,788 @@
...
+		nodeURL = "https://" + nodeURL
+	}
+	nodeClient, err := common.NewClient(ctx, id, &nodeURL, &nodeAuthDomain, &nodeDomain, &nodeUsername, &nodePassword, true, 180)
+	if err != nil {
+		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_cluster_node.go -> Delete]["+id+"]")
+		resp.Diagnostics.AddError("Unable to create HTTPS client for the removed node", err.Error())
+		return
+	}
+
+	// Snapshot member's current node count before removal so we can verify it decrements.
```
nsvora, line 677: if one node is added and other is removed from resource.tf file - which will be done first - addition or removal ? In that case would comparing count here work correctly? Probably it will be fine because both node join and delete are using same mutex - is that right?

internal/provider/resource_cm_cluster_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	return n
+}
+
+// node3Coords returns the third node's coordinates.
+// Skips t if any CLUSTER_NODE3_* variable is unset.
+func node3Coords(t *testing.T) clusterNode {
+	t.Helper()
+	n := clusterNode{
+		host:     os.Getenv("CLUSTER_NODE3_HOST"),
```
nsvora, line 87: where are these environment variables defined?

examples/resources/ciphertrust_cluster_node/resource.tf
praveenkumarchaubey: Done

docs/resources/cluster.md
praveenkumarchaubey: These were already defined under node resource, hence I didn't changed that and kept it as is,
credentials = {
    address  = each.value.address
    username = "test"
    password = each.value.password
  }

docs/resources/cluster.md
praveenkumarchaubey: Done.

docs/resources/cluster_node.md
praveenkumarchaubey: Done

internal/provider/cm/resource_cm_cluster.go
praveenkumarchaubey: I think RequiresReplace() is the correct modifier here. The Update function only handles public_address and nothing else. It will ignore any change made to the local_node_host. Without RequiresReplace(), if a user changes local_node_host in the config, terraform would call update, and would do nothing with it, and apply would succeed giving a false positive result as the node would still be running with the old host. This will be deviation from terraform point of view.
With RequiresReplace(), terraform now catches the change at plan time and tells the user that "this requires destroy + recreate." 
IMO, its user responsibility to configure terraform correctly.

internal/provider/cm/resource_cm_cluster.go
praveenkumarchaubey: It applies to computed attributes only.
Without it, every terraform plan would show node_id = (known only after apply)
Terraform in this case doesn't know if the provider will return a new value or not.
UseStateForUnknown() tells Terraform that "if there's already a value in state, keep it and don't mark it unknown." On first create there is no prior state, so it still correctly shows (known after apply). It only use the previous values on subsequent plans where the value is already known and won't change.

internal/provider/cm/resource_cm_cluster.go
praveenkumarchaubey: Fair point, I will simplify it.

internal/provider/cm/resource_cm_cluster_node.go
praveenkumarchaubey: Good catch. IsUnknown() can't be true here in practice as unknowns only exist at plan time. This check is kind of redundant here. I will remove this.

internal/provider/cm/resource_cm_cluster_node.go
praveenkumarchaubey: Yes, I added the check after that. This is the max retry interval.

internal/provider/cm/resource_cm_cluster_node.go
praveenkumarchaubey: Yes, you are correct, both create and delete acquire the same mutex here. So even if Terraform schedules a concurrent add+remove in the same apply, whichever operation acquires the mutex first runs to completion.
E.g, in a 3 node cluster, if delete goes first, count = 3, removes node2, waits for count = 2, completes. Then Add runs. If Add goes first: completes fully (count = 4). Then Delete runs, count = 4, removes node2, waits for count = 3.
Either way, the count is taken after the mutex is acquired.

internal/provider/resource_cm_cluster_test.go
praveenkumarchaubey: These vars are defined in the framework: terraform-provider-ciphertrust-test

PR #104, opened 2026-04-29
TFIN-147 Fix issues in CTE client resources

chinmay-thales:
**TFIN-147 Fix issues in CTE client resources**

**Changes**
- **profile_identifier** param was not sent properly in the payload json in create function.

**Testing**
- Validated the above change using sample-scripts


chinmay-thales:
@rahulrampura-thales  Please review and merge this change.

PR #106, opened 2026-04-30
TFIN-170 CTE Profile resource test case is failing.

chinmay-thales:
**[TFIN-170] CTE Profile resource test case is failing.**

Changes
- **file_threshold** parameter value has been replaced with a new value.

**Testing**
- Test file is now working correctly.


chinmay-thales:
@rahulrampura-thales  please review and merge above change.

PR #107, opened 2026-04-30
[TFIN 171] Fixed TestCiphertrustCTEClientGroupClientsDataSource TestCase

pratham-thales:
**[TFIN 171] :-  TestCiphertrustCTEClientGroupClientsDataSource TestCase Fixed**

**Summary:-**
The test case was failing because it was not able create CTE client as it was not able to attach DefaultClientProfile to it.

**Changes:-**

- Created a new Profile through its resource.
- Updated the test to use newly created profile instead of relying on default profile.

**Testing:-**

- Verified the changes by running the test case successfully.


pratham-thales:
@rahulrampura-thales , Please review this.

PR #118, opened 2026-05-29
TFIN-203  CTE | Drift Detection (Policy and Policy-rules)

chinmay-thales:
**TFIN-203  CTE | Drift Detection (Policy and Policy-rules)**

**Changes**
- Added Drift detection functionality for CTE Policy and Policy-rules resources.


chinmay-thales:

PR #119, opened 2026-05-29
TFIN-203 CTE | Drift Detection (CTE Profile)

chinmay-thales:
**TFIN-203 CTE | Drift Detection (CTE Profile)**

**Changes**
- Added drift detection functionality for CTE Profile resource

**Testing**

### DIFF
--- a/internal/provider/cte/resource_cte_profile.go
+++ b/internal/provider/cte/resource_cte_profile.go
@@ -10,8 +10,15 @@ import (
 
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
 	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/booldefault"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64default"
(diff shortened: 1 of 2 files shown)


chinmay-thales:

internal/provider/cte/resource_cte_profile.go
```diff
@@ -657,21 +864,36 @@ func (r *resourceCTEProfile) Read(ctx context.Context, req resource.ReadRequest,
 	if resp.Diagnostics.HasError() {
 		return
 	}
-	_, err := r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_CTE_PROFILE)
+
+	response, err := r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_CTE_PROFILE)
```
rahulrampura-thales, line 868: What if err is returned here? Do we need to handle any error case here? If yes, please add the appropriate error handling

internal/provider/cte/resource_cte_profile.go
```diff
@@ -758,6 +980,7 @@ func (r *resourceCTEProfile) Update(ctx context.Context, req resource.UpdateRequ
 		payload.LDTQOSStatusCheckRate = plan.LDTQOSStatusCheckRate.ValueInt64()
 	}
 	// Set client_logger_configs in the request
+	// Set client_logger_configs in the request
```
rahulrampura-thales: duplicate line of line no. 982. Remove it

internal/provider/cte/resource_cte_profile.go
chinmay-thales: This API returns empty string as response in case of invalid ID which is already handled.

internal/provider/cte/resource_cte_profile.go
chinmay-thales: Done

PR #120, opened 2026-05-29
TFIN-203 CTE | Drift Detection (LDT Comms Group) 

chinmay-thales:
**TFIN-203 CTE | Drift Detection (LDT Comms Group)** 

**Changes**
- Added drift detection functionality for LDT Comms Group resource

**Testing**

### DIFF
--- a/internal/provider/cte/resource_cte_ldtgroupcomms.go
+++ b/internal/provider/cte/resource_cte_ldtgroupcomms.go
@@ -9,9 +9,11 @@ import (
 	// "github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	// "github.com/hashicorp/terraform-plugin-framework/schema/validator"
 
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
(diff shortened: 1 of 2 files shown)


chinmay-thales:
@rahulrampura-thales Please review the changes

internal/provider/cte/resource_cte_ldtgroupcomms.go
```diff
@@ -135,21 +142,61 @@ func (r *resourceLDTGroupCommSvc) Read(ctx context.Context, req resource.ReadReq
 	if resp.Diagnostics.HasError() {
 		return
 	}
-	_, err := r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_LDT_GROUP_COMM_SVC)
+
+	// Fetch LDT group details
+	response, err := r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_LDT_GROUP_COMM_SVC)
```
rahulrampura-thales, line 147: Can this return error ? Add error handling if required OR add check for success case like err !=nil to proceed

chinmay-thales: No, In case of invalid ID it will return an empty response, which is already handled.

PR #135, opened 2026-06-03
TFIN-285: Add new value named "ml-dsa" in the "algorithm" field in ciphertrust_cm_key resource 

meetwithabhishek:
## Summary

Automated implementation of TFIN-285: Add new value named "ml-dsa" in the "algorithm" field in ciphertrust_cm_key resource 

## Implementation plan

# Implementation Plan: TFIN-285 — Add "ml-dsa" algorithm to ciphertrust_cm_key

## Objective
Extend the `ciphertrust_cm_key` resource so that the `algorithm` attribute accepts a new value `"ml-dsa"` (Module-Lattice Digital Signature Algorithm, a post-quantum signature scheme), in addition to the currently supported values (`aes`, `rsa`, `ec`, `hmac-sha1`, `hmac-sha256`, `hmac-sha384`, `hmac-sha512`).

## Pattern to Follow
Use **`internal/provider/cckm/aws/resource_aws_key.go`** as the reference for how an algorithm-style enum attribute is declared in schema, validated, and passed through to the CM API payload. For cm-key–specific layout (which lives under `internal/provider/cm/` — locate the file backing the `ciphertrust_cm_key` resource, typically `resource_cm_key.go` and its companion `models_cm_key.go` / `schema_cm_key.go`), mirror the existing structure for `algorithm`.

### 1. `internal/provider/cm/resource_cm_key.go` (or equivalent resource file backing `ciphertrust_cm_key`)
- Locate the `Schema()` method (or the separate schema file) where the `algorithm` attribute is defined as a `schema.StringAttribute`.
- Update the validator: if `stringvalidator.OneOf(...)` is used, add `"ml-dsa"` to the allowed list.
- Update the `MarkdownDescription` / `Description` for the `algorithm` attribute to document the new accepted value and note it is for post-quantum signing keys.
- Confirm `algorithm` retains `RequiresReplace` semantics (algorithm is immutable on a CM key — changing it must force recreation). Do NOT relax this.
- Do **not** modify the `Read()` method (per TFIN-174 constraint).

### 2. `internal/provider/cm/models_cm_key.go` (or wherever the request/response structs live)
- No struct field changes are required: `algorithm` is already a `string` matching the `PostKey.algorithm` field in the CM API. Verify no client-side switch statement on algorithm exists that would reject `ml-dsa`; if one does (e.g., to decide whether `curveid` or `size` is required), extend it to recognize `ml-dsa`.

### 3. Validation / conditional-field logic
- If the resource currently enforces conditional requirements (e.g., `size` required for `aes`/`rsa`, `curveid` required for `ec`), decide how `ml-dsa` parameterization works. ML-DSA variants are typically selected via a parameter set (e.g., ML-DSA-44/65/87). If CM accepts this via `size` or another existing field, document it in the attribute description; otherwise treat `size`/`curveid` as not required for `ml-dsa`. Update any `ConfigValidators` or `ValidateConfig` logic accordingly.

### 4. `internal/provider/cm/data_source_cm_key.go` (if it exists and exposes `algorithm`)
- Mirror the description update so the data source documents `ml-dsa` as a possible returned value. No validator change is needed for read-only computed attributes.

### 1. `examples/resources/ciphertrust_cm_key/ml_dsa.tf`
- Add a small example HCL snippet showing a `ciphertrust_cm_key` resource declared with `algorithm = "ml-dsa"`.

### 2. Changelog / docs
- Add a `.changes/` or `CHANGELOG.md` entry (follow the repo's existing convention) noting: *“ciphertrust_cm_key: `algorithm` now accepts `ml-dsa`.”*
- Regenerate or hand-edit `docs/resources/cm_key.md` to list `ml-dsa` in the `algorithm` attribute description.

## New Schema Attributes
None. This ticket only widens the accepted enum for an existing attribute.

## Breaking Constraints
- `algorithm` must remain **Required** and continue to carry **RequiresReplace** — switching a key's algorithm in-place is not supported by CM.
- Do not implement or expand `Read()` (per TFIN-174).
- Preserve backward compatibility: existing configurations using the seven prior algorithm values must continue to plan/apply unchanged.

## Test Notes
- Add an acceptance test case (or unit-level schema validation test) covering `algorithm = "ml-dsa"` accepted, and an invalid value (`"ml-dsax"`) rejected by the validator. Follow the test pattern used alongside `resource_aws_key.go`.

---

### DIFF
--- a/internal/provider/cm/resource_cm_key.go
+++ b/internal/provider/cm/resource_cm_key.go
@@ -59,7 +59,7 @@ func (r *resourceCMKey) Schema(_ context.Context, _ resource.SchemaRequest, resp
 			},
 			"algorithm": schema.StringAttribute{
 				Optional:    true,
-				Description: "Cryptographic algorithm this key is used with. Defaults to 'aes'",
+				Description: "Cryptographic algorithm this key is used with. Defaults to 'aes'. Use 'ml-dsa' for post-quantum signing keys (Module-Lattice Digital Signature Algorithm).",
 				Validators: []validator.String{
 					stringvalidator.OneOf([]string{"aes",
 						"tdes",
@@ -72,6 +72,7 @@ func (r *resourceCMKey) Schema(_ context.Context, _ resource.SchemaRequest, resp
 						"seed",
 						"aria",
 						"opaque",
+						"ml-dsa",
 						"AES", "EC", "RSA"}...),
 				},
 			},
--- a/docs/resources/cm_key.md
+++ b/docs/resources/cm_key.md
@@ -112,7 +112,7 @@ output "key_name" {
 ### Optional
 
 - `activation_date` (String) Date/time the object becomes active
-- `algorithm` (String) Cryptographic algorithm this key is used with. Defaults to 'aes'
+- `algorithm` (String) Cryptographic algorithm this key is used with. Defaults to 'aes'. Use 'ml-dsa' for post-quantum signing keys (Module-Lattice Digital Signature Algorithm).
 - `aliases` (Attributes List) Aliases associated with the key. The alias and alias-type must be specified. The alias index is assigned by this operation, and need not be specified. (see [below for nested schema](#nestedatt--aliases))
 - `all_versions` (Boolean)
 - `archive_date` (String) Date/time the object becomes archived
--- a/examples/resources/ciphertrust_cm_key/ml_dsa.tf
+++ b/examples/resources/ciphertrust_cm_key/ml_dsa.tf
@@ -1,0 +1,45 @@
+# Terraform Configuration for CipherTrust Provider
+
+# This configuration demonstrates the creation of a post-quantum signing key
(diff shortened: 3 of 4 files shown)


internal/provider/cm/resource_cm_key.go
```diff
@@ -72,6 +72,7 @@ func (r *resourceCMKey) Schema(_ context.Context, _ resource.SchemaRequest, resp
 						"seed",
 						"aria",
 						"opaque",
+						"ml-dsa",
```
ashwani573, line 75: test case is missing , Please add the test case as well

ashwani573 (changes requested):

internal/provider/cm/resource_cm_key.go
meetwithabhishek: Fixed in `8148961`: TFIN-285: address review feedback (round 1)

PR #136, opened 2026-06-03

zubeensyed:
## Summary

- Fixes swapped JSON struct tags on the `CMKeyJSON` payload for the `ciphertrust_cm_key` resource in `internal/provider/cm/schema_cm.go`: `RevocationReason` now serializes to `json:"revocationReason,omitempty"` and `RevocationMessage` to `json:"revocationMessage,omitempty"` (they were crossed).
- Adds `internal/provider/cm/schema_cm_test.go` with unit tests that lock the marshal/unmarshal contract for these fields, guarding against regression.
- Adds a `## Fixes` entry under the existing `# 1.0.1` section in `changelog.md`, including the re-apply note for affected users.

This is a bug fix to an existing resource. No new resource, data source, schema attribute, URL constant, client helper, or provider registration is introduced.

## Motivation

Implements TFIN-286: `ciphertrust_cm_key` `revocation_reason` and `revocation_message` are serialized with swapped JSON tags causing incorrect server-side values.

In `internal/provider/cm/schema_cm.go`, the `CMKeyJSON` struct (the request payload marshaled to the CM Key REST API, consumed by `internal/provider/cm/resource_cm_key.go`) had the `json:"..."` struct tags for `RevocationReason` and `RevocationMessage` swapped:

- `RevocationReason` carried `json:"revocationMessage,omitempty"`.
- `RevocationMessage` carried `json:"revocationReason,omitempty"`.

As a result, the value a user typed into Terraform's `revocation_reason` was persisted on CipherTrust Manager (CM) under `revocationMessage`, and vice versa, on every Create/Update. Because every `Read()` is intentionally empty (provider-wide TFIN-174 constraint), this never surfaced as Terraform drift, but the server-side data was wrong immediately and visible in the CM UI/API. The fix is a tag swap only — no semantic logic changes.

> **Note:** The internal implementation plan and Jira ticket are referenced in the commit message. This description is self-contained for external reviewers.

- Swapped the two `json:"..."` struct tags on the `CMKeyJSON` struct:
  - `RevocationReason` → `json:"revocationReason,omitempty"`
  - `RevocationMessage` → `json:"revocationMessage,omitempty"`
- Go field names, field order, and `omitempty` are preserved. No other field, struct, or schema attribute is touched.

- Package `cm`, standard library `testing` only (no `terraform-plugin-testing` harness, no live CM required).
- `TestCMKeyJSON_RevocationTagsMarshaling` with two sub-tests:
  - "values marshal under matching keys" — constructs a `CMKeyJSON` with `RevocationReason = "REASON_SENTINEL"` and `RevocationMessage = "MESSAGE_SENTINEL"`, marshals via `encoding/json`, and asserts the output contains `"revocationReason":"REASON_SENTINEL"` and `"revocationMessage":"MESSAGE_SENTINEL"`.
  - "omitempty preserved when both fields empty" — marshals an empty `CMKeyJSON{}` and asserts neither `revocationReason` nor `revocationMessage` appears.
- `TestCMKeyJSON_RevocationFieldsUnmarshal` — unmarshals `{"revocationReason":"compromise","revocationMessage":"key exposed in logs"}` and asserts each Go field receives the matching value (symmetric check).

- Adds a `## Fixes` subsection under the existing `# 1.0.1` section, documenting the fix for `ciphertrust_cm_key` and the user-facing re-apply requirement.

### Verified, not changed — `internal/provider/cm/resource_cm_key.go`
- The Create (around line 770) and Update (around line 1208) plan→payload mappings already correctly assign `plan.RevocationReason → payload.RevocationReason` and `plan.RevocationMessage → payload.RevocationMessage` using the correct Go field names. The bug was purely in the JSON tags, so this file required no change.

> **Endpoint / swagger note:** This change touches only Go struct serialization tags; it introduces no new `URL_*` constant and no new client helper call, so there is no new endpoint to verify against the swagger spec. The corrected tags use the camelCase keys (`revocationReason`, `revocationMessage`) that this struct already exchanges with CM and that match the Terraform attribute names.

## Schema attributes

No schema attributes were added, removed, or modified. The Terraform attributes `revocation_reason` and `revocation_message` (both `types.String`, Optional) already existed and are unchanged. No `RequiresReplace` change.

The `Read()` method for `ciphertrust_cm_key` is unchanged and still returns immediately without calling the CM API. This is the deliberate, provider-wide TFIN-174 pattern, and this ticket explicitly does not ask for `Read()` to be implemented.

**User-visible consequence:** After `terraform apply`, subsequent `terraform plan` runs will not show drift for this resource even if it is changed directly in CipherTrust Manager. Implementing `Read()` is tracked separately under TFIN-174 and is out of scope here; doing it as a side effect would surface drift on unrelated fields before the immutable-field fixes land.

**Migration impact:** This correction is not retroactive. Previously-applied keys that set either `revocation_reason` or `revocation_message` retain the inverted values on CM until re-applied. Affected users must re-apply the resource so CM records the correct values. This is called out in `changelog.md`.

## How to test

- [x] `make build` — zero errors (verified passing).
- [x] `make test` — unit tests pass (verified; the new `cm` tests pass, acceptance tests skip without `TF_ACC`). Run the regression guard directly with:
  ```
  go test ./internal/provider/cm/ -run TestCMKeyJSON -v
  ```
  Scenarios covered:
  - **Marshal** — a `CMKeyJSON` with both revocation fields set serializes each value under its matching JSON key.
  - **omitempty** — an empty `CMKeyJSON` emits neither revocation key.
  - **Unmarshal** — the `revocationReason`/`revocationMessage` JSON keys deserialize into the matching Go fields.
- ⚠️ **No acceptance test included.** The plan proposed `TestAccResourceCMKey_RevocationFields` in `internal/provider/cm/resource_cm_key_test.go`, but that file does not exist in the repo and there is no existing `TestAccResourceCMKey_basic` out-of-band HTTP verification pattern to model it on, so it was not added. A real acceptance test would require a live CM and an out-of-band CM API GET to assert server state, since `Read()` is empty (TFIN-174) and cannot surface the bug via state comparison. The unit tests above lock the serialization contract and are sufficient to prevent regression of this specific bug. Reviewers should decide whether to require an acceptance test before merging.

## Checklist

- [ ] Diff is the two-tag swap + new test file + changelog entry only.
- [ ] `resource_cm_key.go` plan→payload mapping verified already correct; not changed.
- [ ] `Read()` left empty per TFIN-174 — not modified.
- [ ] URL constants / client helpers unchanged — no inlined endpoint strings introduced.
- [ ] `changelog.md` updated with user re-apply guidance.
- [ ] No hand-edited files under `docs/` (no schema change, so no doc regeneration expected).

---


CLAUDE.md
```diff
@@ -1,35 +1,306 @@
-# terraform-provider-ciphertrust
```
meetwithabhishek, line 1: remove claude.md file

PR #154, opened 2026-06-05

ashwani573:
## Summary

- Implements real `Read()` for three CM core resources (`ciphertrust_cm_key`, `ciphertrust_groups`, `ciphertrust_cm_reg_token`) so out-of-band deletions and attribute drift are detected on `terraform plan`/refresh.
- Adds `internal/provider/common/errors.go` with `IsNotFoundError(err error) bool`, which matches the CM client's `"status: 404"` error string and drives out-of-band-deletion detection (404 → `resp.State.RemoveResource(ctx)`).
- Documents `Read()` as an intentional state passthrough for the two bootstrap-client resources (`ciphertrust_cm_ssh_key`, `ciphertrust_cm_user_password_change`), where a remote refresh is not feasible.
- Adds acceptance tests for out-of-band delete (key + group) and out-of-band label drift (key).

## Motivation

Implements TFIN-293: `terraform plan` silently shows "no changes" when CM resources are deleted or modified out-of-band, because the affected resources had empty/stub `Read()` implementations and never reconciled Terraform state against CipherTrust Manager.

This ticket explicitly supersedes the provider-wide TFIN-174 "keep `Read()` empty" constraint for the resources listed here: the ticket's expected result mandates refresh-time drift and deletion detection, which requires a GET against CM.

> **Note:** The internal Jira ticket and approved implementation plan are referenced in the commit message. This description is self-contained for external reviewers.

### New file — `internal/provider/common/errors.go`
- Adds `IsNotFoundError(err error) bool`. The CM client formats request failures as `status: %d, body: %s`, so a 404 always yields a string containing `status: 404`. The helper returns `true` on that substring (and `false` for a nil error). Used by every new `Read()` to distinguish out-of-band deletion from real errors.

- Implements `Read()`: loads prior state, calls `r.client.GetById(ctx, id, id, common.URL_KEY_MANAGEMENT)` (`GET api/v1/vault/keys2/{id}`), unmarshals into `CMKeyJSON`.
- 404 → `resp.State.RemoveResource(ctx)` and returns without diagnostics. Other errors → error diagnostic.
- Refreshes `name`, `key_size`, `usage_mask`, `state`, `description`, `uuid`, `muid`, and the `labels` map from the CM response.
- Each scalar is only overwritten when the attribute was already non-null in prior state, so unset `Optional` attributes are not populated into state (which would diff against a config that never specified them).
- Write-only / secret fields (`material`, `wrap_*`, `password`, `id_size`, `template_id`, etc.) are carried forward from prior state — CM never returns them.

- Implements `Read()`: CM groups are keyed by name, so the resource `id` holds the group name. Calls `r.client.GetById(ctx, id, id, common.URL_GROUP)` (`GET api/v1/usermgmt/groups/{name}`), unmarshals into `CMGroupJSON`.
- 404 → `RemoveResource`. Other errors → error diagnostic.
- Refreshes `name` and `id` (both the group name) and refreshes `description` only when set in prior state.
- `app_metadata` / `client_metadata` / `user_metadata` are carried forward from prior state: the schema models them as nested maps, and re-hydrating the loosely typed CM response risks type mismatches on `State.Set`.

- Implements `Read()`: calls `r.client.GetById(ctx, id, id, common.URL_REG_TOKEN)` (`GET api/v1/client-management/regtokens/{id}`), unmarshals into `CMRegTokenJSON`.
- 404 → `RemoveResource`. Other errors → error diagnostic.
- Refreshes `id`, and `ca_id`, `cert_duration`, `client_management_profile_id`, `lifetime`, `max_clients`, `name_prefix` (each guarded by prior non-null state), plus the `label` and `labels` maps when set.
- `token` is a computed secret: kept from prior state unless CM returns a non-empty value.
- Adds a small `stringMapValue` helper to convert a loosely typed CM response map into a `types.Map` of strings.

- `Read()` is now a documented state passthrough (reads prior state, writes it back). This resource is served by the unauthenticated bootstrap client (`*common.CMClientBootstrap`), which exposes no GET method; the CM SSH-key API supports neither GET-by-id nor delete. A remote refresh is therefore not possible. `id` and `key` are write-once on create.

- `Read()` is now a documented state passthrough. This is a write-only action resource: it performs a password change via the bootstrap client and persists no server-side object with a stable identity. Its attributes (`username`, `password`, `new_password`) are all write-only credentials, so there is nothing to refresh against CM.

### Modified files — tests (`internal/provider/resource_cm_key_test.go`, `internal/provider/resource_cm_group_test.go`)
- Added acceptance tests in package `provider`, alongside the existing test files (next to `provider.go`, not in the subsystem subfolder). See "How to test".

**Endpoint verification:** the three GET endpoints map directly to URL constants already present in `internal/provider/common/urls.go` — `URL_KEY_MANAGEMENT = "api/v1/vault/keys2"`, `URL_GROUP = "api/v1/usermgmt/groups"`, `URL_REG_TOKEN = "api/v1/client-management/regtokens"`. No new URL constants were added. (Note: the CM swagger spec was not available in this workspace — `CM_SWAGGER_URL` is not configured — so endpoints were verified against the existing in-repo URL constants and the resources' own Create/Delete usage rather than against the swagger JSON.)

## Read() status

`Read()` is now **implemented** for `ciphertrust_cm_key`, `ciphertrust_groups`, and `ciphertrust_cm_reg_token` (previously empty/stub).

**Fields read from CM response → Terraform state:**
- `ciphertrust_cm_key`: `name`, `key_size`, `usage_mask`, `state`, `description`, `uuid`, `muid`, `labels`.
- `ciphertrust_groups`: `name`, `id` (= name), `description`.
- `ciphertrust_cm_reg_token`: `id`, `ca_id`, `cert_duration`, `client_management_profile_id`, `lifetime`, `max_clients`, `name_prefix`, `label`, `labels`; `token` carried forward when not returned.

**404 handling:** a 404 response from CM removes the resource from Terraform state (`resp.State.RemoveResource(ctx)`) so the next plan proposes recreation. This is the intended deletion-detection behaviour for this ticket.

**Null / absent fields:** unset `Optional` scalar attributes are left untouched when they were null in prior state, to avoid spurious diffs against a config that never specified them. Map attributes are only hydrated when previously set. Write-only / secret fields that CM never returns are carried forward from prior state.

`ciphertrust_cm_ssh_key` and `ciphertrust_cm_user_password_change` remain a documented passthrough (see "What changed") because their backing bootstrap client has no GET method and the underlying APIs expose nothing re-readable.

## Deviations from the approved plan (for reviewers)

- **No new Computed schema attributes on `ciphertrust_cm_key`.** The plan suggested possibly adding `created_at`, `updated_at`, `sha1_fingerprint`, `sha256_fingerprint`, `public_key`, `key_check_value`. They are absent from the existing struct/schema; adding them is unnecessary for deletion/drift detection and would require `docs/` regeneration. Treated as a deliberate scope decision.
- **Plan-listed fields that do not exist on the structs were dropped:** group `users` / `created_at` / `updated_at`; reg-token `valid_until` / `device_id_template` / `created_at` / `updated_at`. Only fields that exist on the framework structs are refreshed.
- **`ciphertrust_cm_ssh_key` and `ciphertrust_cm_user_password_change` are passthroughs, not remote refreshes.** The plan's own embedded research flagged these two as not implementable as written (no GET on the bootstrap client; no GET-by-id/delete on the SSH-key API; no persistent identity for the password-change action). Implemented as documented passthrough with explanatory comments.
- ⚠️ Not included in this PR: the plan also referenced new test files under `internal/provider/cm/`. Tests were instead added to the existing package-`provider` files `resource_cm_key_test.go` and `resource_cm_group_test.go`, per the convention that acceptance tests sit next to `provider.go`. No reg-token acceptance test was added.

## How to test

> ⚠️ **Build/test note:** `make build` and `make test` could not be run in the local dev environment because its Go toolchain is 1.18 while the module requires `go 1.22.0` (see `go.mod`). The CI pipeline (Go 1.22+) compiles and runs them — please rely on CI for build/lint/unit results.

- [ ] Acceptance tests (require a live CM — set `CIPHERTRUST_ADDRESS` and related `CIPHERTRUST_*` env vars; they `t.Skip` otherwise and run in the GitLab pipeline):
  ```
  go test ./internal/provider/... -run 'TestAccCipherTrustCMKey_outOfBand|TestAccCipherTrustCMGroup_outOfBandDelete' -v
  ```
  Scenarios covered:
  - **Out-of-band delete (key)** — `TestAccCipherTrustCMKey_outOfBandDelete`: apply key, delete it via the CM REST client in `PreConfig`, `RefreshState` with `ExpectNonEmptyPlan` (resource dropped from state → recreate proposed), then re-apply recovers it.
  - **Out-of-band drift (key)** — `TestAccCipherTrustCMKey_outOfBandDrift`: apply key with `labels`, mutate the label via REST `UpdateData` in `PreConfig`, `RefreshState` asserts the drifted label is reflected and `name` preserved, with `ExpectNonEmptyPlan`.
  - **Out-of-band delete (group)** — `TestAccCipherTrustCMGroup_outOfBandDelete`: apply group, delete via REST in `PreConfig`, `RefreshState` with `ExpectNonEmptyPlan`, then re-apply recovers it.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every modified `Read()`: `[<file>.go -> Read][id]` pattern.
- [ ] URL endpoints come from constants in `urls.go` — no inlined string literals.
- [ ] `Read()` 404 keeps deletion detection correct — `RemoveResource(ctx)` only on a confirmed 404, error diagnostics on all other failures.
- [ ] Write-only / secret fields carried forward from prior state — not clobbered by the GET response.
- [ ] No new Computed schema attributes added — no `docs/` regeneration required.
- [ ] No hand-edited files under `docs/`.

---


internal/provider/cm/resource_cm_key.go
```diff
@@ -1076,6 +1076,94 @@ func (r *resourceCMKey) Create(ctx context.Context, req resource.CreateRequest,
...
+	var key CMKeyJSON
+	if err := json.Unmarshal([]byte(response), &key); err != nil {
+		resp.Diagnostics.AddError(
+			"Error reading CipherTrust Key",
+			"Could not parse key response: "+err.Error(),
+		)
+		return
+	}
+
+	// Refresh server-authoritative fields onto state so out-of-band drift is
```
rohitpandita-hub, line 1114: Did you made sure all parameters are catered

rohitpandita-hub (changes requested):

internal/provider/cm/resource_cm_key.go
ashwani573: Fixed in `59a7cdc`: TFIN-293: address review feedback (round 1)

PR #155, opened 2026-06-05

zubeensyed:
## Summary

- Fixes swapped JSON serialization tags on the `CMKeyJSON.RevocationReason` and `CMKeyJSON.RevocationMessage` struct fields in `internal/provider/cm/schema_cm.go`, so each field now marshals under its correct CipherTrust Manager key.
- Adds a unit test (`internal/provider/cm/schema_cm_test.go`) asserting the corrected wire format — the first `_test.go` in the `cm` package.
- Extends the existing `ciphertrust_cm_key` acceptance test with a Create-only step that sets `revocation_reason` and `revocation_message` and asserts they round-trip in state.
- Adds a `## Bug Fixes` entry to `changelog.md`.

## Motivation

Implements [TFIN-286]: `ciphertrust_cm_key` `revocation_reason` and `revocation_message` are serialized with swapped JSON tags causing perpetual drift.

The `CMKeyJSON` struct used by the `ciphertrust_cm_key` resource carried transposed `json:"..."` tags on its two revocation fields:

```go
RevocationReason  string `json:"revocationMessage,omitempty"`
RevocationMessage string `json:"revocationReason,omitempty"`
```

The Go-side assignments in `resource_cm_key.go` were already correct (`payload.RevocationReason` is set from the `revocation_reason` attribute and `payload.RevocationMessage` from `revocation_message`). Because the tags were swapped, the user-supplied reason was sent to CipherTrust Manager under the `revocationMessage` key and vice versa. CM therefore persisted the two values transposed, producing incorrect audit data and — once `Read()` is implemented (out of scope per TFIN-174) — perpetual drift.

- Swapped the two JSON tags back so field name and JSON key align:
  - `RevocationReason` → `json:"revocationReason,omitempty"`
  - `RevocationMessage` → `json:"revocationMessage,omitempty"`
- No other change: no field added/removed, no Go type change, `,omitempty` preserved.

> **Correction vs. the approved plan:** the plan stated the CM revocation payload uses the bare keys `reason` / `message`. The actual `CMKeyJSON` struct and all of its sibling fields use camelCase long keys, and the rest of the key payload is camelCase. The fix therefore uses `revocationReason` / `revocationMessage` (matching the surrounding struct convention), **not** `reason` / `message`. This is the only deviation from the plan text and was made to keep the wire format consistent with the existing struct. (CM swagger spec was not available in this workspace — `CM_SWAGGER_URL` is not configured — so the camelCase keys could not be cross-checked against swagger; they were confirmed against the surrounding `CMKeyJSON` field conventions instead.)

- Adds `TestCMKeyJSON_RevocationFieldsJSONTags`: constructs a `CMKeyJSON` with `RevocationReason = "REASON_SENTINEL"` and `RevocationMessage = "MESSAGE_SENTINEL"`, marshals it, and asserts the output:
  - **contains** `"revocationReason":"REASON_SENTINEL"` and `"revocationMessage":"MESSAGE_SENTINEL"`
  - **does not contain** the swapped variants `"revocationReason":"MESSAGE_SENTINEL"` or `"revocationMessage":"REASON_SENTINEL"`
- This is the authoritative regression guard — it runs without a live CM.

- Adds one Create-only step to the existing `TestResourceCMKey` acceptance test that applies a `ciphertrust_cm_key` with `revocation_reason = "Unspecified"` and `revocation_message = "compromised"`, and asserts via `resource.TestCheckResourceAttr` that both attributes round-trip in state. (Test file sits next to `provider.go`, per subsystems.md.)

- Adds a `## Bug Fixes` section noting `ciphertrust_cm_key` no longer swaps `revocation_reason` and `revocation_message` in the CM request body (TFIN-286).

## Schema attributes

No schema changes. The user-facing attributes are unchanged by this PR and listed here for reviewer context only:

|---|---|---|---|
| `revocation_reason` | `types.String` | Optional | Unchanged |
| `revocation_message` | `types.String` | Optional | Unchanged; `Description: "Message explaining revocation."` |

The `Read()` method of `ciphertrust_cm_key` returns without calling the CM API and is **not** modified by this PR. This is a deliberate, provider-wide pattern (TFIN-174): state reconciliation has not yet been implemented for this resource, and the ticket did not ask for it.

**User-visible consequence:** after `terraform apply`, subsequent `terraform plan` runs will show no changes for this resource even if it was modified or deleted directly in CipherTrust Manager. This pre-existing limitation is tracked under TFIN-174 and is unaffected by this fix. Because `Read()` is empty, the corrected wire format cannot be proven by drift detection — the unit test in `schema_cm_test.go` is the authoritative proof, and the acceptance step only confirms the Create path wires the attributes.

## How to test

- [ ] `make test` — unit tests pass (includes the new `TestCMKeyJSON_RevocationFieldsJSONTags`).
  ```
  go test ./internal/provider/... -run TestResourceCMKey -v
  ```
  Scenarios covered:
  - **Create** — apply a key with `revocation_reason = "Unspecified"` and `revocation_message = "compromised"`; assert both attributes round-trip in state.
  - **Destroy** — handled automatically by the test framework `TestCase`.

## Checklist

- [ ] CM API endpoint/wire keys verified against swagger spec — **could not be done**: `CM_SWAGGER_URL` is not configured in this workspace. Reviewers with swagger access should confirm the CM key payload uses `revocationReason` / `revocationMessage` (camelCase). The keys were validated against the surrounding `CMKeyJSON` field conventions.
- [ ] JSON tags align with field names — verified by the new unit test.
- [ ] No `RequiresReplace` / state migration introduced — state already stored values under the correct attribute names; only the outbound HTTP body was wrong. This fix is forward-only (keys revoked before this fix retain their transposed server-side values).
- [ ] `Read()` left intentionally empty (per TFIN-174) — no `resp.State.RemoveResource(ctx)` and no API call added.
- [ ] No hand-edited files under `docs/` — schema unchanged, no regeneration needed.
- [ ] The untracked `terraform-provider-ciphertrust` binary in the working tree is a `make build` artifact and must not be committed.

---

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,8 @@
 # 1.0.1
 
+## Bug Fixes
+    ciphertrust_cm_key no longer swaps revocation_reason and revocation_message in the CipherTrust Manager request body (TFIN-286)
+
 ## New Resources
     ciphertrust_aws_acl
     ciphertrust_aws_key_import_material
--- a/changelog.md
+++ b/changelog.md
@@ -1,8 +1,5 @@
 # 1.0.1
 
-## Bug Fixes
-    ciphertrust_cm_key no longer swaps revocation_reason and revocation_message in the CipherTrust Manager request body (TFIN-286)
-
 ## New Resources
     ciphertrust_aws_acl
     ciphertrust_aws_key_import_material
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 # Added by goreleaser init:
 dist/
 examples/resources/ciphertrust_cm_key/terraform.tfstate.backup
+terraform-provider-ciphertrust
 terraform-provider-ciphertrust.exe
 definition-beta.json
 .claude/swagger/areas/
--- a/internal/provider/cm/schema_cm.go
+++ b/internal/provider/cm/schema_cm.go
@@ -328,8 +328,8 @@ type CMKeyJSON struct {
 	Password                 string                   `json:"password,omitempty"`
 	ProcessStartDate         string                   `json:"processStartDate,omitempty"`
 	ProtectStopDate          string                   `json:"protectStopDate,omitempty"`
-	RevocationReason         string                   `json:"revocationMessage,omitempty"`
-	RevocationMessage        string                   `json:"revocationReason,omitempty"`
+	RevocationReason         string                   `json:"revocationReason,omitempty"`
+	RevocationMessage        string                   `json:"revocationMessage,omitempty"`
 	RotationFrequencyDays    string                   `json:"rotationFrequencyDays,omitempty"`
 	SecretDataEncoding       string                   `json:"secretDataEncoding,omitempty"`
 	SecretDataLink           string                   `json:"secretDataLink,omitempty"`
--- a/internal/provider/cm/schema_cm_test.go
+++ b/internal/provider/cm/schema_cm_test.go
@@ -1,0 +1,43 @@
+package cm
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+// TestCMKeyJSON_RevocationFieldsJSONTags guards against a regression of TFIN-286,
+// where RevocationReason and RevocationMessage carried swapped JSON tags. It asserts
+// the wire format marshals each value under its matching CipherTrust Manager key.
+func TestCMKeyJSON_RevocationFieldsJSONTags(t *testing.T) {
+	payload := CMKeyJSON{
+		RevocationReason:  "REASON_SENTINEL",
+		RevocationMessage: "MESSAGE_SENTINEL",
+	}
+
+	b, err := json.Marshal(payload)
+	if err != nil {
+		t.Fatalf("json.Marshal(CMKeyJSON) failed: %v", err)
+	}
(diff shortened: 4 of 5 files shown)


zubeensyed:
exclude the changelog

changelog.md
```diff
@@ -1,5 +1,8 @@
 # 1.0.1
 
+## Bug Fixes
+    ciphertrust_cm_key no longer swaps revocation_reason and revocation_message in the CipherTrust Manager request body (TFIN-286)
+
```
zubeensyed: get rid of the changelog

terraform-provider-ciphertrust
zubeensyed, line 1: get rid of this binary

zubeensyed:
Addressed in `e4fb593`: TFIN-286: address review feedback (round 1)

changelog.md
zubeensyed: Fixed in `a68792d`: TFIN-286: address review feedback (round 2)

terraform-provider-ciphertrust
zubeensyed: Fixed in `a68792d`: TFIN-286: address review feedback (round 2)

PR #156, opened 2026-06-05

zubeensyed:
## Summary

- Fixes swapped JSON serialization tags for `RevocationReason` and `RevocationMessage` on the `CMKeyJSON` struct in `internal/provider/cm/schema_cm.go`, which affects the `ciphertrust_cm_key` resource.
- Before the fix, `RevocationReason` serialized to `"revocationMessage"` and `RevocationMessage` serialized to `"revocationReason"`, so Create and Update operations stored each value in the wrong CipherTrust Manager field.
- Adds two unit tests pinning the wire contract and `omitempty` behaviour, plus one acceptance test that verifies the server-side values round-trip correctly.
- Adds a "Bug Fixes" entry to `changelog.md`.

## Motivation

Implements [TFIN-286]: `ciphertrust_cm_key` `revocation_reason` and `revocation_message` are serialized with swapped JSON tags causing perpetual drift.

The `CMKeyJSON` struct used by the `ciphertrust_cm_key` resource carried inverted camelCase JSON tags on the two revocation fields. The Terraform attribute names (`revocation_reason`, `revocation_message`) and the Go field names were correct, but the outbound wire payload swapped their values. A user setting `revocation_reason = "Unspecified"` and `revocation_message = "decommissioned"` would have those values stored on CM under the opposite fields, producing perpetual drift and incorrect server-side state.

- Un-swaps the two camelCase JSON tags on the `CMKeyJSON` struct:
  - `RevocationReason` now serializes to `json:"revocationReason,omitempty"` (was `json:"revocationMessage,omitempty"`).
  - `RevocationMessage` now serializes to `json:"revocationMessage,omitempty"` (was `json:"revocationReason,omitempty"`).
- `omitempty` is preserved on both fields. No Go field rename and no Terraform attribute rename.

The fix propagates automatically to both CRUD call sites in `internal/provider/cm/resource_cm_key.go`, which populate the shared `CMKeyJSON` struct rather than an inline map:
- Create (around lines 770–774)
- Update (around lines 1208–1212)

No code change was required in `resource_cm_key.go`.

The affected resource uses the `URL_KEY_MANAGEMENT` endpoint (`api/v1/vault/keys2`), already defined in `internal/provider/common/urls.go`. No new URL constant was added. The corrected tag names (`revocationReason` / `revocationMessage`) match the camelCase field names CM uses on the key object, consistent with the read path in `data_source_cm_keys.go`.

- Adds a "Bug Fixes" entry under the `# 1.0.1` heading noting the fix and that users who previously inverted their HCL values to work around the bug must swap them back.

- `TestCMKeyJSON_RevocationTagSerialization` — marshals a `CMKeyJSON` with both revocation fields set and asserts each value serializes to its correctly-named tag (and explicitly that the tags are not swapped). Pure unit test, no CM dependency.
- `TestCMKeyJSON_RevocationTagOmitempty` — marshals with only one revocation field set at a time and asserts the empty field is omitted from the wire payload. Guards against accidental loss of `omitempty`.

- Adds `TestAccCipherTrustCMKey_RevocationFieldsRoundtrip` — creates a `ciphertrust_cm_key` with `revocation_reason` and `revocation_message` set, then asserts the server-side CM fields hold the matching (non-swapped) values.
- Adds the `testAccCheckCMKeyRevocationOnServer` helper, which fetches the key from CM by resource ID via `client.GetById(ctx, uuid, keyID, common.URL_KEY_MANAGEMENT)` and asserts the server-side `revocationReason` and `revocationMessage` fields (read with `gjson`) match the HCL inputs. This is the only end-to-end guarantee that the tag fix produced the intended CM-side state.

## Schema attributes

No schema attribute changes. The existing attributes on `ciphertrust_cm_key` are unchanged:

|---|---|---|---|
| `revocation_reason` | `types.String` | Optional | Now serializes to `revocationReason` (mutable, no `RequiresReplace`) |
| `revocation_message` | `types.String` | Optional | Now serializes to `revocationMessage` (mutable, no `RequiresReplace`) |

This is a pure outbound-serialization fix.

This change does not touch `Read()`. Per the project-wide TFIN-174 constraint, `Read()` for `ciphertrust_cm_key` remains empty and this ticket did not ask for it to be implemented.

**User-visible consequence:** After `terraform apply`, subsequent `terraform plan` runs will not reconcile the resource against CipherTrust Manager. A dedicated follow-up (TFIN-174) addresses drift detection across the provider. Importantly, because tfstate values were always stored correctly (only the wire payload was wrong), no state migration is required — leaving `Read()` empty avoids surfacing drift before the immutable-field fixes land.

## Breaking changes

- **Behaviour change for workaround users:** Users who previously swapped their `revocation_reason` and `revocation_message` HCL values to compensate for the bug will now see corrected CM-side values after upgrade and must invert their HCL back to the intended meaning. This is documented in `changelog.md`. No tfstate migration is performed.

## How to test

- [ ] `make test` — unit tests pass (`TestCMKeyJSON_RevocationTagSerialization`, `TestCMKeyJSON_RevocationTagOmitempty`).
  ```
  go test ./internal/provider/... -run TestAccCipherTrustCMKey_RevocationFieldsRoundtrip -v
  ```
  Scenarios covered:
  - **Create** — apply config with `revocation_reason = "Unspecified"` and `revocation_message = "decommissioned-by-acceptance-test"`; assert both state attributes match and the server-side `revocationReason`/`revocationMessage` fields hold the matching (non-swapped) values via `testAccCheckCMKeyRevocationOnServer`.
  - **Destroy** — `terraform destroy` cleans up the test key.

## Checklist

- [ ] CM API endpoint (`api/v1/vault/keys2`) verified — existing `URL_KEY_MANAGEMENT` constant reused, no inlined string literals.
- [ ] Confirm the build artifact `terraform-provider-ciphertrust` (compiled binary, currently untracked in the working tree) is not committed.

---

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,8 @@
 # 1.0.1
 
+## Bug Fixes
+    resource/ciphertrust_cm_key: fixed swapped serialization of revocation_reason and revocation_message (TFIN-286). Users who previously swapped their HCL values to compensate must invert them.
+
 ## New Resources
     ciphertrust_aws_acl
     ciphertrust_aws_key_import_material
--- a/internal/provider/cm/schema_cm.go
+++ b/internal/provider/cm/schema_cm.go
@@ -328,8 +328,8 @@ type CMKeyJSON struct {
 	Password                 string                   `json:"password,omitempty"`
 	ProcessStartDate         string                   `json:"processStartDate,omitempty"`
 	ProtectStopDate          string                   `json:"protectStopDate,omitempty"`
-	RevocationReason         string                   `json:"revocationMessage,omitempty"`
-	RevocationMessage        string                   `json:"revocationReason,omitempty"`
+	RevocationReason         string                   `json:"revocationReason,omitempty"`
+	RevocationMessage        string                   `json:"revocationMessage,omitempty"`
 	RotationFrequencyDays    string                   `json:"rotationFrequencyDays,omitempty"`
 	SecretDataEncoding       string                   `json:"secretDataEncoding,omitempty"`
 	SecretDataLink           string                   `json:"secretDataLink,omitempty"`
--- a/internal/provider/cm/schema_cm_test.go
+++ b/internal/provider/cm/schema_cm_test.go
@@ -6,10 +6,8 @@ import (
 	"testing"
 )
 
-// TestCMKeyJSON_RevocationTagSerialization pins the wire contract for the
-// revocation fields on ciphertrust_cm_key (TFIN-286). It guards against the
-// historical bug where RevocationReason and RevocationMessage were serialized
-// with swapped JSON tags, storing each value in the wrong CM field.
+// TestCMKeyJSON_RevocationTagSerialization pins the revocation field wire tags
+// against the swapped-tag bug (TFIN-286).
 func TestCMKeyJSON_RevocationTagSerialization(t *testing.T) {
 	payload := CMKeyJSON{
 		RevocationReason:  "Unspecified",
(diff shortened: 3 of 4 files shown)


zubeensyed:
@cm-agent shorten the comment size.

zubeensyed:
Addressed in `b717873`: TFIN-286: address review feedback (round 1)

PR #158, opened 2026-06-05
TFIN-203 CTE | Drift Detection (CTE CSI Group)

chinmay-thales:
**TFIN-203 CTE | Drift Detection (CTE CSI Group)**

**Changes**
- Added drift detection functionality for CTE CSI Group resource

**Testing**

### DIFF
--- a/examples/resources/cte_csigroup/resource.tf
+++ b/examples/resources/cte_csigroup/resource.tf
@@ -28,15 +28,23 @@ provider "ciphertrust" {
   password = "ChangeMe101!"
 }
 
+# Create CSI Group
 resource "ciphertrust_cte_csigroup" "test_csi_group" {
   name                     = "TF_CSI_GROUP"
+
   kubernetes_namespace     = "default"
+
   kubernetes_storage_class = "standard"
 
-  description    = "tf test csi group"
+  description    = "tf test csi group.."
+ 
+  guard_policies = {
+        test-csi1 = {},
+        test-csi2 = {},
+        test-csi3 = {},
+  }
 
 
-# Create CSI Group
   ############################################################
   # Operation type to perform on the CSI Group
   #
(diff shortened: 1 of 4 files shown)


chinmay-thales:

examples/resources/cte_csigroup/resource.tf
```diff
@@ -45,19 +53,17 @@ resource "ciphertrust_cte_csigroup" "test_csi_group" {
...
-  # - update               : Update description / client_profile
-  # - add-clients          : Add clients to CSI group
-  # - remove-client        : Remove a client from CSI group
-  # - add-guard-policies   : Add guard policies to CSI group
-  # - update-guard-policy  : Enable/Disable guard policies
-  # - remove-guard-policy  : Remove guard policies
-  #
-  # op_type = "add-clients/remove-client/update/add-guard-policies/update-guard-policy/remove-guard-policy"
+  # - update                : Update description / client_profile
+  # - add-clients           : Add clients to CSI group
```
rahulrampura-thales: We can't add clients to CSI group even manually from CM, so this is NOT a valid use case for terraform.
It only gets added automatically on enrollment/registration of CTE CSI client.

examples/resources/cte_csigroup/resource.tf
```diff
@@ -45,19 +53,17 @@ resource "ciphertrust_cte_csigroup" "test_csi_group" {
...
+  # - remove-clients        : Remove clients from CSI group
```
rahulrampura-thales: Same as above. Remove functionality not possible manually from CM. NOT a valid use case.
Can remove this functionality from this change

chinmay-thales: Removed the functionality of add/remove clients from the CSI group resource.

PR #159, opened 2026-06-05
TFIN-292: Dup | terraform plan silently shows "no changes" when CM resources are deleted or modified out-of-band due to empty Read() implementation

ashwani573:
# Implement `Read()` for `ciphertrust_cm_key` to enable drift detection

## Summary

- Implements the previously-empty `Read()` for the `ciphertrust_cm_key` resource so `terraform plan`/`refresh` reconciles state against CipherTrust Manager (CM) via `GET api/v1/vault/keys2/{id}`.
- Adds a shared private helper `setCMKeyState(...)` (with small `refreshString`/`refreshInt64`/`refreshBool` helpers) reused by `Create`, `Read` and `Update` so all three flows converge on identical state for readable attributes.
- Handles out-of-band deletion: a 404 on `Read` removes the resource from state (warning, not error) so the next plan proposes a recreate; a 404 on `Delete` is demoted to a warning so the resource is dropped cleanly.
- Replaces the old single `TestResourceCMKey` with focused acceptance tests covering create, update, out-of-band delete and attribute drift.
- No schema, model, `provider.go`, `urls.go`, client, examples or docs changes.

## Motivation

Implements TFIN-292: `terraform plan` silently shows "no changes" when CM resources are deleted or modified out-of-band, because `ciphertrust_cm_key`'s `Read()` was an empty stub that issued no API call. Terraform never reconciled state against CM, so out-of-band deletion and attribute drift were invisible.

> **Note:** The internal Jira ticket and approved implementation plan are referenced in the commit message. This description is self-contained for external reviewers who cannot access internal systems.

- **`Read` (was empty, now implemented):**
  - Reads prior state, then calls `r.client.GetById(ctx, <traceUUID>, state.ID.ValueString(), common.URL_KEY_MANAGEMENT)`.
  - On 404 (`strings.Contains(err.Error(), notFoundError)`, where `notFoundError = "status: 404"`): emits `tflog.Warn`, adds a warning diagnostic, and calls `resp.State.RemoveResource(ctx)`, then returns. The next plan proposes a recreate.
  - On any other API error: hard `resp.Diagnostics.AddError(...)`, returns, and leaves state untouched (so a transient CM outage does not silently drop the resource).
  - On success: repopulates state via `setCMKeyState(...)` and `resp.State.Set`.
- **New helper `setCMKeyState(ctx, response, state, diags)`** plus `refreshString`/`refreshInt64`/`refreshBool`. Shared by Create, Read and Update.
- **`Create`:** after the existing `PostDataV2` and `plan.ID = ...`, now calls `setCMKeyState` so Create and Read produce identical state. Write-only inputs are untouched (see below).
- **`Update`:** after the existing PATCH (`UpdateData`), now performs a post-PATCH `r.client.GetById(...)` then `setCMKeyState`, mirroring the post-update GET in `resource_aws_cloudhsm_key.go`. A GET failure here is non-fatal — the PATCH already succeeded, so it warns and keeps the plan values.
- **`Delete`:** the error branch now demotes a 404 from `DeleteByID` to a warning (alongside the pre-existing "key is not deletable" + `remove_from_state_on_destroy` handling), so an already-gone key drops from state cleanly.

Endpoint verified against swagger (`operations.md`): `GET /v1/vault/keys2/{id}` (area `Keys`) exists and maps to the existing `common.URL_KEY_MANAGEMENT = "api/v1/vault/keys2"`. No new URL constant was needed.

Acceptance tests (package `provider`, skipped unless `TF_ACC=1`). The old `TestResourceCMKey` is replaced with:

- `TestAccResourceCMKey_basic` — creates a minimal AES key; asserts `id`, `name`, `algorithm`, `key_size`, `usage_mask`.
- `TestAccResourceCMKey_update` — changes `description`, `usage_mask` and `labels`; asserts the post-PATCH GET-driven state reflects the new values.
- `TestAccResourceCMKey_outOfBandDelete` — deletes the key via `client.DeleteByURL` in `PreConfig`, then `RefreshState: true` + `ExpectNonEmptyPlan: true` to verify the 404 → remove-from-state → recreate path.
- `TestAccResourceCMKey_attributeDrift` — mutates `description` + `labels` via `client.UpdateData` in `PreConfig`, then `RefreshState: true` + `ExpectNonEmptyPlan: true` to verify Read-driven drift detection.

Out-of-band calls use the existing `createCMClient()` helper in `provider_test.go`.

## Key design decision — refresh-only-if-set, mutable attributes only

The `ciphertrust_cm_key` schema attributes are **Optional-only** (none are `Computed`). Naively refreshing every attribute from the CM response would produce perpetual spurious drift (config `null` vs server default). To avoid this, `setCMKeyState` follows a deliberate, narrow contract:

- **Refresh only if already set.** An attribute is refreshed only when it is already set in state (user-configured or previously persisted). Unconfigured (null/unknown) attributes are left untouched. When a configured attribute is missing/empty in the response, it is set to `null` to surface that it was cleared out-of-band.
- **Mutable + identity + labels only.** Only the mutable CM `PatchKey` attributes, lifecycle dates, identity (`id`, `uuid`, `name`) and `labels` are refreshed. Immutable create/import-only attributes (`algorithm`, `key_size`, `curveid`, `object_type`, `format`, `default_iv`, `xts`, `cert_type`, `id_size`, `padded`, `signing_algo`) are intentionally **not** refreshed: they cannot change out-of-band, and re-reading them would risk surfacing server-side case/format normalization as spurious drift — a concern tracked separately by TFIN-195 and which the project guidance asks not to surface in this change.
- **`labels` (`types.Map`)** is refreshed from the free-form CM object, coercing scalar values to their string form; non-scalar values are skipped with a warning so the map round-trips without spurious drift.
- **Write-only inputs are never touched** (`material`, `password`, `wrap_*`, `wrapping_*`, `mac_sign_*`, `secret_data_*`, `template_id`, `encoding`, `empty_material`, `generate_key_id`, `assign_self_as_owner`, `hkdf_create_parameters`, `public_key_parameters`, `wrap_hkdf`, `wrap_pbe`, `wrap_rsaaes`, `all_versions`, `remove_from_state_on_destroy`). CM has no readable counterpart for these, so Create/Update keep the plan-supplied values.
- **Nested `aliases` and `meta` are deliberately preserved (not refreshed).** The top-level `aliases` schema declares `index` as a `StringAttribute` while the model field is `types.Int64`; refreshing it would fail `State.Set`. This is a pre-existing schema/model mismatch and is out of scope for TFIN-292.

## Read() now implemented

`Read()` previously returned immediately. It now calls `c.GetById(ctx, uuid, id, common.URL_KEY_MANAGEMENT)`, unmarshals the response with `gjson`, and repopulates state via `setCMKeyState`.

**Fields read from the CM response** (per swagger `GET /v1/vault/keys2/{id}`, Terraform attribute ← CM JSON field):

- `id` ← `id`, `uuid` ← `uuid`, `name` ← `name`
- `activation_date` ← `activationDate`, `archive_date` ← `archiveDate`, `compromise_date` ← `compromiseDate`, `compromise_occurrence_date` ← `compromiseOccurrenceDate`, `deactivation_date` ← `deactivationDate`, `destroy_date` ← `destroyDate`, `process_start_date` ← `processStartDate`, `protect_stop_date` ← `protectStopDate`
- `description` ← `description`, `key_id` ← `keyId`, `muid` ← `muid`, `revocation_reason` ← `revocationReason`, `revocation_message` ← `revocationMessage`, `rotation_frequency_days` ← `rotationFrequencyDays`, `state` ← `state`
- `usage_mask` ← `usageMask`, `unexportable` ← `unexportable`, `undeletable` ← `undeletable`
- `labels` ← `labels` (scalar values coerced to strings; non-scalars skipped with a warning)

**404 handling:** On `Read`, a 404 calls `resp.State.RemoveResource(ctx)` (out-of-band delete → recreate on next plan). Any **non-404** error returns without modifying state — a transient CM outage will not silently drop the resource.

**Null/absent fields:** see the "refresh-only-if-set" contract above. A configured attribute that is missing or empty-string in the response is set to its typed `Null` value rather than `""`, to match Plugin-Framework null semantics and avoid drift.

## How to test

  ```
  go test ./internal/provider/... -run TestAccResourceCMKey -v
  ```
  Scenarios covered:
  - **Create** (`_basic`) — apply config; assert `id`, `name`, `algorithm`, `key_size`, `usage_mask`.
  - **Update** (`_update`) — change `description`, `usage_mask`, `labels`; re-apply; assert the post-PATCH state reflects the new values.
  - **Out-of-band delete** (`_outOfBandDelete`) — delete the key via the CM API in `PreConfig`; `RefreshState` + `ExpectNonEmptyPlan` confirms Read 404 removes the resource and the plan proposes a recreate.
  - **Attribute drift** (`_attributeDrift`) — mutate `description` + `labels` via the CM API in `PreConfig`; `RefreshState` + `ExpectNonEmptyPlan` confirms Read-driven drift detection.

> **Build/test could not be run in the implementation sandbox.** The only available Go toolchain is 1.18, which cannot parse this module's `go.mod` (`go 1.22.0`), and `gofmt` was unavailable. The change was verified by manual review (imports, struct field names, `client`/framework/`gjson` signatures, no duplicate declarations). The GitLab pipeline runs the real build and acceptance tests after the PR opens. **Reviewers should confirm the pipeline is green before merging.**

## Scope reconciliations reviewers should know

- The approved plan referenced `internal/provider/resource_cm_key.go` in package `provider`; the real file is **`internal/provider/cm/resource_cm_key.go`** in package **`cm`**. Reconciled against the live code.
- `Update`/`Delete` use **`UpdateData`/`DeleteByID`** (not the `UpdateDataV2`/`DeleteByURL` the plan assumed). Reconciled against the live client.
- The size attribute's HCL/schema key is **`key_size`** (not `size`).
- ⚠️ **`terraform import` test and `ImportStateVerify` steps from the plan were NOT implemented.** `resourceCMKey` does not implement `resource.ResourceWithImportState`, so import is unsupported; and the Optional-only schema cannot cold-round-trip every attribute. Adding import support is out of scope for this ticket.
- `notFoundError` is defined as a file-local `const notFoundError = "status: 404"`, matching the CM client error format (`"status: <code>, body: <...>"`) and mirroring the CCKM AWS resources.

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Read` and the new helper: `[resource_cm_key.go -> <Func>][uuid]` pattern.
- [ ] Fresh `uuid.New().String()` generated at the top of `Read` (and for the post-PATCH GET in `Update`).
- [ ] CM API endpoint verified against swagger spec (`GET /v1/vault/keys2/{id}`).
- [ ] `Read()` non-404 errors keep the resource in state — `resp.State.RemoveResource(ctx)` is called only on a 404.
- [ ] No schema changes; `docs/` left untouched (generated).
- [ ] Acceptance tests live at `internal/provider/resource_cm_key_test.go` (next to `provider.go`, not inside the `cm/` subfolder).

---


internal/provider/cm/resource_cm_key.go
```diff
@@ -1255,6 +1433,22 @@ func (r *resourceCMKey) Update(ctx context.Context, req resource.UpdateRequest,
...
+	// actually persisted (mirrors the post-update GET in resource_aws_cloudhsm_key.go).
+	// A failure here is non-fatal: the PATCH already succeeded, so keep the plan
+	// values and only warn.
+	getResponse, getErr := r.client.GetById(ctx, uuid.New().String(), plan.ID.ValueString(), common.URL_KEY_MANAGEMENT)
+	if getErr != nil {
+		details := utils.ApiError("Error reading CipherTrust key after update, state may be stale until the next refresh.", map[string]interface{}{"error": getErr.Error(), "id": plan.ID.ValueString()})
+		tflog.Warn(ctx, details)
+		resp.Diagnostics.AddWarning(details, "")
+	} else {
+		setCMKeyState(ctx, getResponse, &plan, &resp.Diagnostics)
```
rohitpandita-hub, line 1446: Does it have all the parameters mentioned in the plan on confluence page?

rohitpandita-hub (changes requested):

PR #164, opened 2026-06-08

ashwani573:
## Summary

- Implements `Read()` for `ciphertrust_groups` (`cm.NewResourceCMGroup`): calls `GET api/v1/usermgmt/groups/{name}` on every `terraform plan` and `terraform refresh`, enabling drift detection for out-of-band deletions and field modifications.
- Adds `ModifyPlan` to enforce immutability of `name` at plan time with an error diagnostic (no destroy-recreate; user must delete and recreate explicitly).
- Adds `ImportState` (`resource.ImportStatePassthroughID`) so existing CM groups can be brought under Terraform management via `terraform import`.
- Adds 404 guard in `Delete()` so out-of-band deletion of a group during `terraform destroy` is silently accepted rather than surfacing a spurious error.
- Adds `Computed: true` to the `description` schema attribute so CM-assigned defaults are preserved in state.
- Adds 6 acceptance test functions and updates the `ciphertrust_groups` example with an import comment.

## Motivation

Implements TFIN-293: `terraform plan` silently showed "no changes" for `ciphertrust_groups` regardless of whether the group had been deleted or modified directly in CipherTrust Manager. The root cause was a completely empty `Read()` body — no API call was ever made during plan or refresh, so Terraform's state was never reconciled against the CM server.

This change is scoped to `ciphertrust_groups` only. Other resources with the same empty-`Read()` issue are tracked in separate tickets, consistent with the provider-wide fix-order policy documented in `CLAUDE.md`.

**New interface assertions:**
- `resource.ResourceWithImportState` — wired via `ImportState` method.
- `resource.ResourceWithModifyPlan` — wired via `ModifyPlan` method.

**New constant:**
- `const notFoundError = "status: 404"` — local to the CM package; matches the value used across CCKM OCI resources.

**`ModifyPlan` (new method):**
- Skips create and destroy operations (null plan or null state).
- Compares `plan.Name` to `state.Name`; adds an `"Immutable attribute change detected"` error diagnostic listing `name` if they differ.
- This is an error-on-change guard, not `RequiresReplace` — the user sees a clear diagnostic and must destroy and recreate the resource explicitly.

**`Read` (previously empty — now implemented):**
1. Loads prior state via `req.State.Get`.
2. Calls `r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_GROUP)` — issues `GET api/v1/usermgmt/groups/{name}` (the group name doubles as the ID).
3. On 404 (`strings.Contains(err.Error(), notFoundError)`): logs a warning and calls `resp.State.RemoveResource(ctx)` to signal that the resource no longer exists in CM, prompting Terraform to propose recreation on the next apply.
4. On any other error: adds an error diagnostic and returns, leaving the resource in state.
5. Calls `setGroupState` to write scalar fields into state.
6. Calls `resp.State.Set` to persist the updated state.

**`setGroupState` (new helper):**
- Maps `id` and `name` from `gjson.Get(response, "name")` and `description` from `gjson.Get(response, "description")`.
- Map-typed attributes (`app_metadata`, `client_metadata`, `user_metadata`) are intentionally left unchanged from prior state — the schema uses `MapNestedAttribute`, which is incompatible with direct string-map assignment via gjson; prior state values are preserved on each refresh.

**`ImportState` (new method):**
- Delegates to `resource.ImportStatePassthroughID` using `path.Root("id")`.
- After import, `Read()` is called immediately and populates all tracked scalar fields.

**`Delete` (updated):**
- Added a 404 guard: if `DeleteByID` returns an error containing `notFoundError`, the error is swallowed — the group was already removed out-of-band, and Terraform removes it from state automatically.

**`Create` (minor):**
- Log messages corrected from `resource_cm_user.go` to `resource_cm_group.go`.
- Log end trace moved immediately after the API response log, before `resp.State.Set`.

**`Update` (minor):**
- Log message and error summary corrected from "Create" to "Update".

**Schema:**
- `description`: added `Computed: true`. This allows CM to return a default (empty string) value for groups created without a description, preventing a persistent diff on the next plan.

Six test functions are present (one pre-existing, extended; five new):

| Function | Scenario |
|---|---|
| `TestResourceCMGroup` | Extended original: adds `id` attribute-set check and `description` attribute check on the update step. |
| `TestAccCMGroup_basic` | Create with `name` and `description`; verify `id`, `name`, `description` in state. |
| `TestAccCMGroup_updateInPlace` | Update `description` in a second step; verify `id` unchanged and new value reflected. |
| `TestAccCMGroup_outOfBandDelete` | Delete group out-of-band via `client.DeleteByURL`; verify `RefreshState` + `ExpectNonEmptyPlan` signals recreation. |
| `TestAccCMGroup_outOfBandFieldDrift` | Patch `description` out-of-band via `client.UpdateData`; verify `PlanOnly` + `ExpectNonEmptyPlan` shows a diff. |
| `TestAccCMGroup_import` | Create, then import with `ImportState: true` + `ImportStateVerify: true`; verify all tracked attributes populated. |
| `TestAccCMGroup_immutableNameChangeRejected` | Attempt to change `name` after creation; assert `ExpectError` matching `(?i)immutable attribute`. |

Test helpers added: `groupName(prefix)` generates a collision-safe name using a UUID suffix; `groupConfig(name, description)` renders the minimal HCL block.

The test file is at `internal/provider/resource_cm_group_test.go` (alongside `provider.go`), consistent with the provider's test-file location convention.

### Modified file — `examples/resources/ciphertrust_groups/resource.tf`

- Added `description = "Managed by Terraform"` to the example block.
- Added inline comment noting that `name` is immutable after creation.
- Added `terraform import` usage comment at the end of the file.
- Fixed missing newline at end of file.

## Read() now implemented

`Read()` previously returned immediately without making any API call. This change implements it by calling `r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_GROUP)`, which issues `GET api/v1/usermgmt/groups/{name}` to CipherTrust Manager.

**Fields read from the CM API response and written to Terraform state:**

| Terraform attribute | CM JSON field | gjson path |
|---|---|---|
| `id` | `name` | `name` |
| `name` | `name` | `name` |
| `description` | `description` | `description` |

**Fields preserved from prior state (not overwritten by `Read()`):**

| Terraform attribute | Reason |
|---|---|
| `app_metadata` | `MapNestedAttribute` — not directly assignable from gjson; prior state preserved |
| `client_metadata` | Same as above |
| `user_metadata` | Same as above |

**404 handling:** A 404 response calls `resp.State.RemoveResource(ctx)`, which signals Terraform that the group no longer exists. On the next `terraform apply`, Terraform will propose recreating it. This is the correct behaviour for confirmed out-of-band deletion.

**Non-404 errors:** A non-404 API failure adds an error diagnostic and returns without modifying state. The resource remains in state; the error is surfaced to the user.

**Note on ID / name duality:** The CM groups API path parameter is `{name}`, not a UUID. The provider sets `plan.ID = plan.Name` in `Create()`, so `state.ID.ValueString()` holds the group name. `GetById` appends this value to `common.URL_GROUP` as the path segment, which is correct for `GET api/v1/usermgmt/groups/{name}`.

**Swagger cross-check:** The endpoint `GET /v1/usermgmt/groups/{name}` is confirmed in the swagger operations index (`auth-users` area). `URL_GROUP = "api/v1/usermgmt/groups"` (unchanged pre-existing constant) maps correctly to this endpoint. No new URL constant was required.

## Breaking changes

- `name`: enforced immutable after creation via `ModifyPlan`. Changing `name` in a `.tf` file now produces an error diagnostic at `terraform plan` time rather than silently succeeding or triggering destroy-recreate. Users must run `terraform destroy` followed by `terraform apply` to rename a group.

Note: `name` is not given `RequiresReplace()`. The error-on-change approach via `ModifyPlan` is the convention mandated for this provider — it gives the user a clear message rather than an implicit destroy-recreate that could cause unintended data loss.

## How to test

  ```
  go test ./internal/provider/... -run 'TestAccCMGroup|TestResourceCMGroup' -v
  ```
  Scenarios covered:
  - **Create** — apply config; assert `id`, `name`, `description` populated in state.
  - **Update** — change `description`; re-apply; assert CM reflects the new value and `id` is unchanged.
  - **Out-of-band delete** — delete group directly in CM; verify `terraform plan` shows recreation.
  - **Out-of-band field drift** — patch `description` directly in CM; verify `terraform plan` shows a diff.
  - **Import** — `terraform import`; verify all tracked attributes populated in imported state.
  - **Immutable name change** — change `name` in config; verify plan-time error diagnostic, no destroy-recreate.
  - **Destroy** — `terraform destroy`; confirm resource is absent in CM.
- [ ] Manual smoke-test: apply example, verify in CM UI, destroy.

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Read`, `Create`, `ImportState` CRUD methods: `[resource_cm_group.go -> <Func>][uuid]` pattern confirmed in diff.
- [ ] Fresh `uuid.New().String()` at the top of each CRUD method — confirmed in diff.
- [ ] URL constant defined in `urls.go` — `URL_GROUP` pre-existing; no inlined string literals introduced.
- [ ] CM API endpoint verified against swagger spec (`GET /v1/usermgmt/groups/{name}` confirmed in `auth-users` area).
- [ ] `Read()` 404 calls `resp.State.RemoveResource(ctx)` — confirmed in diff.
- [ ] `app_metadata`, `client_metadata`, `user_metadata` drift detection is not yet implemented (map fields preserved from prior state). A follow-up ticket should address full map field reconciliation for these attributes.

---


internal/provider/cm/resource_cm_group.go
```diff
@@ -236,9 +322,19 @@ func (d *resourceCMGroup) Configure(_ context.Context, req resource.ConfigureReq
...
+
+// setGroupState maps API response fields into the TFSDK state struct.
+// Map fields (app_metadata, client_metadata, user_metadata) are intentionally left
+// unchanged — the schema declares them as MapNestedAttribute which is incompatible with
+// direct string-map assignment. The caller's prior state (loaded via req.State.Get) is
+// preserved for those fields; only the scalar fields are refreshed from the API.
+func setGroupState(_ context.Context, response string, state *CMGroupTFSDK, _ *diag.Diagnostics) {
+	state.ID = types.StringValue(gjson.Get(response, "name").String())
+	state.Name = types.StringValue(gjson.Get(response, "name").String())
+	state.Description = types.StringValue(gjson.Get(response, "description").String())
```
rohitpandita-hub, line 338: @cm-agent It doesn't cover all attributes, check the confluence page

rohitpandita-hub (changes requested):

internal/provider/cm/resource_cm_group.go
ashwani573: Fixed in `a59d4ba`: TFIN-293: address review feedback (round 1)

PR #166, opened 2026-06-09

ashwani573:
## Summary

- Implements `Read()` for the existing `ciphertrust_groups` resource (`internal/provider/cm/resource_cm_group.go`), ending the silent "no changes" behaviour when groups are deleted or modified directly in CipherTrust Manager.
- Adds `ImportState()` via `resource.ImportStatePassthroughID`, enabling `terraform import` for CM groups.
- Refactors `Create()` and `Update()` to call `GetById` + the new `setGroupState` helper after each write, so Terraform state is always populated from the live API response.
- Extends `internal/provider/resource_cm_group_test.go` with four new acceptance-test scenarios covering import, update drift-free round-trip, 404 out-of-band deletion, and field-level drift.

## Motivation

Implements TFIN-293: `terraform plan` silently shows "no changes" when CM groups are deleted or modified out-of-band because `Read()` was completely empty — it never called the CM API, so Terraform's refresh cycle had nothing to compare against.

> **Note:** The Jira ticket and associated implementation plan are referenced in the commit message. This description is self-contained for external reviewers.

- Adds `const notFoundError = "status: 404"` — matches the sentinel used in the CCKM/OCI package (`resource_oci_byok_key.go`). No equivalent constant existed in the CM package before this change.
- Adds `setGroupState(ctx, response, state, diags)` helper. Uses `gjson` to extract `name` and `description` from the CM API JSON response and writes them into the TFSDK struct. The `id` field is set to the group's `name` value, consistent with the existing convention in `Create()` (`plan.ID = plan.Name`) — the CM Groups API identifies groups by name, not a separate UUID. Map fields (`app_metadata`, `client_metadata`, `user_metadata`) are intentionally excluded from this helper because the schema declares them as `schema.MapNestedAttribute` without a `NestedObject` definition, which means their element type is indeterminate; overwriting them from the API response would cause `resp.State.Set` to fail.
- Implements `Read()`:
  1. Generates a trace UUID and emits `tflog.Debug` start/end via `defer`.
  2. Loads current state via `req.State.Get`.
  3. Calls `r.client.GetById(ctx, id, groupID, common.URL_GROUP)`.
  4. On 404 (`strings.Contains(err.Error(), notFoundError)`): emits `tflog.Warn`, calls `resp.Diagnostics.AddWarning`, calls `resp.State.RemoveResource(ctx)`, and returns.
  5. On non-404 error: emits `tflog.Error`, calls `resp.Diagnostics.AddError` (includes `groupID` and the error message), and returns without modifying state.
  6. On success: calls `setGroupState`, checks `resp.Diagnostics.HasError()`, then calls `resp.State.Set`.
- Adds `resource.ResourceWithImportState` to the compile-time interface assertions and implements `ImportState` via `resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)`.
- Fixes pre-existing copy-paste bugs in log messages: `resource_cm_user.go -> Create/Update` corrected to `resource_cm_group.go -> Create/Update`.
- Fixes error message in `Update()`: `"Invalid data input: Group Creation"` corrected to `"Invalid data input: Group Update"`.
- Refactors `Create()` and `Update()` to call `GetById` + `setGroupState` after the write operation, so computed fields are always populated from the API response rather than left empty.

**Endpoint verified against swagger (`operations.md`, area `auth-users`):**
- `GET /v1/usermgmt/groups/{name}` — confirmed present; this is what `GetById` calls.
- `PATCH /v1/usermgmt/groups/{name}` — confirmed present; used by `Update()` via `c.UpdateData`.

**Note — `created_at` / `updated_at` not mapped:** The implementation plan listed these attributes as candidates for state population, but `CMGroupTFSDK` (defined in `schema_cm.go`) does not include `created_at` or `updated_at` fields, and the resource schema does not declare them. They are therefore absent from `setGroupState`. If those attributes are needed, a separate schema-extension ticket is required.

**Note — `id` attribute lacks `UseStateForUnknown`:** The `id` schema attribute is declared as `Computed` without `stringplanmodifier.UseStateForUnknown()`. This is a pre-existing condition not introduced by this PR. Without this modifier, plans may show a spurious `id` change on configurations that reference the id attribute. A follow-up change should add the modifier.

- `TestResourceCMGroup` (improved): now asserts `id` is set and both `name` and `description` are populated after `Create` and `Update`, confirming that `Read()` feeds back into state.
- `TestAccCMGroup_import` (new): creates a group, then runs `terraform import` using the group's `id`; verifies state is fully populated.
- `TestAccCMGroup_update` (new): applies, updates `description`, verifies state matches the new value, then runs a `PlanOnly` step with `ExpectNonEmptyPlan: false` to confirm `Read()` returns the updated value and state is consistent.
- `TestAccCMGroup_readDrift_404` (new): creates a group, deletes it out-of-band via `client.DeleteByURL`, then runs a `RefreshState: true` + `ExpectNonEmptyPlan: true` step to verify `Read()` removes the resource from state on 404 and the plan proposes recreation.
- `TestAccCMGroup_readFieldDrift` (new): creates a group, updates `description` out-of-band via `client.UpdateData`, then runs a `RefreshState: true` + `ExpectNonEmptyPlan: true` step to verify `Read()` propagates the out-of-band change and the plan shows a diff.

No changes were made to `internal/provider/common/urls.go` or `provider.go` — `URL_GROUP = "api/v1/usermgmt/groups"` already exists and `cm.NewResourceCMGroup` was already registered.

## Schema attributes

The resource schema was not changed by this PR. For reference, the attributes that `Read()` now populates from the CM API response are:

| Attribute | Type | Cardinality | Populated by Read() | Notes |
|---|---|---|---|---|
| `id` | `types.String` | Computed | Yes | Set to `gjson.Get(response, "name")` — group name is the CM identifier |
| `name` | `types.String` | Required | Yes | `gjson.Get(response, "name")` |
| `description` | `types.String` | Optional | Yes | `gjson.Get(response, "description")` |
| `app_metadata` | `types.Map` | Optional | No | Excluded from `setGroupState` — see note above |
| `client_metadata` | `types.Map` | Optional | No | Excluded from `setGroupState` — see note above |
| `user_metadata` | `types.Map` | Optional | No | Excluded from `setGroupState` — see note above |

## Read() now implemented

`Read()` previously returned immediately with no API call. This change implements it by calling `r.client.GetById(ctx, id, groupID, common.URL_GROUP)`, unmarshalling the CM API response via `gjson`, and repopulating Terraform state.

- `name` (CM JSON: `name`) → `state.Name` and `state.ID`
- `description` (CM JSON: `description`) → `state.Description`

**404 handling:** A 404 response removes the resource from Terraform state (`resp.State.RemoveResource(ctx)`) and emits a warning diagnostic. This is intentional: the group no longer exists in CM, so Terraform should treat it as gone and plan to recreate it. This differs from a non-404 error (e.g. network timeout), where state is preserved and an error diagnostic is raised instead, to avoid destroying state on transient failures.

**Map field handling:** `app_metadata`, `client_metadata`, and `user_metadata` are not overwritten from the API response. Their values in state remain whatever was last written by `Create()` or `Update()`. This is a known limitation arising from the schema definition; see the note in "What changed" above.

## How to test

- [ ] Acceptance tests (require a live CM instance — set `TF_ACC=1` and `CIPHERTRUST_ADDRESS`, `CIPHERTRUST_USERNAME`, `CIPHERTRUST_PASSWORD` env vars):
  ```
  go test ./internal/provider/... -run 'TestResourceCMGroup|TestAccCMGroup' -v
  ```
  Scenarios covered:
  - **Create + Read state population** (`TestResourceCMGroup`) — apply config with `name` and `description`; assert `id`, `name`, and `description` are correct in state.
  - **Update** (`TestResourceCMGroup`, `TestAccCMGroup_update`) — change `description`; re-apply; assert CM reflects the new value; confirm subsequent plan is empty.
  - **Import** (`TestAccCMGroup_import`) — apply, then `terraform import`; verify `id` and `name` are populated.
  - **404 drift** (`TestAccCMGroup_readDrift_404`) — delete group out-of-band; refresh state; plan must propose recreation.
  - **Field drift** (`TestAccCMGroup_readFieldDrift`) — change `description` out-of-band; refresh state; plan must show a diff on `description`.
  - **Destroy** — implicit in all test cases via `TestCase` teardown.
- [ ] Manual smoke-test: `terraform apply` with `ciphertrust_groups`, verify in CM UI, `terraform destroy`.

## Breaking changes

None. This PR only adds behaviour to a method that previously did nothing. No existing attribute is removed or made immutable.

## Checklist

- [ ] `tflog.Debug` at entry/exit of `Read()` and `tflog.Trace` at entry/exit of `Create()` — consistent with house style.
- [ ] `URL_GROUP` constant used throughout — no inlined string literals.
- [ ] CM API endpoint `GET /v1/usermgmt/groups/{name}` verified against swagger `operations.md`.
- [ ] `Read()` 404 removes resource from state via `resp.State.RemoveResource(ctx)` and emits a warning (not an error).
- [ ] Non-404 errors in `Read()` surface as error diagnostics; state is not cleared.
- [ ] `ImportState` implemented via `resource.ImportStatePassthroughID`.
- [ ] Map attributes (`app_metadata`, `client_metadata`, `user_metadata`) are not read back from CM API — follow-up ticket needed to fix schema definition before these can be reconciled.
- [ ] `id` attribute missing `UseStateForUnknown` — pre-existing; follow-up ticket needed.

---


internal/provider/cm/resource_cm_group.go
```diff
@@ -62,10 +69,22 @@ func (r *resourceCMGroup) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+// setGroupState maps API response fields into Terraform state for a CM group.
+// Only scalar fields (id, name, description) are updated from the response;
+// map fields (app_metadata, client_metadata, user_metadata) are left unchanged
+// because schema.MapNestedAttribute without NestedObject has an indeterminate
+// element type that would cause State.Set to fail if overwritten.
+func setGroupState(_ context.Context, response string, state *CMGroupTFSDK, _ *diag.Diagnostics) {
+	name := gjson.Get(response, "name").String()
+	state.ID = types.StringValue(name)
+	state.Name = types.StringValue(name)
+	state.Description = types.StringValue(gjson.Get(response, "description").String())
```
rohitpandita-hub, line 81: @cm-agent , did we have all parameters catered

rohitpandita-hub (changes requested):

internal/provider/cm/resource_cm_group.go
ashwani573: Fixed in `ec768f8`: TFIN-293: address review feedback (round 1)

rohitpandita-hub: @cm-agent , match the attributes listed on confluence plan and make sure all are covered

rohitpandita-hub (changes requested):

internal/provider/cm/resource_cm_group.go
rohitpandita-hub: @cm-agent i dont see following parameters catered:
parent_ca_id 
created_at | 
updated_at

rohitpandita-hub (changes requested):

PR #178, opened 2026-06-10

ashwani573:
## Summary

- Implements `Read()` for `ciphertrust_cm_key`, replacing the empty stub with a full CM API call via `GET /v1/vault/keys2/{id}`, enabling drift detection for out-of-band key changes.
- Fixes a schema type bug: `aliases[*].index` was declared as `schema.StringAttribute` but the `KeyAliasTFSDK` struct field is `types.Int64`; corrected to `schema.Int64Attribute` (matching the CM API which returns `index` as an integer).
- Adds `"github.com/hashicorp/terraform-plugin-framework/attr"` import required for `labels` map construction in `Read()`.
- Adds four acceptance tests in `internal/provider/resource_cm_key_test.go` covering drift detection, state population, out-of-band deletion, and write-only field preservation.

## Motivation

Implements TFIN-273: Drift Detection | ciphertrust_cm_key.

Before this change, the `Read()` method for `ciphertrust_cm_key` was an empty stub — it returned immediately without calling the CM API. As a result, any key attribute changes made directly in CipherTrust Manager (out-of-band) were invisible to Terraform: `terraform plan` would always report no changes even when the live key differed from state. This change implements the full Read() cycle so that Terraform can detect and report drift.

**Schema fix (`aliases[*].index`):**

The `index` field inside the `aliases` list nested attribute was typed as `schema.StringAttribute`. The corresponding Go struct field `KeyAliasTFSDK.Index` is `types.Int64`, and the CM API `KeyAlias` definition specifies `index` as an integer. The mismatch caused a panic at `resp.State.Set` when the CM API returned a numeric index. Fixed by changing the declaration to `schema.Int64Attribute`.

**Import addition:**

Added `"github.com/hashicorp/terraform-plugin-framework/attr"` to the import block. This package is needed for `attr.Value` when building the `types.MapValue` for the `labels` attribute in `Read()`.

**`Read()` implementation:**

- Loads prior state via `req.State.Get` to preserve all write-only fields (`material`, `password`, `encoding`, `format`, all `wrap_*` fields) that are not returned by the CM API.
- Returns early (no error) if `state.ID` is null or empty — handles the pre-create state during Terraform planning.
- Calls `r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_KEY_MANAGEMENT)` — confirmed against swagger: `GET /v1/vault/keys2/{id}`.
- On 404: calls `resp.State.RemoveResource(ctx)` to signal out-of-band deletion, causing Terraform to plan recreation.
- On other errors: calls `resp.Diagnostics.AddError`.
- Populates 27 scalar Saved=Yes fields from the gjson-parsed response: `id`, `name`, `algorithm`, `state`, `usage_mask`, `unexportable`, `undeletable`, `description`, `activation_date`, `deactivation_date`, `archive_date`, `process_start_date`, `protect_stop_date`, `rotation_frequency_days`, `revocation_reason`, `revocation_message`, `compromise_occurrence_date`, `curveid`, `object_type`, `key_size`, `xts`, `cert_type`, `id_size`, `destroy_date`, `compromise_date`, `default_iv`.
- Populates `aliases` as `[]*KeyAliasTFSDK` with `alias`, `index` (`types.Int64`), and `type` fields.
- Populates `labels` as `types.MapValue(types.StringType, ...)`, falling back to an empty map when the field is absent from the response.
- Populates `meta` nested struct: `owner_id`, nine `permissions` string-slice fields (using the uppercase JSON paths `DecryptWithKey`, `EncryptWithKey`, etc. as they appear in the CM API response), and three `cte` sub-fields (`persistent_on_client`, `encryption_mode`, `cte_versioned`).
- Populates `public_key_parameters` nested struct: eight scalar fields and a `[]KeyAliasTFSDK` aliases slice.

**Known pre-existing issue mirrored by `Read()`:**

The `CMKeyTFSDK` struct has `RevocationReason` and `RevocationMessage` with their JSON struct tags swapped relative to their Terraform attribute names. `Read()` intentionally mirrors this behaviour: `state.RevocationReason` is populated from `gjson.Get(response, "revocationMessage")` and `state.RevocationMessage` from `gjson.Get(response, "revocationReason")`. Fixing the swap is out of scope for this ticket — it would constitute a breaking state migration for existing users and is noted here for a follow-up ticket.

Adds four acceptance tests. All tests use the existing `testAccProtoV6ProviderFactories`, `providerConfig`, `checkStep`, and `createCMClient` helpers already present in the `provider` package.

- `TestAccCMKey_driftDetection`: Creates a key, confirms no spurious drift on an immediate plan, then mutates `description` out-of-band via `client.UpdateData` and asserts that the subsequent plan detects a non-empty diff (`ExpectNonEmptyPlan: true`).
- `TestAccCMKey_readPopulatesState`: Creates a key with explicit `usage_mask`, `unexportable`, `undeletable`, and `description`, then uses `RefreshState: true` to trigger `Read()` and asserts all scalar fields are populated correctly in state.
- `TestAccCMKey_oobDeleteRemovesFromState`: Creates a key, deletes it out-of-band via `client.DeleteByID`, then uses `RefreshState: true` and asserts Terraform plans to recreate it (`ExpectNonEmptyPlan: true`), confirming the 404 → `RemoveResource` path.
- `TestAccCMKey_writeOnlyFieldsPreserved`: Creates a key with an explicit `material` value and confirms via `RefreshState` that `Read()` leaves the write-only field intact at its prior-state value.

## CM API endpoint verification

The `Read()` call uses `common.URL_KEY_MANAGEMENT = "api/v1/vault/keys2"` appended with the resource ID. This resolves to `GET /v1/vault/keys2/{id}`, which is confirmed present in the swagger operations index (`cm-keys` area, verb `GET`, path `/v1/vault/keys2/{id}`, description "Get"). No mismatch was found between the implementation and the swagger specification.

The swagger `KeyAlias` definition specifies `index` as type `integer` — consistent with the schema fix from `StringAttribute` to `Int64Attribute`.

## Read() now implemented

`Read()` previously returned immediately. This change implements it by calling `r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_KEY_MANAGEMENT)`, unmarshalling the CM API response via `gjson`, and repopulating the Terraform state.

**Fields read from CM response** (swagger `GET /v1/vault/keys2/{id}` response):

|---|---|
| `id` | `id` |
| `name` | `name` |
| `algorithm` | `algorithm` |
| `state` | `state` |
| `usage_mask` | `usageMask` |
| `unexportable` | `unexportable` |
| `undeletable` | `undeletable` |
| `description` | `description` |
| `activation_date` | `activationDate` |
| `deactivation_date` | `deactivationDate` |
| `archive_date` | `archiveDate` |
| `process_start_date` | `processStartDate` |
| `protect_stop_date` | `protectStopDate` |
| `rotation_frequency_days` | `rotationFrequencyDays` |
| `revocation_reason` | `revocationMessage` (swapped — mirrors pre-existing bug) |
| `revocation_message` | `revocationReason` (swapped — mirrors pre-existing bug) |
| `compromise_occurrence_date` | `compromiseOccurrenceDate` |
| `curveid` | `curveid` |
| `object_type` | `objectType` |
| `key_size` | `size` |
| `xts` | `xts` |
| `cert_type` | `certType` |
| `id_size` | `idSize` |
| `destroy_date` | `destroyDate` |
| `compromise_date` | `compromiseDate` |
| `default_iv` | `defaultIV` |
| `aliases[*].alias` | `aliases[*].alias` |
| `aliases[*].index` | `aliases[*].index` |
| `aliases[*].type` | `aliases[*].type` |
| `labels` | `labels` (map of string) |
| `meta.owner_id` | `meta.owner_id` |
| `meta.permissions.*` | `meta.permissions.*` (nine lists, uppercase JSON keys) |
| `meta.cte.*` | `meta.cte.*` |
| `public_key_parameters.*` | `publicKeyParameters.*` |

**Write-only fields not populated from CM response** (preserved from prior state): `material`, `password`, `encoding`, `format`, `empty_material`, `secret_data_encoding`, `wrap_key_id_type`, `wrap_key_name`, `wrap_public_key`, `wrap_public_key_padding`, `wrap_hkdf`, `wrap_pbe`, `wrap_rsaaes`, and all other input-only creation parameters.

**404 handling:** A 404 response calls `resp.State.RemoveResource(ctx)`, removing the resource from Terraform state and causing the next plan to show a create action. This correctly handles the case where a key was deleted directly in CipherTrust Manager.

**Null/absent fields:** Absent scalar string fields default to the empty string via `gjson`'s zero-value behaviour. The `aliases` slice is set to `nil` when the response array is empty. The `labels` map defaults to an empty `types.MapValue` when absent from the response. Nested structs (`meta`, `public_key_parameters`) are only populated when the corresponding JSON object is present.

## How to test

  ```
  go test ./internal/provider/... -run 'TestAccCMKey_' -v
  ```
  Scenarios covered:
  - **Drift detection** — out-of-band attribute change is detected on the next plan (`TestAccCMKey_driftDetection`).
  - **State population** — `terraform refresh` populates all Saved=Yes scalar fields from CM (`TestAccCMKey_readPopulatesState`).
  - **Out-of-band delete** — key deleted directly in CM causes `RemoveResource`; next plan shows create (`TestAccCMKey_oobDeleteRemovesFromState`).
  - **Write-only preservation** — `material` field is not cleared by `Read()` (`TestAccCMKey_writeOnlyFieldsPreserved`).

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Read()`: `[resource_cm_key.go -> Read][uuid]` pattern.
- [ ] CM API endpoint verified against swagger spec (`GET /v1/vault/keys2/{id}` confirmed in `cm-keys` area).
- [ ] `Read()` 404 calls `resp.State.RemoveResource(ctx)` — confirmed in diff.
- [ ] Schema `aliases[*].index` type corrected from `StringAttribute` to `Int64Attribute` — matches swagger `KeyAlias.index: integer` and Go struct `types.Int64`.
- [ ] Write-only fields (`material`, `password`, all `wrap_*`) are not overwritten in `Read()` — preserved from prior state loaded at the start of the method.
- [ ] Pre-existing `revocationReason`/`revocationMessage` tag swap is mirrored (not fixed) in `Read()` — a dedicated follow-up ticket is needed for that breaking change.

---


internal/provider/cm/resource_cm_key.go
```diff
@@ -1076,6 +1077,171 @@ func (r *resourceCMKey) Create(ctx context.Context, req resource.CreateRequest,
...
+	state.ID = types.StringValue(gjson.Get(response, "id").String())
+	state.Name = types.StringValue(gjson.Get(response, "name").String())
+	state.Algorithm = types.StringValue(gjson.Get(response, "algorithm").String())
+	state.State = types.StringValue(gjson.Get(response, "state").String())
+	state.UsageMask = types.Int64Value(gjson.Get(response, "usageMask").Int())
+	state.UnExportable = types.BoolValue(gjson.Get(response, "unexportable").Bool())
+	state.UnDeletable = types.BoolValue(gjson.Get(response, "undeletable").Bool())
+	state.Description = types.StringValue(gjson.Get(response, "description").String())
+	state.ActivationDate = types.StringValue(gjson.Get(response, "activationDate").String())
+	state.DeactivationDate = types.StringValue(gjson.Get(response, "deactivationDate").String())
```
zubeensyed: @terraform-agent a lot of these fields are server genrate and that needs to be handled in this change for example archive_date               = "" -> null
              - cert_type                  = "" -> null
              - compromise_date            = "" -> null
              - compromise_occurrence_date = "" -> null
              - curveid                    = "" -> null
              - deactivation_date          = "" -> null
              - default_iv                 = "ff6aab1c39c09bca5bd1f1476ea44974" -> null
              - description                = "" -> null
              - destroy_date               = "" -> null
all these values are computed by the server, validate all fields mentioned in this implementation.

ashwani573: Fixed in `cbc02d4`: TFIN-273: address review feedback (round 1)

SarahThompson (commented):
I think it looks like good provider.

PR #187, opened 2026-06-11
TFIN-274: Drift Detection |  ciphertrust_cm_group

ashwani573:
## Summary

- Implements `Read()` in `ciphertrust_cm_group`, enabling drift detection by calling `GET api/v1/usermgmt/groups/{name}` via `c.GetById` and repopulating Terraform state from the CM API response.
- Renames the Terraform resource type from the incorrect `ciphertrust_groups` to `ciphertrust_cm_group` (fixes the `Metadata()` suffix).
- Changes `app_metadata`, `client_metadata`, and `user_metadata` schema attributes from `schema.MapNestedAttribute` / `types.Map` to `schema.StringAttribute` / `types.String`, storing metadata as raw JSON strings to support `gjson`-based hydration.
- Adds `internal/provider/cm/constants.go` with a package-private `notFoundError` sentinel and replaces the placeholder test in `resource_cm_group_test.go` with three acceptance tests covering basic create, out-of-band deletion drift, and out-of-band attribute drift.

## Motivation

Implements TFIN-274: Drift Detection | ciphertrust_cm_group.

`Read()` was previously an empty stub — it made no API call and returned without touching state. As a result, `terraform plan` would always show no changes even when the group had been modified or deleted directly in CipherTrust Manager. This change replaces the empty body with a real implementation so that out-of-band changes are detected on every `terraform plan` or `terraform refresh`.

- `Metadata()`: corrects the suffix from `"_groups"` to `"_cm_group"`.
- `Schema()`: replaces the three `schema.MapNestedAttribute` definitions (`app_metadata`, `client_metadata`, `user_metadata`) with `schema.StringAttribute{Optional: true}`. Adds `stringplanmodifier.UseStateForUnknown()` to the `id` attribute.
- `Create()` and `Update()`: replaces the `.Elements()` iteration loops (which assumed `types.Map`) with `json.Unmarshal` into `map[string]interface{}` for building the metadata payload. Corrects log messages that still referenced `resource_cm_user.go` and fixes `AddError` titles to title-case form.
- `Read()`: fully implemented — see dedicated section below.
- `Delete()`: adds a 404 fast-path (silent `return`) using `notFoundError`, making destroy idempotent when the group has already been deleted out-of-band.

### New file — `internal/provider/cm/constants.go`

- Defines `notFoundError = "status: 404"` as a package-private constant, used by `Read()` and `Delete()` to identify 404 responses.

- Changes `CMGroupTFSDK` fields `AppMetadata`, `ClientMetadata`, and `UserMetadata` from `types.Map` to `types.String` to align with the updated schema.

- Corrects all resource addresses from `ciphertrust_groups` to `ciphertrust_cm_group`.
- Replaces `TestResourceCMGroup` with `TestAccCMGroup_basicCreate`, `TestAccCMGroup_driftDetection`, and `TestAccCMGroup_attributeDrift` (details in the "How to test" section).

## Read() now implemented

`Read()` previously returned immediately without any API call. This change implements it:

1. Reads current state to obtain `resourceID` (`state.ID.ValueString()`).
2. Calls `c.GetById(ctx, id, resourceID, common.URL_GROUP)`, which issues `GET {cm_url}/api/v1/usermgmt/groups/{resourceID}`.
3. On a 404 response: emits a warning diagnostic and calls `resp.State.RemoveResource(ctx)`, so the next plan proposes recreation.
4. On any other error: emits an error diagnostic and returns without modifying state.
5. On success: repopulates state attributes from the gjson-parsed JSON response body.

**Fields read from CM response** (per swagger `GET /v1/usermgmt/groups/{name}`, `auth-users` area, `Group` definition):

| Terraform attribute | CM JSON field | Null / absent handling |
|---|---|---|
| `id` | `name` | Set to group name (see path parameter note below) |
| `name` | `name` | Always present |
| `description` | `description` | `types.StringNull()` when absent, JSON null, or empty string |
| `app_metadata` | `app_metadata` | `types.StringNull()` when absent, JSON null, or empty object |
| `client_metadata` | `client_metadata` | `types.StringNull()` when absent, JSON null, or empty object |
| `user_metadata` | `user_metadata` | `types.StringNull()` when absent, JSON null, or empty object |

**Swagger path parameter note:** The CM API path is `GET /v1/usermgmt/groups/{name}` — the path identifier is the group's `name`, not an opaque API-assigned `id`. In `Create()`, `plan.ID` is set to `plan.Name`, so `state.ID` holds the group name and `GetById` constructs the correct URL. The `Group` definition in swagger does not expose a separate top-level `id` field; the group name is the primary key.

**404 handling:** On a 404, `resp.State.RemoveResource(ctx)` is called so that the next `terraform plan` proposes recreating the group. This is the correct drift-detection behaviour for out-of-band deletion.

**Null/absent field handling:** Fields absent from the CM JSON response are set to `types.StringNull()`. Metadata fields are additionally set to null when the JSON value is `null` or the object is empty (`{}`), which prevents a perpetual diff when the user has not set a metadata attribute.

## Schema attributes

|---|---|---|---|
| `name` | `types.String` | Required | CM group name; also used as `id`; `UseStateForUnknown` |
| `description` | `types.String` | Optional | |
| `app_metadata` | `types.String` | Optional | Raw JSON string; was `types.Map` — **breaking type change** |
| `client_metadata` | `types.String` | Optional | Raw JSON string; was `types.Map` — **breaking type change** |
| `user_metadata` | `types.String` | Optional | Raw JSON string; was `types.Map` — **breaking type change** |
| `id` | `types.String` | Computed | Set to group name from CM response; `UseStateForUnknown` |

## Breaking changes

- **Terraform type rename** (`ciphertrust_groups` → `ciphertrust_cm_group`): any existing `.tf` file referencing `resource "ciphertrust_groups" ...` will fail after this change. Users must rename the resource block and run `terraform state mv ciphertrust_groups.<name> ciphertrust_cm_group.<name>`, or destroy and recreate.

- **Metadata attribute type change** (`MapNestedAttribute` → `StringAttribute`): existing `.tf` files that set `app_metadata`, `client_metadata`, or `user_metadata` as HCL map blocks must be updated to pass JSON-encoded strings.

  Before (no longer valid):
  ```hcl
  app_metadata = { env = "prod" }
  ```

  After:
  ```hcl
  app_metadata = jsonencode({ env = "prod" })
  ```

  Existing state files containing these attributes as Terraform map objects are incompatible. Affected resources must be removed from state and re-imported or re-created.

## Open issues requiring reviewer sign-off

The following discrepancies between the implementation and the approved plan remain unresolved in the current working tree. They are noted here so reviewers can decide whether to block merge or accept with an explanation:

1. **`state.ID` is set from `gjson.Get(response, "name")` not `"id"` (resource_cm_group.go line 176)**: The plan table specifies `gjson.Get(response, "id").String()`. Because the CM API's `Group` definition does not expose a separate `id` field and the path parameter is `{name}`, setting `state.ID` to `name` is likely functionally correct — but it deviates from the approved plan and must be explicitly confirmed.

2. **`description` null check includes `v.String() != ""` (line 179)**: The plan specifies only `v.Exists()` as the condition. The current code additionally skips setting `description` when the API returns an empty string, treating an explicitly empty description the same as an absent one. An empty string returned by CM will be stored as `types.StringNull()` instead of `types.StringValue("")`.

3. **Metadata null checks include `len(v.Map()) > 0` (lines 185, 191, 197)**: The plan specifies `v.Exists() && v.Type != gjson.Null`. The current code additionally rejects empty objects (`{}`), meaning a CM response with `"app_metadata": {}` will be stored as `types.StringNull()` rather than `"{}"`.

## How to test

  ```
  go test ./internal/provider/... -run TestAccCMGroup -v
  ```
  Scenarios covered:
  - **Create** (`TestAccCMGroup_basicCreate`): apply config with `name`, `description`, and `app_metadata`; assert `id` is set and attributes match; run a second plan-only step to confirm zero drift after a successful apply.
  - **Delete drift** (`TestAccCMGroup_driftDetection`): create the group; delete it out-of-band via `client.DeleteByURL`; a plan-only step must show a non-empty plan (Terraform detects the resource is gone and proposes recreation).
  - **Attribute drift** (`TestAccCMGroup_attributeDrift`): create the group; patch `description` out-of-band via `client.UpdateData`; a plan-only step must show a non-empty plan (attribute mismatch detected).
  - **Destroy** — handled automatically by the test framework at the end of each `TestCase`.

## Checklist

- [ ] Schema attribute `Description` fields are not populated — all attributes have empty description strings. This will affect `make generate` doc output. Add descriptions before merging if provider docs are part of the release criteria.
- [ ] `tflog.Trace` / `tflog.Debug` at entry/exit of every CRUD method — present. `Read()` uses `defer tflog.Debug` at entry/exit rather than explicit `Trace` calls; confirm this matches the project logging convention.
- [ ] Fresh `uuid.New().String()` at the top of each CRUD method — present in Create, Read, and Update. `Delete()` does not generate a uuid before calling `DeleteByID` (uses `state.Name` as the request correlator instead).
- [ ] `notFoundError` constant defined in `cm/constants.go` — no inlined string literals for 404 detection.
- [ ] CM API endpoint `GET /v1/usermgmt/groups/{name}` verified against swagger `operations.md` (line 50): confirmed.
- [ ] `Read()` 404 removes resource from state via `resp.State.RemoveResource(ctx)` — intentional; required for drift detection on out-of-band deletion.
- [ ] `.claude/indexes/resources.md` still lists the old type name `ciphertrust_groups` — update the index to `ciphertrust_cm_group` to keep it accurate for future contributors.
- [ ] Three open plan-vs-implementation discrepancies listed above must be resolved before merging.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-274

PR #189, opened 2026-06-11

ashwani573:
## Summary

- Implements `Read()` for the existing `ciphertrust_cm_key` resource by calling `GET /v1/vault/keys2/{id}` via `c.GetById` and hydrating all readable attributes into Terraform state.
- Fixes a pre-existing schema gap: `object_type` and `state` gain `Computed: true` and `UseStateForUnknown()` so server-assigned values no longer appear as spurious `-> null` drift after a refresh.
- Fixes a pre-existing schema type mismatch: `aliases[].index` is corrected from `schema.StringAttribute` to `schema.Int64Attribute`, matching the CM API response field type.
- Adds 404 handling to `Delete()` so out-of-band key deletions do not produce errors during post-test cleanup.
- Adds three acceptance tests covering drift detection, state hydration, and write-only field preservation.

## Motivation

Before this change `Read()` returned immediately without contacting CipherTrust Manager. Any out-of-band modification to a key (description, usage mask, undeletable flag, aliases, labels, etc.) was invisible to Terraform — `terraform plan` would always show no changes even after a human had mutated the key directly in CM. This change wires `Read()` to `GET /v1/vault/keys2/{id}` so that `terraform refresh` and subsequent plans correctly surface divergence between Terraform state and the live CM resource.

**Schema fixes**

| Attribute | Before | After | Reason |
|---|---|---|---|
| `object_type` | `Optional` only | `Computed: true, Optional: true` + `UseStateForUnknown()` | CM infers and returns this for most key types; without `Computed`, the provider plans `-> null` on every refresh for users who do not set it |
| `state` | `Optional` only | `Computed: true, Optional: true` + `UseStateForUnknown()` | CM always returns a state value (`Active`, `Pre-Active`, etc.); same issue |
| `aliases[].index` | `schema.StringAttribute` | `schema.Int64Attribute` | CM returns `index` as a JSON integer; the prior string type caused type-assertion panics when populating alias state |

**Read() implementation**

`Read()` was previously an empty stub. It now:

1. Loads current Terraform state into a `CMKeyTFSDK` struct via `req.State.Get`.
2. Generates a fresh `uuid.New().String()` trace ID and calls `r.client.GetById(ctx, traceID, state.ID.ValueString(), common.URL_KEY_MANAGEMENT)`, issuing `GET api/v1/vault/keys2/{id}` to CM.
3. Returns an error diagnostic if the response contains no `id` field.
4. Hydrates attributes back into state using `gjson` path lookups against the raw JSON response.

Hydration strategy:

- **Always hydrated** (unconditional): `id`, `name`, `algorithm` (lowercased to match config normalisation — CM returns `"AES"`, configs use `"aes"`), `size`, `curveid`, `object_type`, `state`, `description`, `labels`.
- **Conditionally hydrated** (`!state.X.IsNull()` guard — only when the user explicitly configured the attribute): `usage_mask`, `unexportable`, `undeletable`, `uuid`, `muid`, `key_id`, `activation_date`, `deactivation_date`, `archive_date`, `destroy_date`, `compromise_date`, `compromise_occurrence_date`, `process_start_date`, `protect_stop_date`, `rotation_frequency_days`, `default_iv`, `cert_type`, `xts`, `id_size`.
- **Aliases** — hydrated only when `len(state.Aliases) > 0`. CM automatically creates an alias matching the key name on every key creation; unconditionally hydrating it would inject an alias into state for any user who never declared one in their config, producing permanent drift.
- **Complex nested objects** — `meta` (with `owner_id`, nine-field `permissions` block, and `cte` sub-struct) and `public_key_parameters` (with its own aliases list) are fully traversed via gjson.
- **Write-only fields preserved from prior state** — `material`, `password`, `wrap_*`, `mac_sign_*`, `encoding`, `format`, and similar input-only fields that CM does not echo back in `GET` responses are not touched by `Read()`.
- **Revocation fields preserved from prior state** — `revocation_reason` and `revocation_message` are also left untouched. There is a pre-existing bug in `CMKeyJSON` where their `json:` struct tags are swapped, meaning serialisation/deserialisation via that struct inverts the values. Preserving prior state is the correct workaround until that is fixed in a separate ticket.

**404 handling in Read()**

If `GetById` returns an error containing `"status: 404"`, `Read()` calls `resp.State.RemoveResource(ctx)` and returns without error. This removes the resource from Terraform state so that the next `terraform plan` shows it as needing to be created — the primary mechanism by which OOB key deletion is surfaced to users.

**404 handling in Delete()**

A `status: 404` from `DeleteByID` is now silently swallowed rather than propagated as an error diagnostic. This prevents test teardown failures when a key was already deleted out-of-band during the acceptance test run.

**Imported package**

`"github.com/hashicorp/terraform-plugin-framework/attr"` added to support `map[string]attr.Value` construction when building the labels map value.

Three acceptance test functions added:

- **`TestAccCMKey_driftDetection`** (5-step test): creates a key; in step 2 mutates `description` OOB via `cmClient.UpdateData` and asserts `ExpectNonEmptyPlan: true`; re-applies original config in step 3 and asserts description is restored; asserts stable plan in step 4; deletes the key OOB via `cmClient.DeleteByURL` in step 5 and asserts `ExpectNonEmptyPlan: true` (the RemoveResource path). All key names use `uuid.NewString()[:8]` suffix to avoid 409 conflicts from stale previous runs.
- **`TestAccCMKey_readHydratesState`** (3-step test): creates a key with `description`, `usage_mask`, `undeletable`, and `object_type` configured; triggers a `RefreshState: true` step and asserts all four are populated in state plus that `state` attribute is set; asserts unconfigured optional attributes (`activation_date`, `key_id`, `cert_type`, `xts`, `id_size`) remain absent; asserts no drift on a subsequent plan.
- **`TestAccCMKey_writeOnlyPreserved`** (3-step test): creates a key with `material` and `encoding` set; triggers `RefreshState: true`; asserts both write-only fields retain their original values; asserts no drift on a subsequent plan.

New imports added to the test file: `context`, `encoding/json`, `fmt`, `common`, `uuid`, `terraform`.

### Modified file — `go.mod`

`go 1.22.0` changed to `go 1.22`. This is a cosmetic fix for local toolchain version parsing with no functional impact.

## API endpoint verification

`GET /v1/vault/keys2/{id}` is confirmed present in the swagger operations index under the `cm-keys` area. The `PatchKey` swagger definition lists all fields that are mutable via `PATCH /v1/vault/keys2/{id}`: `meta`, `unexportable`, `undeletable`, `activationDate`, `deactivationDate`, `archiveDate`, `rotationFrequencyDays`, `compromiseOccurrenceDate`, `revocationReason`, `revocationMessage`, `processStartDate`, `protectStopDate`, `aliases`, `muid`, `keyId`, `usageMask`, `labels`, `description`. Fields absent from `PatchKey` (e.g. `algorithm`, `size`, `curveid`, `name`) are immutable and carry `RequiresReplace` plan modifiers in the existing schema; they are unconditionally hydrated here so state remains accurate after any refresh.

## Read() now implemented

`Read()` previously returned immediately. This change implements it by calling `r.client.GetById(ctx, traceID, id, common.URL_KEY_MANAGEMENT)`, parsing the CM API JSON response with `gjson`, and repopulating the Terraform state.

**Fields read from CM response** (per swagger `GET /v1/vault/keys2/{id}`):

| Terraform attribute | CM JSON field | Hydration |
|---|---|---|
| `id` | `id` | Always |
| `name` | `name` | Always |
| `algorithm` | `algorithm` | Always; lowercased |
| `key_size` | `size` | Always |
| `curve_id` | `curveid` | Always |
| `object_type` | `objectType` | Always |
| `state` | `state` | Always |
| `description` | `description` | Always |
| `labels` | `labels` | Always; `MapNull` when absent |
| `usage_mask` | `usageMask` | Only when user-configured (non-null in prior state) |
| `unexportable` | `unexportable` | Only when user-configured |
| `undeletable` | `undeletable` | Only when user-configured |
| `uuid` | `uuid` | Only when user-configured |
| `muid` | `muid` | Only when user-configured |
| `key_id` | `keyId` | Only when user-configured |
| `activation_date` | `activationDate` | Only when user-configured |
| `deactivation_date` | `deactivationDate` | Only when user-configured |
| `archive_date` | `archiveDate` | Only when user-configured |
| `destroy_date` | `destroyDate` | Only when user-configured |
| `compromise_date` | `compromiseDate` | Only when user-configured |
| `compromise_occurrence_date` | `compromiseOccurrenceDate` | Only when user-configured |
| `process_start_date` | `processStartDate` | Only when user-configured |
| `protect_stop_date` | `protectStopDate` | Only when user-configured |
| `rotation_frequency_days` | `rotationFrequencyDays` | Only when user-configured |
| `default_iv` | `defaultIV` | Only when user-configured |
| `cert_type` | `certType` | Only when user-configured |
| `xts` | `xts` | Only when user-configured |
| `id_size` | `idSize` | Only when user-configured |
| `aliases` | `aliases` | Only when user configured aliases (`len > 0`) |
| `meta` | `meta` | Full nested hydration; nil when absent |
| `public_key_parameters` | `publicKeyParameters` | Full nested hydration; nil when absent |

**Write-only fields not hydrated** (CM does not return these in GET responses): `material`, `password`, `wrap_*`, `mac_sign_*`, `encoding`, `format`. `revocation_reason` and `revocation_message` are also not hydrated (pre-existing swapped JSON tags bug — see Pre-existing bugs section).

**404 handling:** A 404 from `GetById` calls `resp.State.RemoveResource(ctx)`, removing the resource from Terraform state so the next `terraform plan` shows it as to-be-created.

**Null/absent fields:** Each `gjson` extraction produces a typed null (`types.StringNull()`, `types.Int64Null()`, `types.BoolNull()`) when the field is absent from the response.

## Pre-existing bugs noted (not fixed in this PR)

1. **`aliases[].index` schema/struct type mismatch** — The schema change in this PR corrects the sub-attribute type from `StringAttribute` to `Int64Attribute`, aligning schema with the `KeyAliasTFSDK.Index types.Int64` field. Acceptance tests avoid asserting on alias index values to decouple from any residual type handling issues; a follow-up pass should verify round-trip behaviour under full alias CRUD.

2. **`CMKeyJSON` swapped revocation JSON tags** — `RevocationReason` and `RevocationMessage` in `CMKeyJSON` have their `json:` tags interchanged. Any code serialising both fields via this struct inverts them. `Read()` works around this by preserving both from prior state. A dedicated bug ticket should fix the tags and then enable hydration of these fields.

## How to test

  ```
  go test ./internal/provider/... -run TestAccCMKey_driftDetection -v
  go test ./internal/provider/... -run TestAccCMKey_readHydratesState -v
  go test ./internal/provider/... -run TestAccCMKey_writeOnlyPreserved -v
  ```
  Scenarios covered:
  - **Drift detection (OOB update)** — apply config; mutate `description` via CM API; assert non-empty plan; re-apply; assert description restored; assert stable plan.
  - **Drift detection (OOB delete)** — delete key via CM API; assert non-empty plan (RemoveResource path).
  - **State hydration** — create key; `RefreshState`; assert `description`, `usage_mask`, `undeletable`, `object_type`, `state` are hydrated; assert unconfigured optional fields are absent; assert stable plan.
  - **Write-only field preservation** — create key with `material` and `encoding`; `RefreshState`; assert both are unchanged.
- [ ] Manual smoke-test: apply a key resource, modify its `description` in the CM UI, run `terraform plan` and verify a non-empty plan is reported, run `terraform apply` to reconcile, run `terraform destroy`.

## Checklist

- [ ] Schema `Description` fields populated on all modified attributes.
- [ ] `tflog.Trace` at exit of `Read()` using the `MSG_METHOD_END` prefix and `[resource_cm_key.go -> Read][id]` pattern.
- [ ] Fresh `uuid.New().String()` generated inside `Read()` for the trace ID — not reused across calls.
- [ ] `URL_KEY_MANAGEMENT` constant used throughout `Read()` — no inlined URL string literals.
- [ ] CM API endpoint `GET /v1/vault/keys2/{id}` verified against swagger `cm-keys` area.
- [ ] `Read()` 404 calls `resp.State.RemoveResource(ctx)` to surface OOB key deletion to Terraform.
- [ ] Write-only fields (`material`, `encoding`, etc.) preserved from prior state — not nulled out by `Read()`.
- [ ] Pre-existing `CMKeyJSON` swapped revocation tags documented; revocation fields preserved from prior state as workaround.
- [ ] Pre-existing `aliases[].index` type mismatch corrected in schema (this PR); residual alias round-trip testing deferred to follow-up.

---


ashwani573:

PR #196, opened 2026-06-15

ashwani573:
## Summary

- Fixes `ciphertrust_policies` `Read()` to hydrate `conditions` and `include_descendant_accounts` from the CM API response (both were previously missing from state, making any out-of-band change to those fields invisible to Terraform).
- Fixes `ciphertrust_policy_attachments` `Read()` to hydrate `policy`, `jurisdiction`, `principal_selector`, `actions`, and `resources` from the CM API response (all five were previously missing).
- Replaces the unconditional `"Updating Policy is not supported"` error in `Update()` for both resources with working implementations that call `c.UpdateDataV2` against `URL_CM_POLICIES` and `URL_CM_POLICY_ATTACHMENTS`.
- Adds four acceptance tests covering no-drift round-trip, out-of-band drift detection, and in-place update for both resources.

## Motivation

Implements TFIN-278: Drift Detection for `ciphertrust_policy` and `ciphertrust_policy_attachments`.

Before this change, `terraform plan` always showed no changes after `terraform apply` — even when a policy or attachment was modified directly in CipherTrust Manager. This was because `Read()` did not populate the attributes that matter most for policy behaviour (`conditions`, `include_descendant_accounts`, `actions`, `resources`, etc.) from the API response. Equally, any call to `terraform apply` with a changed policy configuration would fail immediately with a hard-coded error rather than patching CM.

**Schema:**
- `include_descendant_accounts` gains `Computed: true`. It was previously `Optional`-only, which caused a plan-time error when CM returns the field after a create where the user did not set it (the provider had no way to store the server-assigned value).

**`Create()`:**
- Adds an unknown-value guard: if `IncludeDescendantAccounts` is still unknown after the POST response (CM did not echo it back), it is set to `types.BoolNull()` so state is consistent with the now-Computed schema.

**`Read()`:**
- `name`: previously always written as `types.StringValue` even when empty; now uses `types.StringNull()` when the API returns an empty string to prevent false drift against an unset Optional attribute.
- `allow`: previously called `gjson.Bool()` unconditionally (returns `false` when the key is absent, indistinguishable from an explicit `false`); now uses `gjson.Exists()` guard — `types.BoolNull()` when absent.
- `include_descendant_accounts`: newly populated from the API response using the same `Exists()` guard pattern.
- `conditions`: newly populated. Each entry reads `negate` (with `Exists()` guard), `op`, `path`, and `values` from the API response into `CMPolicyConditionTFSDK` structs.

**`Update()`:**
- Replaced the unconditional error with a full implementation: reads plan and state, builds a `CMPolicyJSON` payload from the plan values, marshals to JSON, and calls `r.client.UpdateDataV2(ctx, state.ID.ValueString(), common.URL_CM_POLICIES, payloadJSON)`. On success, computed fields (`id`, `uri`, `account`, `created_at`) are refreshed from the PATCH response.

**Swagger note:** The pre-split swagger index (`operations.md`) lists only `GET`, `POST`, and `DELETE` for `/v1/admin/policies`. No `PATCH /v1/admin/policies/{id}` entry is present. Reviewers should verify against the live CM API spec that a PATCH endpoint exists before merging. The implementation will surface an HTTP error at runtime if the endpoint does not exist; it will not cause a compile error or plan failure.

### Modified file — `internal/provider/cm/resource_policy_attachments.go`

**`Read()`:**
- `policy`: newly populated via `gjson.Get(response, "policy").String()`.
- `jurisdiction`: newly populated using `Exists()` guard — `types.StringNull()` when absent.
- `principal_selector`: newly populated by iterating `gjson.Get(response, "principalSelector").Map()` and building a `types.Map` from the key/value pairs.
- `actions`: newly populated; `types.ListNull(types.StringType)` when the API returns an empty array, `types.ListValueMust` otherwise.
- `resources`: same pattern as `actions`.

**`Update()`:**
- Replaced the unconditional error with a full implementation: builds a `CMPolicyAttachmentJSON` payload including `policy`, `principal_selector`, `jurisdiction`, `actions`, and `resources`, then calls `r.client.UpdateDataV2(ctx, state.ID.ValueString(), common.URL_CM_POLICY_ATTACHMENTS, payloadJSON)`.

**Import added:** `github.com/hashicorp/terraform-plugin-framework/attr` — required for `attr.Value` when constructing `types.Map` and `types.List` values in `Read()`.

**Swagger note:** The pre-split swagger index lists only `GET`, `POST`, and `DELETE` for `/v1/admin/policy-attachments`. No `PATCH /v1/admin/policy-attachments/{id}` entry is present. Same caveat as above — reviewers should confirm the endpoint exists in the live CM spec.

- Adds `Actions []string \`json:"actions,omitempty"\`` and `Resources []string \`json:"resources,omitempty"\`` to `CMPolicyAttachmentJSON`. These fields were missing from the struct, so the `Update()` payload for policy attachments silently dropped `actions` and `resources` even if they were set in the plan.

- **`TestAccCMPolicy_DriftConditions`** (new): creates a policy with `conditions` and `include_descendant_accounts`; verifies all attributes land in state; asserts `ExpectNonEmptyPlan: false` on a subsequent plan (no false drift); then performs an out-of-band PATCH via `createCMClient()` to shrink the conditions list, and asserts `ExpectNonEmptyPlan: true` (drift is now detected).
- **`TestAccCMPolicy_Update`** (new): create → update changing `actions`, `allow`, `effect`, `include_descendant_accounts`, and `conditions.values` → no-drift plan step after the update.

- **`TestAccCMPolicyAttachment_DriftRead`** (new): creates an attachment with all five previously-unhydrated attributes (`policy`, `jurisdiction`, `principal_selector`, `actions`, `resources`); verifies each attribute is in state; then asserts `ExpectNonEmptyPlan: false` on a subsequent plan.
- **`TestAccCMPolicyAttachment_Update`** (new): create → update expanding `actions` and `resources` lists → no-drift plan step after the update.

## Read() now implemented

`Read()` for both resources previously returned without populating key attributes. This change implements full hydration by calling `r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), URL_*)`, parsing the JSON response with `gjson`, and writing every attribute back into state.

**Fields read from CM response for `ciphertrust_policies`:**
- `id` <- `id`
- `uri` <- `uri`
- `account` <- `account`
- `created_at` <- `createdAt`
- `name` <- `name` (StringNull when empty)
- `resources` <- `resources[]`
- `actions` <- `actions[]`
- `allow` <- `allow` (BoolNull when key absent)
- `effect` <- `effect`
- `include_descendant_accounts` <- `include_descendant_accounts` (BoolNull when absent)
- `conditions[].negate` <- `conditions[].negate` (BoolNull when absent)
- `conditions[].op` <- `conditions[].op`
- `conditions[].path` <- `conditions[].path`
- `conditions[].values[]` <- `conditions[].values[]`

**Fields read from CM response for `ciphertrust_policy_attachments`:**
- `id` <- `id`
- `uri` <- `uri`
- `account` <- `account`
- `created_at` <- `createdAt`
- `policy` <- `policy`
- `jurisdiction` <- `jurisdiction` (StringNull when absent)
- `principal_selector` <- `principalSelector` (map)
- `actions` <- `actions[]` (ListNull when empty)
- `resources` <- `resources[]` (ListNull when empty)

**404 handling:** Neither `Read()` calls `resp.State.RemoveResource(ctx)` on a 404. Consistent with the provider-wide convention documented in `conventions.md` (commit `43f3b14`): a 404 keeps the resource in Terraform state so that transient CM unavailability does not silently delete managed resources. The existing error-path diagnostic reports the read failure; the resource remains in state until explicitly destroyed.

**Null/absent fields:** Optional bool and string fields that are absent from the CM response are set to their null equivalents (`types.BoolNull()`, `types.StringNull()`) rather than zero values. This prevents Terraform from seeing a diff between a config that omits the attribute and a state that contains `false` or `""`.

## Breaking changes

None. All schema changes are additive (`Computed: true` added to an existing `Optional` attribute; two new `omitempty` fields added to an internal JSON struct). No `RequiresReplace` plan modifiers were added or changed. No attributes were removed.

## How to test

  ```
  go test ./internal/provider/... -run TestAccCMPolicy_DriftConditions -v
  go test ./internal/provider/... -run TestAccCMPolicy_Update -v
  go test ./internal/provider/... -run TestAccCMPolicyAttachment_DriftRead -v
  go test ./internal/provider/... -run TestAccCMPolicyAttachment_Update -v
  ```
  Scenarios covered:
  - **Drift detection** — apply config; verify all state attributes match; do an out-of-band mutation via CM API; re-plan and confirm `ExpectNonEmptyPlan: true`.
  - **No false drift** — apply config; re-plan without any changes; confirm `ExpectNonEmptyPlan: false`.
  - **Update** — apply initial config; apply updated config (changed `actions`, `resources`, `conditions`, `effect`, `include_descendant_accounts`); verify updated values in state; re-plan and confirm no drift.
  - **Destroy** — implicit at test cleanup; `terraform destroy` removes both the policy and its attachment from CM.

## Checklist

- [ ] URL constants defined in `urls.go` — no inlined string literals (`URL_CM_POLICIES`, `URL_CM_POLICY_ATTACHMENTS` were pre-existing; no new constants required).
- [ ] CM API PATCH endpoints verified against swagger spec — the pre-split index does not list `PATCH /v1/admin/policies/{id}` or `PATCH /v1/admin/policy-attachments/{id}`; confirm against the live CM OpenAPI spec before merging.

---

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module github.com/ThalesGroup/terraform-provider-ciphertrust
 
 go 1.22.0
 
+toolchain go1.22.4
+
 require (
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/terraform-plugin-framework v1.13.0
--- a/internal/provider/cm/resource_policy.go
+++ b/internal/provider/cm/resource_policy.go
@@ -11,6 +11,7 @@ import (
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringdefault"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
@@ -82,7 +83,11 @@ func (r *resourceCMPolicy) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Description: "Specifies the effect of the policy. Possible values are 'allow', 'deny', 'obligate_on_allow', and 'obligate_on_deny'. Default is 'deny'.",
 			},
 			"include_descendant_accounts": schema.BoolAttribute{
-				Optional:    true,
+				Optional: true,
+				Computed: true,
+				PlanModifiers: []planmodifier.Bool{
+					boolplanmodifier.UseStateForUnknown(),
+				},
 				Description: "When false, only the resources in the principal's account can be accessed if the policy allows it.",
 			},
 			"name": schema.StringAttribute{
@@ -94,9 +99,24 @@ func (r *resourceCMPolicy) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Description: "Resources is a list of URI strings, which must be in URI format.",
 				ElementType: types.StringType,
 			},
-			"uri":        schema.StringAttribute{Computed: true},
-			"account":    schema.StringAttribute{Computed: true},
-			"created_at": schema.StringAttribute{Computed: true},
+			"uri": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"account": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"created_at": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
 		},
 	}
 }
@@ -193,6 +213,12 @@ func (r *resourceCMPolicy) Create(ctx context.Context, req resource.CreateReques
 	plan.URI = types.StringValue(gjson.Get(response, "uri").String())
 	plan.Account = types.StringValue(gjson.Get(response, "account").String())
 	plan.CreatedAt = types.StringValue(gjson.Get(response, "createdAt").String())
+	plan.Effect = types.StringValue(gjson.Get(response, "effect").String())
+	if gjson.Get(response, "include_descendant_accounts").Exists() {
+		plan.IncludeDescendantAccounts = types.BoolValue(gjson.Get(response, "include_descendant_accounts").Bool())
+	} else {
+		plan.IncludeDescendantAccounts = types.BoolNull()
+	}
 
 	tflog.Debug(ctx, "[resource_policy.go -> Create Output]["+response+"]")
 
@@ -229,7 +255,12 @@ func (r *resourceCMPolicy) Read(ctx context.Context, req resource.ReadRequest, r
 	state.URI = types.StringValue(gjson.Get(response, "uri").String())
 	state.Account = types.StringValue(gjson.Get(response, "account").String())
 	state.CreatedAt = types.StringValue(gjson.Get(response, "createdAt").String())
-	state.Name = types.StringValue(gjson.Get(response, "name").String())
+
+	if gjson.Get(response, "name").String() != "" {
+		state.Name = types.StringValue(gjson.Get(response, "name").String())
+	} else {
+		state.Name = types.StringNull()
+	}
 
 	arrResources := gjson.Get(response, "resources").Array()
 	var resources []types.String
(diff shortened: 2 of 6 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-278

ashwani573:
@terraform-agent The smoke tests are failing due to schema and Read() hydration issues. Fix all three:

**1. `resource_policy.go` — add `Computed: true` to `account`, `uri`, `created_at`**
These show `-> (known after apply)` because Read() sets them but the schema lacks `Computed: true`. Add it to each:
```go
"account": schema.StringAttribute{ Computed: true, ... }
"uri":     schema.StringAttribute{ Computed: true, ... }
"created_at": schema.StringAttribute{ Computed: true, ... }
```
Also add `UseStateForUnknown()` plan modifier to each so they do not show as unknown on plan.

**2. `resource_policy.go` — hydrate `include_descendant_accounts` in Read()**
The field is missing from state after Create/Read. Add to Read() hydration:
```go
if r := gjson.Get(response, "include_descendant_accounts"); r.Exists() {
    state.IncludeDescendantAccounts = types.BoolValue(r.Bool())
} else {
    state.IncludeDescendantAccounts = types.BoolNull()
}
```

**3. `resource_policy_attachments.go` — fix `jurisdiction` and `policy` hydration in Read()**
Read() is returning full server URIs for `policy` and `jurisdiction` but the config has short names, causing perpetual drift. Add `!state.X.IsNull()` guards for these fields so they are only updated when the user explicitly set them in config:
```go
if r := gjson.Get(response, "jurisdiction"); r.Exists() && !state.Jurisdiction.IsNull() {
    state.Jurisdiction = types.StringValue(r.String())
}
if r := gjson.Get(response, "policy"); r.Exists() && !state.Policy.IsNull() {
    state.Policy = types.StringValue(r.String())
}
```

Do not commit. Do not push. Just edit the files and run `make build` + `make test`.

PR #200, opened 2026-06-16
TFIN-220  CTE | Immutable-Field Handling

chinmay-thales:
**TFIN-220  CTE | Immutable-Field Handling**

**Changes**
- Added support for handling immutable fields for all the CTE resources.

- If a user attempts to update an immutable field in any CTE resource, the request will be rejected with an error indicating that the field is immutable and cannot be modified.

- In **client_guardpoint** and **clientgroup_guardpoint** resources **guard_path** field is handled differently: since it is the map key for guard_points, it isn't covered by the immutable-field rejection. Changing it is instead treated as an implicit unguard of the old path and create of the new one, shown in the plan as a remove/add rather than an in-place update or an error.


chinmay-thales:

examples/resources/cte_clientgroup/resource.tf
```diff
@@ -38,6 +38,12 @@ resource "ciphertrust_cte_client_group" "test_cg_group_test1" {
 
   description = "tf test cg.."      # Optional description for the client group
 
+  # - List of clients to be added to client group
```
rahulrampura-thales, line 41: Duplicate code at line 72. Remove the duplication

sample-scripts/cte/client-group/main.tf
```diff
@@ -18,6 +18,13 @@ resource "ciphertrust_cte_client_group" "test_cg_group_test1" {
   cluster_type = "NON-CLUSTER"
   communication_enabled = true
   description = "tf test cg.."
+
+  # - List of clients to be added to client group
+  client_list = [
```
rahulrampura-thales, line 23: Duplicate block at line 53. Remove the duplication.

PR #214, opened 2026-06-18

ashwani573:
## Summary

- Implements `Read()` for `ciphertrust_domain` by calling `GET api/v1/domains/{id}` via `c.ReadDataByParam` and repopulating all Terraform state attributes, enabling drift detection on `terraform plan`.
- Fixes `Update()` and `Delete()` to use `state.ID` (the CM-assigned UUID) as the path key, correcting 404 errors caused by the previous code sending the human-readable domain name.
- Adds 404 guards in both `Read()` and `Delete()` so that a domain already absent from CipherTrust Manager leaves Terraform state unchanged rather than returning a terminal error.
- Adds three acceptance tests — `TestAccCipherTrustCMDomain_drift`, `TestAccCipherTrustCMDomain_updatePathKey`, and `TestAccCipherTrustCMDomain_destroyOOB` — plus a `requireDomainLicenseOrSkip` helper that skips all domain tests on CM instances without the appropriate license.

## Motivation

Implements TFIN-305: Drift Detection for `ciphertrust_cm_domain`.

After the initial resource implementation, three correctness bugs prevented reliable drift detection and in-place updates:

1. `Read()` used the `gjson.Bool()` zero-value fallback for `allow_user_management`, coercing an absent field to `false` and causing false drift. The `admins` and `meta_data` fields had no else-branch, leaving stale state when the API returned neither. Log strings incorrectly named `resource_cte_client.go` throughout.
2. `Update()` passed `plan.Name` as the path key to `UpdateData`, which builds `PATCH api/v1/domains/{key}`. CM returns 404 on every update because domain names are not UUIDs. `meta_data` was also omitted from the post-update read-back, causing perpetual drift for users who set metadata.
3. `Delete()` passed `state.Name` as the path key and had no 404 guard, so `terraform destroy` failed when the domain had already been removed directly in CM.

**Read() — correctness fixes enabling drift detection**

- Adds `tflog.Trace` entry log at method start (was missing).
- Adds a `strings.Contains(err.Error(), "status: 404")` guard: on 404 the method logs and returns immediately without modifying state or surfacing a diagnostic. The resource stays in state until explicitly destroyed.
- Fixes `allow_user_management` hydration: uses `gjson.Result.Exists()` with an explicit `else` branch that sets `types.BoolNull()`, preventing a false-`false` coercion when CM omits the field from the response.
- Adds `else { state.Admins = nil }` so the admins list is cleared rather than left stale when the API returns no array.
- Adds `else { state.Meta = types.MapNull(types.StringType) }` and guards against a JSON `null` literal (`metaResult.Type != gjson.Null`) before iterating meta entries.
- Corrects two `tflog` messages that referenced the wrong source file (`resource_cte_client.go` -> `resource_cm_domain.go`).
- Corrects `AddError` title to title-case `"Error Reading CipherTrust Domain"` and fixes the detail string.

Fields populated from the `GET api/v1/domains/{id}` response (endpoint confirmed via swagger):

| Terraform attribute | CM JSON field | Null handling |
|---|---|---|
| `id` | `id` | unconditional |
| `name` | `name` | unconditional |
| `uri` | `uri` | unconditional |
| `account` | `account` | unconditional |
| `dev_account` | `devAccount` | unconditional |
| `application` | `application` | unconditional |
| `created_at` | `createdAt` | unconditional |
| `updated_at` | `updatedAt` | unconditional |
| `allow_user_management` | `allow_user_management` | `types.BoolNull()` when field absent from response |
| `hsm_connection_id` | `hsm_connection_id` | `types.StringNull()` when empty string |
| `hsm_kek_label` | `hsm_kek_label` | `types.StringNull()` when empty string |
| `parent_ca_id` | `parent_ca_id` | `types.StringNull()` when empty string |
| `admins` | `admins[]` | `nil` when absent or non-array |
| `meta_data` | `meta` | `types.MapNull(types.StringType)` when absent or JSON null |

**Update() — bug fix: path key was plan.Name, now state.ID**

- Changes `r.client.UpdateData(ctx, plan.Name.ValueString(), ...)` to `r.client.UpdateData(ctx, state.ID.ValueString(), ...)`. The CM endpoint is `PATCH api/v1/domains/{id}` where `{id}` must be the server-assigned UUID. The prior code caused a 404 on every update.
- Applies the same `gjson.Result.Exists()` guard to `allow_user_management` in the post-update read-back, consistent with the corrected `Read()`.
- Adds `meta_data` hydration in the post-update read-back (was previously missing entirely), using the same `gjson.Null` guard.
- Fixes the error log identifier and `AddError` title to use `state.ID.ValueString()` and title-case `"Error Updating CipherTrust Domain"`.

**Delete() — bug fix: path key was state.Name, now state.ID; adds 404 guard**

- Changes path key from `state.Name.ValueString()` to `state.ID.ValueString()` in both URL construction and the `DeleteByID` call. The CM endpoint is `DELETE api/v1/domains/{id}`.
- Adds a `strings.Contains(err.Error(), "status: 404")` guard: if the domain is already absent in CM, the method logs and returns without a diagnostic error, allowing `terraform destroy` to complete cleanly.
- Moves the `tflog.Trace` success log to after the error check so it fires only on a confirmed successful delete.

- Adds `requireDomainLicenseOrSkip(t *testing.T)` helper: constructs a CM client from `CIPHERTRUST_*` environment variables via the existing `createCMClient()` helper, then probes `POST api/v1/domains` with a random name. If CM returns `"Operation not permitted by current license"`, the test is skipped. On any other outcome it proceeds; if the probe domain was created successfully it is deleted best-effort using `client.DeleteByID` with the UUID extracted via `gjson.Get(resp, "id")`.
- Extends existing `TestResourceCMDomain` to call `requireDomainLicenseOrSkip(t)` before running.
- Adds `TestAccCipherTrustCMDomain_drift`: creates a domain with `allow_user_management=false`, uses a `PreConfig` to PATCH the field to `true` out-of-band via `client.UpdateData`, then runs `RefreshState: true` with `ExpectNonEmptyPlan: true` to assert that `Read()` surfaces the change.
- Adds `TestAccCipherTrustCMDomain_updatePathKey`: creates a domain without `meta_data`, applies an in-place update adding `meta_data = { "env" = "test" }` (exercises the PATCH UUID path-key fix), then runs a plan-only step with `ExpectNonEmptyPlan: false` to assert stable state.
- Adds `TestAccCipherTrustCMDomain_destroyOOB`: creates a domain, deletes it out-of-band via `client.DeleteByID` in a `PreConfig`, then runs `Destroy: true` to verify the 404 guard in `Delete()` allows destroy to complete without error.

## Read() now implemented

`Read()` was present but returned incorrect state due to the defects described above. This change makes it fully correct: it calls `r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), common.URL_DOMAIN)` and maps every API response field to Terraform state as described in the table above.

**404 handling:** A 404 response from CM causes `Read()` to return immediately without modifying state. `resp.State.RemoveResource(ctx)` is NOT called. Removing on 404 would silently drop a resource from state if CM is temporarily unreachable; the resource is considered still present until an explicit `terraform destroy`.

**Null/absent fields:** The `allow_user_management` bool uses `gjson.Result.Exists()` rather than the raw `.Bool()` zero-value fallback, so a CM instance that omits the field does not generate false drift. Optional string fields that return an empty value are stored as `types.StringNull()`. The `meta` map guards against both absence and a JSON `null` literal before iterating.

## How to test

  ```
  go test ./internal/provider/ -run 'TestResourceCMDomain|TestAccCipherTrustCMDomain' -v -timeout 15m
  ```

  Scenarios covered:

  - **Create/Update/Destroy** (`TestResourceCMDomain`) — create with `meta_data`; update metadata; destroy.
  - **Drift detection** (`TestAccCipherTrustCMDomain_drift`) — create domain, mutate `allow_user_management` out-of-band, refresh state, assert non-empty plan.
  - **Update path key** (`TestAccCipherTrustCMDomain_updatePathKey`) — create domain, add `meta_data` via in-place update (PATCH must use UUID), assert no drift after apply.
  - **Out-of-band destroy** (`TestAccCipherTrustCMDomain_destroyOOB`) — create domain, delete it directly in CM, run `terraform destroy`, assert no error.
  - All four tests skip automatically on CM instances without a domain license.

## Checklist

- [ ] Schema attribute `Description` fields are populated — required for `make generate` docs.
- [ ] `tflog.Trace` at entry/exit of CRUD methods now uses the correct `[resource_cm_domain.go -> <Func>][uuid]` pattern — stale `resource_cte_client.go` references corrected in this PR.
- [ ] URL constant `URL_DOMAIN = "api/v1/domains"` already defined in `urls.go` — no inlined string literals added.
- [ ] CM API endpoints verified against swagger: `GET api/v1/domains/{id}`, `PATCH api/v1/domains/{id}`, `DELETE api/v1/domains/{id}` all confirmed; `{id}` is the server-assigned UUID, not the resource name.

---

### DIFF
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -248,11 +248,6 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 	if err != nil {
 		if strings.Contains(err.Error(), notFoundError) {
 			tflog.Debug(ctx, "[resource_cm_domain.go -> Read] domain not found (404), leaving state unchanged["+id+"]")
-			resp.Diagnostics.AddWarning(
-				"CipherTrust Domain Not Found",
-				"Domain "+state.ID.ValueString()+" was not found in CM and has been kept in Terraform state. "+
-					"If it was intentionally deleted, run `terraform state rm` before the next apply.",
-			)
 			return
 		}
 		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_domain.go -> Read]["+id+"]")
@@ -309,7 +304,7 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 		}
 		state.Admins = admins
 	} else {
-		state.Admins = []types.String{}
+		state.Admins = nil
 	}
 
 	// Read meta_data map
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -169,7 +169,7 @@ func (r *resourceCMDomain) Create(ctx context.Context, req resource.CreateReques
 
 	payloadJSON, err := json.Marshal(payload)
 	if err != nil {
-		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_group.go -> Create]["+id+"]")
+		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_domain.go -> Create]["+id+"]")
 		resp.Diagnostics.AddError(
 			"Invalid data input: Domain Creation",
 			err.Error(),
@@ -179,7 +179,7 @@ func (r *resourceCMDomain) Create(ctx context.Context, req resource.CreateReques
 
 	response, err := r.client.PostDataV2(ctx, id, common.URL_DOMAIN, payloadJSON)
 	if err != nil {
-		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_group.go -> Create]["+id+"]")
+		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_domain.go -> Create]["+id+"]")
 		resp.Diagnostics.AddError(
 			"Error creating domain on CipherTrust Manager: ",
 			"Could not create domain, unexpected error: "+err.Error(),
@@ -193,7 +193,11 @@ func (r *resourceCMDomain) Create(ctx context.Context, req resource.CreateReques
 	plan.CreatedAt = types.StringValue(gjson.Get(response, "createdAt").String())
 	plan.UpdatedAt = types.StringValue(gjson.Get(response, "updatedAt").String())
 	plan.Account = types.StringValue(gjson.Get(response, "account").String())
-	plan.AllowUserManagement = types.BoolValue(gjson.Get(response, "allow_user_management").Bool())
+	if aum := gjson.Get(response, "allow_user_management"); aum.Exists() {
+		plan.AllowUserManagement = types.BoolValue(aum.Bool())
+	} else {
+		plan.AllowUserManagement = types.BoolNull()
+	}
 
 	// Handle optional fields - set to null if empty string to avoid inconsistent state
 	hsmConnectionIdResp := gjson.Get(response, "hsm_connection_id").String()
@@ -232,6 +236,7 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 	var state CMDomainTFSDK
 	id := uuid.New().String()
 	tflog.Trace(ctx, common.MSG_METHOD_START+"[resource_cm_domain.go -> Read]["+id+"]")
+	defer tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_cm_domain.go -> Read]["+id+"]")
 
 	diags := req.State.Get(ctx, &state)
 	resp.Diagnostics.Append(diags...)
@@ -241,8 +246,13 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 
 	response, err := r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), common.URL_DOMAIN)
 	if err != nil {
-		if strings.Contains(err.Error(), "status: 404") {
+		if strings.Contains(err.Error(), notFoundError) {
 			tflog.Debug(ctx, "[resource_cm_domain.go -> Read] domain not found (404), leaving state unchanged["+id+"]")
+			resp.Diagnostics.AddWarning(
+				"CipherTrust Domain Not Found",
+				"Domain "+state.ID.ValueString()+" was not found in CM and has been kept in Terraform state. "+
+					"If it was intentionally deleted, run `terraform state rm` before the next apply.",
+			)
 			return
 		}
 		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_domain.go -> Read]["+id+"]")
@@ -299,15 +309,15 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 		}
 		state.Admins = admins
 	} else {
-		state.Admins = nil
+		state.Admins = []types.String{}
 	}
 
 	// Read meta_data map
 	metaResult := gjson.Get(response, "meta")
 	if metaResult.Exists() && metaResult.Type != gjson.Null {
-		metaMap := make(map[string]types.String)
+		metaMap := make(map[string]string)
 		metaResult.ForEach(func(key, value gjson.Result) bool {
-			metaMap[key.String()] = types.StringValue(value.String())
+			metaMap[key.String()] = value.String()
 			return true
 		})
 		mapValue, diags2 := types.MapValueFrom(ctx, types.StringType, metaMap)
@@ -320,7 +330,6 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 		state.Meta = types.MapNull(types.StringType)
 	}
 
-	tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_cm_domain.go -> Read]["+id+"]")
 	// Set refreshed state
 	diags = resp.State.Set(ctx, &state)
 	resp.Diagnostics.Append(diags...)
@@ -360,21 +369,14 @@ func (r *resourceCMDomain) Update(ctx context.Context, req resource.UpdateReques
 		hasChanges = true
 	}
 
-	// Check metadata
-	if !plan.Meta.Equal(state.Meta) {
+	// Check allow_user_management
+	if !plan.AllowUserManagement.Equal(state.AllowUserManagement) {
 		hasChanges = true
 	}
 
-	// Check admins list
-	if len(plan.Admins) != len(state.Admins) {
+	// Check metadata
+	if !plan.Meta.Equal(state.Meta) {
 		hasChanges = true
-	} else {
-		for i := range plan.Admins {
-			if plan.Admins[i].ValueString() != state.Admins[i].ValueString() {
-				hasChanges = true
-				break
-			}
-		}
 	}
 
 	// If no changes detected, preserve existing state and return
@@ -391,6 +393,10 @@ func (r *resourceCMDomain) Update(ctx context.Context, req resource.UpdateReques
 	}
 	if plan.HSMConnectionId.ValueString() != "" && plan.HSMConnectionId.ValueString() != types.StringNull().ValueString() {
 		payload.HSMConnectionId = plan.HSMConnectionId.ValueString()
+	}
+	if !plan.AllowUserManagement.IsNull() && !plan.AllowUserManagement.IsUnknown() {
+		val := plan.AllowUserManagement.ValueBool()
+		payload.AllowUserManagement = &val
 	}
 	metadataPayload := make(map[string]interface{})
 	for k, v := range plan.Meta.Elements() {
@@ -467,9 +473,9 @@ func (r *resourceCMDomain) Update(ctx context.Context, req resource.UpdateReques
 
 	metaReadBack := gjson.Get(readResponse, "meta")
 	if metaReadBack.Exists() && metaReadBack.Type != gjson.Null {
-		metaMap := make(map[string]types.String)
+		metaMap := make(map[string]string)
 		metaReadBack.ForEach(func(key, value gjson.Result) bool {
-			metaMap[key.String()] = types.StringValue(value.String())
+			metaMap[key.String()] = value.String()
 			return true
 		})
 		mapValue, diags2 := types.MapValueFrom(ctx, types.StringType, metaMap)
@@ -503,7 +509,7 @@ func (r *resourceCMDomain) Delete(ctx context.Context, req resource.DeleteReques
 	url := fmt.Sprintf("%s/%s/%s", r.client.CipherTrustURL, common.URL_DOMAIN, state.ID.ValueString())
 	output, err := r.client.DeleteByID(ctx, "DELETE", state.ID.ValueString(), url, nil)
 	if err != nil {
-		if strings.Contains(err.Error(), "status: 404") {
+		if strings.Contains(err.Error(), notFoundError) {
 			tflog.Debug(ctx, "[resource_cm_domain.go -> Delete] domain not found (404), treating as already deleted["+state.ID.ValueString()+"]")
 			return
 		}
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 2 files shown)


ashwani573:

ashwani573:
@terraform-agent

## Code Review — TFIN-305

**Overall: Good work — 2 real bugs to fix before merging.**

---

### ✅ Correct changes

- `Update()` path key: `state.ID.ValueString()` ✓
- `Delete()` path key: `state.ID.ValueString()` ✓
- `Delete()` 404 guard ✓
- `allow_user_management` `r.Exists()` absent-branch → `types.BoolNull()` in both `Read()` and `Update()` ✓
- `meta_data` absent/null → `types.MapNull(types.StringType)` in `Read()` ✓
- `meta_data` hydration added to `Update()` post-read-back ✓
- Error title case fixed ✓
- Log filename corrected from `resource_cte_client.go` ✓
- `requireDomainLicenseOrSkip` helper cleanly handles 409 license errors ✓
- `TestAccCipherTrustCMDomain_destroyOOB` correctly uses closure to capture ID ✓

---

### ❌ Bug 1 — `state.Admins = nil` should be `[]types.String{}`

```go
} else {
    state.Admins = nil   // ← wrong
}
```

A `nil` slice serialises as a **null list** in Terraform state. If the API omits `admins` and the config has `admins = ["admin"]`, Terraform sees `null → ["admin"]` as perpetual drift. Fix:

```go
} else {
    state.Admins = []types.String{}
}
```

---

### ❌ Bug 2 — `meta_data` in `Update()` uses `map[string]types.String` instead of `map[string]string`

```go
metaMap := make(map[string]types.String)   // ← should be map[string]string
```

`types.MapValueFrom` with element type `types.StringType` expects `map[string]string`. Passing `map[string]types.String` can cause a runtime reflection error. The same issue exists in `Read()`. Both should use:

```go
metaMap := make(map[string]string)
metaReadBack.ForEach(func(key, value gjson.Result) bool {
    metaMap[key.String()] = value.String()
    return true
})
```

---

### ⚠️ Minor — 404 in `Read()` returns silently without a warning diagnostic

```go
    tflog.Debug(ctx, "...leaving state unchanged...")
    return
}
```

Keeping the resource in state is correct per project conventions (`conventions.md`, commit `43f3b14`). However, the operator gets no visibility. Consider adding:

```go
resp.Diagnostics.AddWarning(
    "CipherTrust Domain Not Found",
    "Domain "+state.ID.ValueString()+" was not found in CM and has been kept in Terraform state. "+
    "If it was intentionally deleted, run `terraform state rm` before the next apply.",
)
```

---

### ⚠️ Minor — `MSG_METHOD_END` not deferred in `Read()`

The end-of-Read trace is placed manually at the bottom and does not fire on the 404 early-return path. Should be:

```go
id := uuid.New().String()
tflog.Trace(ctx, common.MSG_METHOD_START+"[resource_cm_domain.go -> Read]["+id+"]")
defer tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_cm_domain.go -> Read]["+id+"]")
```

---

### Summary

| Check | Status |
|---|---|
| Update()/Delete() path key fixes | ✅ |
| `allow_user_management` drift fix | ✅ |
| `meta_data` absent-branch in Read() | ✅ |
| Delete() 404 guard | ✅ |
| Tests: drift, update, destroy-OOB | ✅ |
| **`admins` nil vs empty slice** | ❌ fix required |
| **`meta_data` map type in Update()/Read()** | ❌ fix required |
| 404 warning diagnostic in Read() | ⚠️ minor |
| `defer` MSG_METHOD_END in Read() | ⚠️ minor |

ashwani573:
@terraform-agent

## Additional Review — TFIN-305

### 🔴 Critical — Must Fix Before Merge

**1. Use `notFoundError` constant instead of inline string literals**

`cm/constants.go` already declares `const notFoundError = "status: 404"`. The PR introduces two raw string duplicates:

```go
// Read() and Delete() — both wrong:

// Should be:
```

Two instances to fix: Read() and Delete().

---

**2. Create() still uses bare `.Bool()` for `allow_user_management` — causes Read-after-Create drift**

Read() and Update() were correctly fixed to use `.Exists()`, but Create() at line 196 is untouched:

```go
// Create() — NOT fixed:
plan.AllowUserManagement = types.BoolValue(gjson.Get(response, "allow_user_management").Bool())
```

If CM omits `allow_user_management` from the create response, `gjson.Get(...).Bool()` returns `false` and stores it in state. The next `terraform plan` calls Read(), which now correctly returns `types.BoolNull()`. Terraform sees `false → null` and shows spurious drift. Fix:

```go
if aum := gjson.Get(response, "allow_user_management"); aum.Exists() {
    plan.AllowUserManagement = types.BoolValue(aum.Bool())
} else {
    plan.AllowUserManagement = types.BoolNull()
}
```

---

**3. Update() `hasChanges` block silently ignores `allow_user_management` changes**

The `hasChanges` guard checks HSM fields, meta, and admins — but not `AllowUserManagement`. A config change from `false` to `true` on this field alone will hit the early-return path and never reach the API. The change is silently dropped.

```go
// Missing from hasChanges block:
if !plan.AllowUserManagement.Equal(state.AllowUserManagement) {
    hasChanges = true
}
```

This is directly connected to the drift detection this PR is supposed to fix.

---

### 🟡 Significant — Should Fix

**4. Update() detects admins changes but never sends them in the payload**

`hasChanges` compares `plan.Admins` vs `state.Admins` (lines 355–363) which triggers an API call, but the payload builder only populates `HSMKEKLabel`, `HSMConnectionId`, and `Meta`. Admins are detected as changed but silently dropped from the PATCH body. Either add `payload.Admins` if the CM API supports updating admins via PATCH, or remove the admins comparison from `hasChanges`.

---

**5. No drift tests for `meta_data` or `admins`**

Three tests cover: `allow_user_management` drift, update path-key, OOB destroy. Given the PR title is "Drift Detection", consider adding:
- OOB change to `meta_data` via API → surfaces as drift in Read()
- OOB change to `admins` → drift test (would also expose issue #4 above)

---

### 🟢 Minor / Nits

**6. Read() 404 diverges from `resource_cm_group.go` (inconsistency, not a bug)**

The group resource calls `resp.State.RemoveResource(ctx)` on 404; this PR correctly uses `return` (keep in state per `conventions.md`). The domain PR implements the right behavior, but the inconsistency within the `cm/` package should be noted in the PR description and `resource_cm_group.go` filed as a follow-up.

**7. `TestResourceCMDomain` (existing) does not assert `allow_user_management` or `meta_data` values**

After this PR fixes hydration, worth adding:
```go
resource.TestCheckResourceAttr("ciphertrust_domain.testDomain", "allow_user_management", "false"),
resource.TestCheckResourceAttr("ciphertrust_domain.testDomain", "meta_data.abc", "xyz"),
```

**8. Create() error logs still reference wrong file**

Two `tflog.Debug` calls in Create() still log `[resource_cm_group.go -> Create]`. Should be `resource_cm_domain.go`.

**9. `requireDomainLicenseOrSkip` probe cleanup is best-effort**

The probe creates a real domain and deletes it with `_, _ = client.DeleteByID(...)`. A comment noting the intentional best-effort cleanup would help:
```go
// best-effort cleanup — a leaked probe domain is benign
_, _ = client.DeleteByID(...)
```

ashwani573:
Addressed in `562fb5c`: TFIN-305: address review feedback (round 1)

rohitpandita-hub:
@terraform-agent following tests are failing, FIX them
TestAccCipherTrustCMDomain_updatePathKey
TestResourceCMDomain
TestAccCipherTrustCMDomain_drift
TestAccCipherTrustCMDomain_destroyOOB

ashwani573:
Addressed in `80beac6`: TFIN-305: address review feedback (round 2)

ashwani573:

### 🔴 HIGH — always blocking
- The plan explicitly states `updated_at` is 'Computed (no UseStateForUnknown)', but the diff adds `stringplanmodifier.UseStateForUnknown()` to `updated_at` — this is an unplanned schema change that alters provider API surface behaviour.
- In `Update()`, the diff removes the `admins` list change-detection block (length comparison plus element-by-element loop) without any plan justification; if `admins` changes in config, `hasChanges` will remain false and the update will be silently skipped, causing missed updates.

- In `Read()`, the `admins` absent-branch does not set `state.Admins = nil` as the plan requires; instead the implementation preserves prior state for both empty-array and absent cases, meaning a Required field removed server-side will never be reflected as nil in state.
- In `Read()`, the `meta_data` empty-object branch keeps prior state rather than setting `state.Meta = types.MapNull(types.StringType)` as the plan specifies; only the absent/null case correctly sets MapNull, so a server-side removal of all metadata keys returning an empty object will not be reflected as null in state.

ashwani573:

- In Read(), the `admins` absent-branch correctly sets `state.Admins = nil`, but the empty-array branch (when `adminsResult.Exists() && adminsResult.IsArray()` is true but `len(arr) == 0`) falls through without setting `state.Admins`, leaving stale prior state. The plan requires `else { state.Admins = nil }` for the absent case only; an empty array from the server should set `state.Admins` to an empty/nil slice, not preserve prior state. Add `state.Admins = nil` (or `state.Admins = []types.String{}`) in the `len(arr) == 0` branch inside the `if adminsResult.Exists() && adminsResult.IsArray()` block.
- In Update()'s post-update read-back, the empty-object branch for `meta_data` (when `metaReadBack.Exists() && metaReadBack.Type != gjson.Null` is true but `len(metaMap) == 0`) falls through with a comment 'keep plan.Meta from user config', leaving `plan.Meta` as the user-configured value rather than reflecting the server's actual empty-object state. The plan specifies setting `types.MapNull(types.StringType)` in the absent/null case; an empty object returned by CM should also set `plan.Meta = types.MapNull(types.StringType)` for consistency with Read().

ashwani573:
Addressed in `57e7657`: TFIN-305: address review feedback (round 2)

ashwani573:

### 🔴 HIGH — always blocking
- The `notFoundError` constant is used in the diff (`strings.Contains(err.Error(), notFoundError)`) but is never defined or imported in the shown diff for `internal/provider/cm/resource_cm_domain.go`. If it is not already defined in a peer file in the same package visible in the diff context, this will cause a compile error. The plan specifies the literal string `"status: 404"` — verify that `notFoundError` is a package-level constant equal to `"status: 404"` in the `cm` package; if not, define it or use the literal directly.

- In `Update()` post-update read-back, when `metaReadBack` exists, is non-null, but `len(metaMap) == 0` (empty object returned by CM), the code falls through without setting `plan.Meta`, leaving it as whatever was in the plan config rather than setting it to `types.MapNull(types.StringType)`. The plan states 'empty object — keep plan.Meta from user config' in Update() but also specifies the absent-branch sets null; the empty-object case in Update() silently differs from Read() which sets null for empty maps, causing potential drift. Consider aligning both to set `types.MapNull` for empty maps, or explicitly document the intentional divergence.
- In `Update()`, `allow_user_management` is added to the change-detection block (`if !plan.AllowUserManagement.Equal(state.AllowUserManagement) { hasChanges = true }`) and to the payload — these additions are not described in the plan's Update() section (which only lists three corrections: path key fix, allow_user_management hydration in read-back, and meta_data absent-branch in read-back). Adding `allow_user_management` and `admins` to the payload is unplanned schema/behaviour changes that may alter what gets PATCH-ed to CM.
- The test file uses `ciphertrust_domain` as the Terraform resource type name (e.g. `resource "ciphertrust_domain" "test"`), but the existing `TestResourceCMDomain` test in the same file also uses `ciphertrust_domain`. Verify this matches the registered resource type string in the provider — if the actual registration uses `ciphertrust_cm_domain` or any other name, all test configs and the `resourceName` constant reference the wrong type, causing test failures.

ashwani573:

### 🔴 HIGH — always blocking
- The diff uses `notFoundError` as a constant (e.g. `strings.Contains(err.Error(), notFoundError)`) in `internal/provider/cm/resource_cm_domain.go`, but no definition of `notFoundError` appears in the diff or in any visible peer file in the `cm` package. The plan specifies the literal string `"status: 404"` — either define `const notFoundError = "status: 404"` in the `cm` package or use the literal directly to prevent a compile error.

- The test file uses `"ciphertrust_domain"` as the Terraform resource type name (e.g. `resource "ciphertrust_domain" "test"`) in all three new test functions, but the plan names the resource `ciphertrust_cm_domain` and the `resourceName` constant is set to `"ciphertrust_domain.test"`. If the provider registers this resource as `ciphertrust_cm_domain` (as suggested by the plan's attribute table header and the file name `resource_cm_domain.go`), all new tests will fail at runtime with an unknown resource type error. Verify the registered type string and update the test configs accordingly.
- In `Read()`, the `admins` absent-branch does NOT set `state.Admins = nil` as the plan requires; instead, the code silently preserves prior state when the API returns an absent or empty array. The plan explicitly states `else { state.Admins = nil }` for the absent case, and `nil` (null list) must be set rather than leaving the prior value, so that Terraform can detect drift when admins are removed out-of-band.

PR #217, opened 2026-06-19

ashwani573:
## Summary

- Fixes drift detection for the `ciphertrust_domain` resource by correcting Read(), Create(), Update(), and Delete() in `internal/provider/cm/resource_cm_domain.go`.
- Removes `boolplanmodifier.UseStateForUnknown()` from the `allow_user_management` schema attribute — this modifier was silently preventing the attribute from being re-read after updates (HIGH finding from code review).
- Fixes Update() and Delete() to route via `state.ID.ValueString()` (the CM UUID) rather than the resource name, eliminating guaranteed 404 errors on every in-place update and every destroy.
- Extends `internal/provider/resource_cm_domain_test.go` with three new acceptance tests: `TestAccCMDomain_drift`, `TestAccCMDomain_deleteOOB`, and `TestAccCMDomain_updateUsesID`.

## Motivation

Implements TFIN-305: Drift Detection for `ciphertrust_domain`.

Prior to this change, `terraform refresh` and `terraform plan` would not detect changes made directly in CipherTrust Manager. Several independent bugs caused silent state corruption, guaranteed 404 errors on update and delete, and incorrect attribute types in state. All bugs are fixed together because they share a single root cause: the CRUD methods were written against incorrect assumptions about the CM API response shape and URL routing contract.

**Schema**

- Removed `boolplanmodifier.UseStateForUnknown()` from `allow_user_management`. This modifier is only appropriate for attributes whose server-assigned value is stable after creation (such as `id`). `allow_user_management` is a user-settable Optional+Computed field; keeping the modifier caused Terraform to freeze the attribute at its creation-time value and never re-read it during plan or refresh, which is the opposite of drift detection. This was flagged as a HIGH finding during code review.

**Create()**

- `allow_user_management`: replaced unconditional `types.BoolValue(gjson.Get(response, "allow_user_management").Bool())` with a `gjson.Result.Exists()` guard. When CM omits the field from the POST response, state is now set to `types.BoolNull()` rather than `false`, preventing a spurious diff on the first subsequent `terraform plan`.
- `admins`: added hydration from the POST response. When the server returns a non-empty `admins` array, the planned value is overwritten with the server-confirmed list so that state order matches what CM actually stored. When CM omits or returns an empty list, the planned value is preserved.

**Read()**

- `allow_user_management`: same `gjson.Result.Exists()` guard as Create() — sets `types.BoolNull()` when the field is absent from the GET response.
- `meta_data`: corrected the intermediate map type from `map[string]types.String` to `map[string]string`. The former type caused a runtime framework error when passed to `types.MapValueFrom` because the framework expects the native Go element type, not the wrapped type. Added an `else` branch that sets `state.Meta = types.MapNull(types.StringType)` when the CM response returns an absent or empty `meta` object, so that clearing metadata out-of-band is reflected in state rather than being silently ignored.

**Update()**

- Path key bug: `r.client.UpdateData(ctx, plan.Name.ValueString(), ...)` replaced with `r.client.UpdateData(ctx, state.ID.ValueString(), ...)`. The CM API routes domain mutations via `PATCH api/v1/domains/{uuid}`; using the human-readable name caused a 404 on every update without exception.
- `allow_user_management` and `parent_ca_id` added to the `hasChanges` detection block. Without these comparisons, changes to either field left `hasChanges = false` and the PATCH was silently skipped.
- `allow_user_management`, `parent_ca_id` (with an explicit empty string `""` when clearing, to signal the clear to the server), and `admins` added to the PATCH payload builder so they are sent to CM when they change.
- Post-update readback: applies the same `gjson.Result.Exists()` guard for `allow_user_management` and the corrected `map[string]string` + `types.MapNull` logic for `meta_data`, consistent with Read().

**Delete()**

- Path key bug: `state.Name.ValueString()` replaced with `state.ID.ValueString()` in both the URL construction and the `DeleteByID` argument. The CM API requires the UUID, not the name.
- 404 guard added: when `DeleteByID` returns an error containing `"status: 404"`, the function returns cleanly without adding a diagnostic. This prevents `terraform destroy` from failing when the domain was already deleted out-of-band.

Three new acceptance test functions added to the existing file. Each is gated on `RequireCM(t)` and uses `createCMClient()` for out-of-band operations:

- `TestAccCMDomain_drift`: Creates a domain, patches `allow_user_management` and `meta_data` directly via the CM API (bypassing Terraform), uses `RefreshState: true` to confirm Read() surfaces both changes as a non-empty plan, re-applies the original config to confirm Update() corrects the drift, then asserts a clean plan with `PlanOnly: true, ExpectNonEmptyPlan: false`.
- `TestAccCMDomain_deleteOOB`: Creates a domain, deletes it via `client.DeleteByID` out-of-band, uses `RefreshState: true` to confirm Read() detects the 404 and removes the resource from state so that plan shows a re-create rather than an error.
- `TestAccCMDomain_updateUsesID`: Creates a domain, changes `meta_data`, and re-applies via Terraform — confirms end-to-end that Update() succeeds. A name-based URL path would return 404 and fail the test immediately, making this an implicit regression guard for the path-key fix.

## Read() now implemented

Read() was already present and calling the CM API (`ReadDataByParam` → `GET api/v1/domains/{id}`). This PR fixes the correctness bugs within that existing implementation that prevented accurate drift detection.

**Fields read from CM response** (per swagger `GET /v1/domains/{id}`):

| Terraform attribute | CM JSON field | Null handling |
|---|---|---|
| `id` | `id` | — |
| `name` | `name` | — |
| `allow_user_management` | `allow_user_management` | `types.BoolNull()` when field absent |
| `meta_data` | `meta` | `types.MapNull(types.StringType)` when absent or empty |
| `admins` | `admins` | preserved as-is from array |
| `hsm_connection_id` | `hsm_connection_id` | `types.StringNull()` when empty string |
| `hsm_kek_label` | `hsm_kek_label` | `types.StringNull()` when empty string |
| `parent_ca_id` | `parent_ca_id` | `types.StringNull()` when empty string |
| `uri` | `uri` | — |
| `dev_account` | `devAccount` | — |
| `application` | `application` | — |
| `created_at` | `createdAt` | — |
| `updated_at` | `updatedAt` | — |
| `account` | `account` | — |

**404 handling:** A 404 response from `ReadDataByParam` triggers `resp.State.RemoveResource(ctx)` and issues an `AddWarning` diagnostic. When CM confirms the domain is gone, removing it from state allows Terraform to plan a re-create on the next apply rather than erroring. Delete() uses a separate 404 guard that returns silently — the two paths have different semantics; Read() needs to communicate the absence to the plan engine, while Delete() is idempotent.

**Null/absent fields:** The null assignments for `allow_user_management` and `meta_data` are what enable drift detection for these fields. Without them, prior state is preserved and out-of-band changes to these fields are invisible to Terraform.

## How to test

- [ ] Acceptance tests (require a live CM — set `TF_ACC=1` and `CIPHERTRUST_*` env vars):
  ```
  go test ./internal/provider/ -run 'TestAccCMDomain_drift|TestAccCMDomain_deleteOOB|TestAccCMDomain_updateUsesID' -v -timeout 15m
  ```
  Scenarios covered:
  - **Drift detection** — OOB patch to `allow_user_management` and `meta_data`; `RefreshState` surfaces both as a non-empty plan; re-apply corrects drift; clean plan follows.
  - **OOB deletion** — OOB delete via CM client; `RefreshState` removes resource from state; plan shows re-create.
  - **Update uses UUID path** — `meta_data` change applied successfully via `PATCH /v1/domains/{uuid}`; confirms name-based path regression is absent.
  - **No spurious drift** — `PlanOnly: true, ExpectNonEmptyPlan: false` step present in each test after the final apply.

## Checklist

- [ ] URL constant used from `urls.go` — no inlined string literals (`URL_DOMAIN = "api/v1/domains"` was already present; no new constant required).
- [ ] CM API endpoints verified against swagger spec (`POST /v1/domains`, `GET /v1/domains/{id}`, `PATCH /v1/domains/{id}`, `DELETE /v1/domains/{id}` confirmed in the `auth-users` swagger area).
- [ ] `boolplanmodifier.UseStateForUnknown()` removed from `allow_user_management` — the modifier was preventing re-reads after updates and is only correct for server-assigned stable identifiers such as `id`.
- [ ] `Read()` 404 removes resource from state via `resp.State.RemoveResource(ctx)` with a warning diagnostic.
- [ ] `Delete()` 404 silently succeeds — correct for idempotent destroy when domain was already removed out-of-band.

---

### DIFF
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -12,7 +12,6 @@ import (
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
-	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
@@ -68,9 +67,6 @@ func (r *resourceCMDomain) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Optional:    true,
 				Computed:    true,
 				Description: "To allow user creation and management in the domain, set it to true. The default value is false.",
-				PlanModifiers: []planmodifier.Bool{
-					boolplanmodifier.UseStateForUnknown(),
-				},
 			},
 			"hsm_connection_id": schema.StringAttribute{
 				Optional:    true,
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -196,7 +196,24 @@ func (r *resourceCMDomain) Create(ctx context.Context, req resource.CreateReques
 	plan.CreatedAt = types.StringValue(gjson.Get(response, "createdAt").String())
 	plan.UpdatedAt = types.StringValue(gjson.Get(response, "updatedAt").String())
 	plan.Account = types.StringValue(gjson.Get(response, "account").String())
-	plan.AllowUserManagement = types.BoolValue(gjson.Get(response, "allow_user_management").Bool())
+	allowUserMgmtResp := gjson.Get(response, "allow_user_management")
+	if allowUserMgmtResp.Exists() {
+		plan.AllowUserManagement = types.BoolValue(allowUserMgmtResp.Bool())
+	} else {
+		plan.AllowUserManagement = types.BoolNull()
+	}
+	// Hydrate admins from POST response so stored state matches server-confirmed order.
+	// Only override when the server returns a non-empty list; otherwise keep the
+	// planned value to avoid "element vanished" consistency errors when CM omits admins
+	// from the POST response.
+	respAdminsResult := gjson.Get(response, "admins")
+	if respAdminsResult.Exists() && len(respAdminsResult.Array()) > 0 {
+		var respAdmins []types.String
+		for _, v := range respAdminsResult.Array() {
+			respAdmins = append(respAdmins, types.StringValue(v.String()))
+		}
+		plan.Admins = respAdmins
+	}
 
 	// Handle optional fields - set to null if empty string to avoid inconsistent state
 	hsmConnectionIdResp := gjson.Get(response, "hsm_connection_id").String()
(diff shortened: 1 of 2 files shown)


ashwani573:

PR #218, opened 2026-06-19
TFIN-296: Drift Detection | ciphertrust_password_policy

ashwani573:
## Summary

- Fixes perpetual drift in `ciphertrust_password_policy` by replacing unconditional `types.Int64Value(gjson.Get(...).Int())` assignments in `Read()` and `Create()` with `!state.X.IsNull()`-guarded hydration for all nine Optional Int64 attributes and the `failed_logins_lockout_thresholds` list.
- Adds a 404 guard to `Delete()` so `terraform destroy` succeeds cleanly when a non-global policy has already been removed outside Terraform.
- Adds three acceptance tests to `internal/provider/resource_password_policy_test.go` covering drift detection, no-default-drift, and out-of-band deletion.

## Motivation

Implements TFIN-296: Drift Detection | ciphertrust_password_policy.

`ciphertrust_password_policy` stores Optional Int64 attributes (e.g. `inclusive_min_digits`, `password_lifetime`) that the user may leave unconfigured (null in Terraform state). The CM API always returns a numeric value for these fields — typically `0` when not explicitly set. Before this fix, `Read()` and `Create()` wrote that server-supplied `0` back to state unconditionally. On the next `terraform plan`, Terraform compared `0` (in state) against `null` (in config) and reported a perpetual, spurious diff on every refresh. The same issue affected `failed_logins_lockout_thresholds` when the user did not configure it.

**Read() — null-guard hydration for Optional Int64 fields**

Previously, all nine Optional Int64 attributes were hydrated unconditionally:

```go
state.InclusiveMaxTotalLength = types.Int64Value(gjson.Get(response, "inclusive_max_total_length").Int())
// ... repeated for all nine fields
```

The fixed pattern skips hydration when the field is null in prior state, preventing the CM-default `0` from being written into state for unconfigured attributes:

```go
if !state.InclusiveMaxTotalLength.IsNull() {
    if r := gjson.Get(response, "inclusive_max_total_length"); r.Exists() {
        state.InclusiveMaxTotalLength = types.Int64Value(r.Int())
    } else {
        state.InclusiveMaxTotalLength = types.Int64Null()
    }
}
```

This pattern is applied to all nine fields: `inclusive_max_total_length`, `inclusive_min_digits`, `inclusive_min_lower_case`, `inclusive_min_other`, `inclusive_min_total_length`, `inclusive_min_upper_case`, `password_change_min_days`, `password_history_threshold`, and `password_lifetime`.

For `failed_logins_lockout_thresholds`, an outer `if state.FailedLoginsLockoutThresholds != nil` guard was added around the list-hydration block, so an unconfigured list (nil in state) is not overwritten with an empty or server-default list from the API response.

**Create() — equivalent null-guard hydration before resp.State.Set**

The CM API PATCH/POST response includes the full policy object with server defaults populated. Immediately after `resp.State.Set` was called with the plan, CM's returned `0` values could overwrite null plan attributes in state. The same `!plan.X.IsNull()` guard pattern was added for all nine fields, and the same `if plan.FailedLoginsLockoutThresholds != nil` guard was added for the list, so state after `terraform apply` matches the plan exactly — not the API response defaults.

**Delete() — 404 guard for out-of-band deletion**

```go
    return
}
```

Added before the `resp.Diagnostics.AddError` call inside the non-global policy deletion branch. Without this guard, `terraform destroy` would fail with an error if the policy had been deleted directly in CipherTrust Manager between the last apply and the destroy.

Three acceptance tests were added (extending the existing file):

- `TestAccCipherTrustPasswordPolicy_drift` — creates a policy with all nine numeric fields and `failed_logins_lockout_thresholds` configured, then patches all fields out-of-band via `client.UpdateDataV2`, then performs `RefreshState` and asserts `ExpectNonEmptyPlan: true`. Verifies that `Read()` correctly surfaces out-of-band changes.

- `TestAccCipherTrustPasswordPolicy_noDefaultDrift` — creates a policy with only `policy_name` (all numeric fields null), then performs `RefreshState` and asserts `ExpectNonEmptyPlan: false`. Verifies that CM's numeric defaults (`0`) do not create a spurious diff for unconfigured attributes.

- `TestAccCipherTrustPasswordPolicy_oobDelete` — creates a policy, deletes it out-of-band via `client.DeleteByID`, then runs `PlanOnly: true` with `ExpectNonEmptyPlan: true`. Verifies that `Read()` handles the 404 cleanly (removes resource from state) so Terraform plans a recreation rather than erroring.

## Read() now implemented (prior behaviour and fix)

`Read()` was already calling the CM API via `r.client.ReadDataByParam`. The bug was in the hydration logic after the API call: all Optional Int64 fields were unconditionally overwritten from the API response regardless of whether the user had configured them.

The fix applies the provider-wide convention for Optional+Computed server-auto-populated fields: hydrate the field only when `!state.X.IsNull()` so that unconfigured (null) attributes remain null after a refresh, matching the user's configuration.

**Fields hydrated from CM response** (guarded by `!state.X.IsNull()`):

|---|---|
| `inclusive_max_total_length` | `inclusive_max_total_length` |
| `inclusive_min_digits` | `inclusive_min_digits` |
| `inclusive_min_lower_case` | `inclusive_min_lower_case` |
| `inclusive_min_other` | `inclusive_min_other` |
| `inclusive_min_total_length` | `inclusive_min_total_length` |
| `inclusive_min_upper_case` | `inclusive_min_upper_case` |
| `password_change_min_days` | `password_change_min_days` |
| `password_history_threshold` | `password_history_threshold` |
| `password_lifetime` | `password_lifetime` |
| `failed_logins_lockout_thresholds` | `failed_logins_lockout_thresholds` (guarded by `!= nil`) |

**404 handling:** A 404 from CM calls `resp.State.RemoveResource(ctx)` (via the existing path already present in `Read()` before this change). This removes the resource from state so Terraform plans a recreation. The new 404 guard in `Delete()` is distinct: it silences the error when the resource is already gone from CM during a destroy operation.

## How to test

  ```
  go test ./internal/provider/ -run TestAccCipherTrustPasswordPolicy_drift -v -timeout 15m
  go test ./internal/provider/ -run TestAccCipherTrustPasswordPolicy_noDefaultDrift -v -timeout 15m
  go test ./internal/provider/ -run TestAccCipherTrustPasswordPolicy_oobDelete -v -timeout 15m
  ```

  Scenarios covered:

  - **Drift detection** — apply with all nine fields set; patch all fields out-of-band; `RefreshState` must show a non-empty plan.
  - **No-default-drift** — apply with only `policy_name`; `RefreshState` must show an empty plan (CM defaults must not bleed into state).
  - **Out-of-band delete** — apply; delete out-of-band; plan must show recreation without error (404 guard in Read() and Delete() both exercised).

## Checklist

- [ ] Schema attribute `Description` fields populated — no changes to schema in this PR.
- [ ] `tflog.Trace` at entry/exit of every CRUD method — existing logging unchanged.
- [ ] Fresh `uuid.New().String()` at the top of each CRUD method — unchanged.
- [ ] URL constant `URL_CM_PASSWORD_POLICY` already defined in `urls.go` — no inlined string literals introduced.
- [ ] CM API endpoint `api/v1/usermgmt/pwdpolicies` verified against swagger spec (operations.md lines 54-60): `GET`, `POST`, `PATCH`, `DELETE` all confirmed.
- [ ] No hand-edited files under `docs/` — none modified.

---

### DIFF
--- a/internal/provider/cm/resource_password_policy.go
+++ b/internal/provider/cm/resource_password_policy.go
@@ -235,6 +235,84 @@ func (r *resourceCMPasswordPolicy) Create(ctx context.Context, req resource.Crea
 	plan.ID = types.StringValue(passwordPolicyName)
 	plan.Name = types.StringValue(passwordPolicyName)
 
+	if !plan.InclusiveMaxTotalLength.IsNull() {
+		if r := gjson.Get(response, "inclusive_max_total_length"); r.Exists() {
+			plan.InclusiveMaxTotalLength = types.Int64Value(r.Int())
+		} else {
+			plan.InclusiveMaxTotalLength = types.Int64Null()
+		}
+	}
+	if !plan.InclusiveMinDigits.IsNull() {
+		if r := gjson.Get(response, "inclusive_min_digits"); r.Exists() {
+			plan.InclusiveMinDigits = types.Int64Value(r.Int())
+		} else {
+			plan.InclusiveMinDigits = types.Int64Null()
+		}
+	}
+	if !plan.InclusiveMinLowerCase.IsNull() {
+		if r := gjson.Get(response, "inclusive_min_lower_case"); r.Exists() {
+			plan.InclusiveMinLowerCase = types.Int64Value(r.Int())
+		} else {
+			plan.InclusiveMinLowerCase = types.Int64Null()
+		}
+	}
+	if !plan.InclusiveMinOther.IsNull() {
+		if r := gjson.Get(response, "inclusive_min_other"); r.Exists() {
+			plan.InclusiveMinOther = types.Int64Value(r.Int())
+		} else {
+			plan.InclusiveMinOther = types.Int64Null()
+		}
+	}
+	if !plan.InclusiveMinTotalLength.IsNull() {
+		if r := gjson.Get(response, "inclusive_min_total_length"); r.Exists() {
+			plan.InclusiveMinTotalLength = types.Int64Value(r.Int())
+		} else {
+			plan.InclusiveMinTotalLength = types.Int64Null()
+		}
+	}
+	if !plan.InclusiveMinUpperCase.IsNull() {
+		if r := gjson.Get(response, "inclusive_min_upper_case"); r.Exists() {
+			plan.InclusiveMinUpperCase = types.Int64Value(r.Int())
+		} else {
+			plan.InclusiveMinUpperCase = types.Int64Null()
+		}
+	}
+	if !plan.PasswordChangeMinDays.IsNull() {
+		if r := gjson.Get(response, "password_change_min_days"); r.Exists() {
+			plan.PasswordChangeMinDays = types.Int64Value(r.Int())
+		} else {
+			plan.PasswordChangeMinDays = types.Int64Null()
+		}
+	}
+	if !plan.PasswordHistoryThreshold.IsNull() {
+		if r := gjson.Get(response, "password_history_threshold"); r.Exists() {
+			plan.PasswordHistoryThreshold = types.Int64Value(r.Int())
+		} else {
+			plan.PasswordHistoryThreshold = types.Int64Null()
+		}
+	}
+	if !plan.PasswordLifetime.IsNull() {
+		if r := gjson.Get(response, "password_lifetime"); r.Exists() {
+			plan.PasswordLifetime = types.Int64Value(r.Int())
+		} else {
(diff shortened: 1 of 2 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-296

ashwani573:
## 🔍 Automated Code Review — TFIN-296

PR #220, opened 2026-06-19
TFIN-304: Drift Detection |  ciphertrust_cm_log_forwarder, ciphertrust_cm_property

ashwani573:
## Summary

- Completes `Read()` in `internal/provider/cm/resource_log_forwarder.go` to hydrate `updated_at` and all three nested parameter blocks (`elasticsearch_params`, `loki_params`, `syslog_params`) from the CM API response (`GET /v1/configs/log-forwarders/{id}`) on every refresh cycle, enabling drift detection for `ciphertrust_log_forwarder`.
- Adds an HTTP 404 guard to `Read()` that calls `resp.State.RemoveResource(ctx)` when CM returns 404, so resources deleted outside of Terraform are removed from state rather than leaving stale entries.
- Fixes a copy-paste bug in `Delete()` where `URL_CM_POLICIES` was used instead of `URL_CM_LOG_FORWARDS`, causing every `terraform destroy` on a log forwarder to target the wrong API endpoint.
- Adds a 404 guard to `Delete()` so `terraform destroy` exits cleanly when the resource was already removed out-of-band.
- Adds six acceptance tests in `internal/provider/resource_cm_log_forwarder_test.go` covering per-type drift detection, `updated_at` hydration, out-of-band destroy, and full destroy lifecycle.

## Motivation

Implements TFIN-304: Drift Detection for `ciphertrust_log_forwarder`.

Before this change, `Read()` hydrated only top-level scalar fields (`id`, `name`, `type`, `connection_id`, `account`, `created_at`) and never populated `updated_at` or any of the nested forwarder-type parameter blocks. Out-of-band changes to `elasticsearch_params`, `loki_params`, or `syslog_params` were therefore completely invisible to `terraform plan`. Additionally, `Delete()` contained a copy-paste mistake that sent the destroy request to the access-policy endpoint instead of the log-forwarder endpoint, causing every `terraform destroy` to fail silently or target the wrong resource.

**Read() — extended to support drift detection:**

- Adds `state.UpdatedAt = types.StringValue(gjson.Get(response, "updatedAt").String())`. The field was declared `Computed: true` in the schema but was never populated in `Read()`, meaning its value never changed after the initial `Create`.
- Removes a duplicate `state.Name` assignment that appeared at the end of the existing scalar-field block (the name was already set earlier in the same block; the line was replaced by the `updated_at` assignment).
- Adds a 404 guard immediately after `ReadDataByParam` returns an error: if the error string contains `"status: 404"`, the method emits a warning diagnostic, calls `resp.State.RemoveResource(ctx)`, and returns. All other errors still surface as error diagnostics.
- Adds conditional hydration for all three nested forwarder-type blocks using `gjson.Get(...).Exists()` guards:
  - `elasticsearch_params` block — reads `elasticsearch_params.indices.{activity_kmip,activity_nae,client_audit_records,server_audit_records}` as `types.String`. The whole block pointer is set to `nil` when `elasticsearch_params` is absent from the response.
  - `loki_params` block — reads `loki_params.labels.{activity_kmip,activity_nae,client_audit_records,server_audit_records}` as `types.String`. The whole block pointer is set to `nil` when absent.
  - `syslog_params` block — reads `syslog_params.syslog_params.{activity_kmip,activity_nae,client_audit_records,server_audit_records}` as `types.Bool`. The whole block pointer is set to `nil` when absent.
  - Each individual leaf field uses an `.Exists()` check: present in the response yields a typed value; absent yields a typed null (`types.StringNull()` or `types.BoolNull()`). This ensures server-side field removals are reflected in state rather than preserved from the prior read.

**Delete() — bug fix and 404 guard:**

- Replaces `common.URL_CM_POLICIES` with `common.URL_CM_LOG_FORWARDS` in the URL construction. This was a copy-paste mistake that caused every `terraform destroy` on a log forwarder to issue `DELETE {base}/api/v1/admin/policies/{id}` instead of `DELETE {base}/api/v1/configs/log-forwarders/{id}`.
- Adds a `notFoundError` (`"status: 404"`, defined in `cm/constants.go`) guard after `DeleteByID` returns an error. When matched, the method returns immediately without adding an error diagnostic, allowing `terraform destroy` to succeed cleanly when the resource was deleted out-of-band.

Swagger verification: `GET /v1/configs/log-forwarders/{id}`, `PATCH /v1/configs/log-forwarders/{id}`, and `DELETE /v1/configs/log-forwarders/{id}` are all confirmed present in the cm-system area of the swagger spec. The URL constant `URL_CM_LOG_FORWARDS = "api/v1/configs/log-forwarders"` (defined in `common/urls.go`) correctly maps to these endpoints. No new URL constants were added.

### New file — `internal/provider/resource_cm_log_forwarder_test.go`

Six new acceptance test functions, all operating on `ciphertrust_log_forwarder` (the resource type name is not changed by this PR):

- `TestAccCMLogForwarder_elasticsearchDrift` — creates an elasticsearch log forwarder with all four `elasticsearch_params.indices` fields, then mutates all four out-of-band via `client.UpdateDataV2(ctx, capturedID, URL_CM_LOG_FORWARDS, payload)`. A `RefreshState` step with `ExpectNonEmptyPlan` verifies that `Read()` detects the drift. Skipped unless `CIPHERTRUST_ES_LOG_FORWARDER_CONN_ID` is set.
- `TestAccCMLogForwarder_lokiDrift` — same pattern for `loki_params.labels.*`. Skipped unless `CIPHERTRUST_LOKI_LOG_FORWARDER_CONN_ID` is set.
- `TestAccCMLogForwarder_syslogDrift` — flips all four `syslog_params.syslog_params.*` boolean fields from `true` to `false` out-of-band. `RefreshState` with `ExpectNonEmptyPlan` verifies the drift is surfaced. Uses `requireLogForwarderConnID(t)` from the existing `resource_log_forwarder_test.go` in the same package.
- `TestAccCMLogForwarder_updatedAtRefresh` — applies a minimal syslog forwarder, asserts `updated_at` is non-empty after create, then does a `RefreshState` step and asserts `updated_at` is still non-empty with no plan diff.
- `TestAccCMLogForwarder_destroyOOB` — captures the resource ID, deletes it out-of-band via `client.DeleteByURL(ctx, uuid, URL_CM_LOG_FORWARDS+"/"+capturedID)`, then does `RefreshState`. Verifies that `Read()` calls `RemoveResource` on 404 so the subsequent implicit framework destroy succeeds cleanly.
- `TestAccCMLogForwarder_destroyLifecycle` — applies a minimal syslog forwarder and allows the framework to run an implicit `terraform destroy` at test end. Verifies the `URL_CM_LOG_FORWARDS` fix in `Delete()` by confirming the destroy succeeds.

Out-of-band mutation calls in the test correctly use `client.UpdateDataV2(ctx, capturedID, common.URL_CM_LOG_FORWARDS, payloadJSON)` — `capturedID` as the first argument (the resource ID that `UpdateDataV2` appends to the URL) and `URL_CM_LOG_FORWARDS` as the second argument (the base endpoint without the ID). This matches the `UpdateDataV2` argument contract used throughout the provider.

Helper functions `createCMClient()`, `checkStep(t, label, ...checks)`, and `RequireCM(t)` are all defined in the existing `provider` package test files and are reused here without modification.

## Read() now implemented

`Read()` previously hydrated only six top-level scalar fields. This change extends it to fully repopulate all supported Terraform state attributes from the CM API response.

**Fields now read from CM** (`GET /v1/configs/log-forwarders/{id}`):

| Terraform attribute | CM JSON path |
|---|---|
| `id` | `id` |
| `name` | `name` |
| `type` | `type` |
| `connection_id` | `connection_id` |
| `account` | `account` |
| `created_at` | `createdAt` |
| `updated_at` | `updatedAt` (added in this PR) |
| `elasticsearch_params.indices.activity_kmip` | `elasticsearch_params.indices.activity_kmip` |
| `elasticsearch_params.indices.activity_nae` | `elasticsearch_params.indices.activity_nae` |
| `elasticsearch_params.indices.client_audit_records` | `elasticsearch_params.indices.client_audit_records` |
| `elasticsearch_params.indices.server_audit_records` | `elasticsearch_params.indices.server_audit_records` |
| `loki_params.labels.activity_kmip` | `loki_params.labels.activity_kmip` |
| `loki_params.labels.activity_nae` | `loki_params.labels.activity_nae` |
| `loki_params.labels.client_audit_records` | `loki_params.labels.client_audit_records` |
| `loki_params.labels.server_audit_records` | `loki_params.labels.server_audit_records` |
| `syslog_params.syslog_params.activity_kmip` | `syslog_params.syslog_params.activity_kmip` |
| `syslog_params.syslog_params.activity_nae` | `syslog_params.syslog_params.activity_nae` |
| `syslog_params.syslog_params.client_audit_records` | `syslog_params.syslog_params.client_audit_records` |
| `syslog_params.syslog_params.server_audit_records` | `syslog_params.syslog_params.server_audit_records` |

**404 handling:** A 404 response calls `resp.State.RemoveResource(ctx)` and emits a warning diagnostic. Terraform will plan a re-create on the next `terraform plan`, which is the correct behaviour for a resource that no longer exists on the server.

**Null/absent fields:** Each nested block pointer is set to `nil` when the top-level key is absent from the CM response, so the Terraform state correctly reflects that no forwarder parameters of that type are configured. Each individual leaf field within a present block is set to `types.StringNull()` or `types.BoolNull()` when absent from the response, so field-level removals on the server side are also reflected correctly.

## How to test

- [ ] Acceptance tests (requires a live CM instance — set `TF_ACC=1` and the `CIPHERTRUST_*` env vars):
  ```
  go test ./internal/provider/ -run 'TestAccCMLogForwarder' -v -timeout 15m
  ```
  Environment variables:
  - `CIPHERTRUST_LOG_FORWARDER_CONNECTION_ID` — required for syslog-type tests (`TestAccCMLogForwarder_syslogDrift`, `TestAccCMLogForwarder_updatedAtRefresh`, `TestAccCMLogForwarder_destroyOOB`, `TestAccCMLogForwarder_destroyLifecycle`).
  - `CIPHERTRUST_ES_LOG_FORWARDER_CONN_ID` — required for `TestAccCMLogForwarder_elasticsearchDrift`.
  - `CIPHERTRUST_LOKI_LOG_FORWARDER_CONN_ID` — required for `TestAccCMLogForwarder_lokiDrift`.

  Scenarios covered:
  - **Elasticsearch drift** — all four index fields mutated out-of-band; `RefreshState` must show a non-empty plan.
  - **Loki drift** — all four label fields mutated out-of-band; `RefreshState` must show a non-empty plan.
  - **Syslog drift** — all four boolean forwarding flags flipped out-of-band; `RefreshState` must show a non-empty plan.
  - **`updated_at` refresh** — `updated_at` is populated after create and remains set after a `RefreshState` with no plan diff.
  - **Out-of-band destroy** — resource deleted directly in CM; `RefreshState` must remove it from state without a provider error.
  - **Destroy lifecycle** — standard framework-implicit destroy must succeed (validates the `URL_CM_LOG_FORWARDS` fix in `Delete()`).

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_log_forwarder.go -> <Func>][uuid]` pattern.
- [ ] URL constant `URL_CM_LOG_FORWARDS` used throughout — no inlined path string literals.
- [ ] CM API endpoints verified against swagger (`GET`, `PATCH`, `DELETE /v1/configs/log-forwarders/{id}` confirmed in `cm-system` area).
- [ ] `Read()` 404 removes resource from state — `resp.State.RemoveResource(ctx)` called on 404.

---

### DIFF
--- a/internal/provider/cm/resource_log_forwarder.go
+++ b/internal/provider/cm/resource_log_forwarder.go
@@ -310,7 +310,106 @@ func (r *resourceCMLogForwarders) Read(ctx context.Context, req resource.ReadReq
 	state.ConnectionID = types.StringValue(gjson.Get(response, "connection_id").String())
 	state.Account = types.StringValue(gjson.Get(response, "account").String())
 	state.CreatedAt = types.StringValue(gjson.Get(response, "createdAt").String())
-	state.Name = types.StringValue(gjson.Get(response, "name").String())
+	state.UpdatedAt = types.StringValue(gjson.Get(response, "updatedAt").String())
+
+	if gjson.Get(response, "elasticsearch_params").Exists() {
+		var esBlock CMLogForwardersESTFSDK
+		if gjson.Get(response, "elasticsearch_params.indices").Exists() {
+			var esIndices CMLogForwardersESOrLokiParamsTFSDK
+			if r := gjson.Get(response, "elasticsearch_params.indices.activity_kmip"); r.Exists() {
+				esIndices.ActivityKMIP = types.StringValue(r.String())
+			} else {
+				esIndices.ActivityKMIP = types.StringNull()
+			}
+			if r := gjson.Get(response, "elasticsearch_params.indices.activity_nae"); r.Exists() {
+				esIndices.ActivityNAE = types.StringValue(r.String())
+			} else {
+				esIndices.ActivityNAE = types.StringNull()
+			}
+			if r := gjson.Get(response, "elasticsearch_params.indices.client_audit_records"); r.Exists() {
+				esIndices.ClientAuditRecords = types.StringValue(r.String())
+			} else {
+				esIndices.ClientAuditRecords = types.StringNull()
+			}
+			if r := gjson.Get(response, "elasticsearch_params.indices.server_audit_records"); r.Exists() {
+				esIndices.ServerAuditRecords = types.StringValue(r.String())
+			} else {
+				esIndices.ServerAuditRecords = types.StringNull()
+			}
+			esBlock.Indices = &esIndices
+		} else {
+			esBlock.Indices = nil
+		}
+		state.ElasticsearchParams = &esBlock
+	} else {
+		state.ElasticsearchParams = nil
+	}
+
+	if gjson.Get(response, "loki_params").Exists() {
+		var lokiBlock CMLogForwardersLokiTFSDK
+		if gjson.Get(response, "loki_params.labels").Exists() {
+			var lokiLabels CMLogForwardersESOrLokiParamsTFSDK
+			if r := gjson.Get(response, "loki_params.labels.activity_kmip"); r.Exists() {
+				lokiLabels.ActivityKMIP = types.StringValue(r.String())
+			} else {
+				lokiLabels.ActivityKMIP = types.StringNull()
+			}
+			if r := gjson.Get(response, "loki_params.labels.activity_nae"); r.Exists() {
+				lokiLabels.ActivityNAE = types.StringValue(r.String())
+			} else {
+				lokiLabels.ActivityNAE = types.StringNull()
+			}
+			if r := gjson.Get(response, "loki_params.labels.client_audit_records"); r.Exists() {
+				lokiLabels.ClientAuditRecords = types.StringValue(r.String())
+			} else {
+				lokiLabels.ClientAuditRecords = types.StringNull()
+			}
+			if r := gjson.Get(response, "loki_params.labels.server_audit_records"); r.Exists() {
+				lokiLabels.ServerAuditRecords = types.StringValue(r.String())
+			} else {
+				lokiLabels.ServerAuditRecords = types.StringNull()
+			}
+			lokiBlock.Labels = &lokiLabels
+		} else {
+			lokiBlock.Labels = nil
+		}
+		state.LokiParams = &lokiBlock
+	} else {
+		state.LokiParams = nil
+	}
+
+	if gjson.Get(response, "syslog_params").Exists() {
+		var syslogBlock CMLogForwardersSyslogTFSDK
+		if gjson.Get(response, "syslog_params.syslog_params").Exists() {
+			var syslogInner CMLogForwardersSyslogParamsTFSDK
+			if r := gjson.Get(response, "syslog_params.syslog_params.activity_kmip"); r.Exists() {
+				syslogInner.ActivityKMIP = types.BoolValue(r.Bool())
+			} else {
+				syslogInner.ActivityKMIP = types.BoolNull()
+			}
+			if r := gjson.Get(response, "syslog_params.syslog_params.activity_nae"); r.Exists() {
+				syslogInner.ActivityNAE = types.BoolValue(r.Bool())
+			} else {
+				syslogInner.ActivityNAE = types.BoolNull()
+			}
+			if r := gjson.Get(response, "syslog_params.syslog_params.client_audit_records"); r.Exists() {
+				syslogInner.ClientAuditRecords = types.BoolValue(r.Bool())
+			} else {
+				syslogInner.ClientAuditRecords = types.BoolNull()
+			}
+			if r := gjson.Get(response, "syslog_params.syslog_params.server_audit_records"); r.Exists() {
+				syslogInner.ServerAuditRecords = types.BoolValue(r.Bool())
+			} else {
+				syslogInner.ServerAuditRecords = types.BoolNull()
+			}
+			syslogBlock.SyslogParams = &syslogInner
+		} else {
+			syslogBlock.SyslogParams = nil
+		}
+		state.SyslogParams = &syslogBlock
+	} else {
+		state.SyslogParams = nil
+	}
 
 	tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_log_forwarder.go -> Read]["+id+"]")
 	// Set refreshed state
@@ -446,10 +545,13 @@ func (r *resourceCMLogForwarders) Delete(ctx context.Context, req resource.Delet
 	}
 
 	// Delete existing order
-	url := fmt.Sprintf("%s/%s/%s", r.client.CipherTrustURL, common.URL_CM_POLICIES, state.ID.ValueString())
+	url := fmt.Sprintf("%s/%s/%s", r.client.CipherTrustURL, common.URL_CM_LOG_FORWARDS, state.ID.ValueString())
 	output, err := r.client.DeleteByID(ctx, "DELETE", state.ID.ValueString(), url, nil)
 	tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_log_forwarder.go -> Delete]["+state.ID.ValueString()+"]["+output+"]")
 	if err != nil {
+		if strings.Contains(err.Error(), notFoundError) {
+			return
+		}
 		resp.Diagnostics.AddError(
 			"Error Deleting CM Log Forwarder",
 			"Could not delete Log Forwarder, unexpected error: "+err.Error(),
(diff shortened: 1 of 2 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-304 (attempt 1)

### 🔴 HIGH — always blocking
- The test file `internal/provider/resource_cm_log_forwarder_test.go` uses `const res = "ciphertrust_log_forwarder.test"` and all HCL configs reference `resource "ciphertrust_log_forwarder" "test"` instead of `resource "ciphertrust_cm_log_forwarder" "test"`. The previous HIGH issue about resource type rename was flagged because the old test file was being modified; here the new test file uses the OLD resource type name `ciphertrust_log_forwarder` throughout. If the Terraform resource type is registered as `ciphertrust_cm_log_forwarder`, the tests will fail at runtime because the resource type string does not match the provider registration. If the resource type is registered as `ciphertrust_log_forwarder`, then the previous HIGH rename issue still exists. The new test file must use whichever type string is actually registered in the provider — if it is `ciphertrust_cm_log_forwarder`, all occurrences of `ciphertrust_log_forwarder` in `resource_cm_log_forwarder_test.go` must be updated to `ciphertrust_cm_log_forwarder`.
- `TestAccCMLogForwarder_destroyOOB` still calls `client.DeleteByURL(context.Background(), uuid.NewString(), common.URL_CM_LOG_FORWARDS+"/"+capturedID)` — the previous HIGH issue flagged this call because the plan specifies `client.DeleteByID(ctx, "DELETE", capturedID, common.URL_CM_LOG_FORWARDS+"/"+capturedID, nil)`. If `DeleteByURL` does not exist on the client type or has a different signature, this will fail to compile; replace with `client.DeleteByID(context.Background(), "DELETE", capturedID, common.URL_CM_LOG_FORWARDS+"/"+capturedID, nil)`.
- All four tests that call `createCMClient()` (no arguments, returning `(client, bool)`) — `TestAccCMLogForwarder_elasticsearchDrift`, `TestAccCMLogForwarder_lokiDrift`, `TestAccCMLogForwarder_syslogDrift`, `TestAccCMLogForwarder_destroyOOB` — were flagged in the previous review as potentially mismatching the actual helper signature `createCMClient(t)` taking `*testing.T`; the diff still uses `createCMClient()` with no arguments. Verify the actual signature of `createCMClient` in the provider package and update the calls accordingly if the signature requires `t`.
- `checkStep` is called as `checkStep(t, "Step N: ...", ...)` with `t *testing.T` and a string description as the first two arguments throughout the new test file; the previous review flagged that if the existing `checkStep` helper in the provider package does not accept `(t *testing.T, string, ...resource.CheckFunc)` this will fail to compile. The diff still uses this calling convention — verify the actual signature of `checkStep` in the provider package.

- `TestAccCMLogForwarder_elasticsearchDrift` and `TestAccCMLogForwarder_lokiDrift` call `client.UpdateDataV2(context.Background(), capturedID, common.URL_CM_LOG_FORWARDS, payloadJSON)` passing the bare URL constant as the URL argument instead of the full per-resource URL `common.URL_CM_LOG_FORWARDS+"/"+capturedID`; this will target the collection endpoint rather than the specific resource, causing the out-of-band mutation to fail or modify the wrong resource.
- `TestAccCMLogForwarder_syslogDrift` calls `client.UpdateDataV2(context.Background(), capturedID, common.URL_CM_LOG_FORWARDS, payloadJSON)` with the bare constant instead of the full per-resource URL `common.URL_CM_LOG_FORWARDS+"/"+capturedID`, same as the elasticsearch and loki drift tests — the out-of-band mutation will target the wrong endpoint.
- `TestAccCMLogForwarder_syslogDrift`, `TestAccCMLogForwarder_updatedAtRefresh`, `TestAccCMLogForwarder_destroyOOB`, and `TestAccCMLogForwarder_destroyLifecycle` call `requireLogForwarderConnID(t)` which is defined in the pre-existing `resource_log_forwarder_test.go` file; the previous review flagged this as a dependency that may not be accessible or may conflict. The new test file still relies on this helper without defining or re-exporting it locally, which is acceptable only if both files share the same package (`provider`) and compile together — verify this is the case and that `requireLogForwarderConnID` is indeed accessible from the existing test file.

PR #222, opened 2026-06-19
TFIN-306: AWS connection update fails with IsRoleAnywhere cannot be updated error when updating any field

ashwani573:
## Summary

- Fixes `ciphertrust_aws_connection` `Update()` failing with a 422 "IsRoleAnywhere cannot be updated" error when any field is changed, by converting `AWSConnectionModelJSON` value-type fields to pointer types with `omitempty` so unset fields are omitted from PATCH/POST bodies.
- Fixes a secondary bug where `plan.ID` was overwritten with the `updatedAt` timestamp after every successful update, corrupting subsequent Terraform operations on the resource.
- Adds `IsRoleAnywhereImmutableModifier` to the `is_role_anywhere` attribute schema to produce a clear plan-time error if a caller attempts to change it after creation (the CM API rejects this at the API layer with 422).
- Applies `NameImmutableModifier` to the `name` attribute in `resource_aws_connection.go` (the modifier already existed for other connection types but was missing from this resource).

## Motivation

Implements TFIN-306: AWS connection update fails with "IsRoleAnywhere cannot be updated" error when updating any field.

### Modified file — `internal/provider/connections/schema_connections.go`

- Converts 10 fields in `AWSConnectionModelJSON` from plain value types to pointer types with `omitempty` json tags: `Name *string`, `Description *string`, `AccessKeyID *string`, `AssumeRoleARN *string`, `AssumeRoleExternalID *string`, `AWSRegion *string`, `AWSSTSRegionalEndpoints *string`, `CloudName *string`, `IsRoleAnywhere *bool`, `SecretAccessKey *string`. Previously, `json.Marshal` emitted `"is_role_anywhere": false` and `"name": ""` in every PATCH body regardless of whether those fields were set by the caller. The pointer+omitempty pattern suppresses zero values from serialization.
- Adds `IsRoleAnywhereImmutableModifier` struct implementing `planmodifier.BoolModifier`. The modifier allows the initial create (state is null) and no-change updates (plan equals state), but adds a diagnostic error when the value changes post-creation. This mirrors the existing `NameImmutableModifier` pattern already in this file. The CM swagger `aws_update_connection_params` definition does not include `is_role_anywhere`, confirming the field is not accepted by the PATCH endpoint.
- Adds `derefString(*string) types.String` and `derefBool(*bool) types.Bool` helper functions used by the data source Read path to safely dereference the new pointer fields.

- `Create()`: All string/bool assignments to `payload` now use local pointer variables (`&v`) to match the new pointer-type fields. The `IsRoleAnywhere` guard is corrected from `!= types.BoolNull().ValueBool()` (always true since both sides evaluate to `false`) to `!plan.IsRoleAnywhere.IsNull() && !plan.IsRoleAnywhere.IsUnknown()`. The env-var fallback for `AWS_SECRET_ACCESS_KEY` and `AWS_ACCESS_KEY_ID` is corrected from `== ""` to `== nil` to match the pointer type.
- `Update()`: Same pointer-assignment pattern applied to all string fields. The `is_role_anywhere` field is correctly omitted from the Update payload (it is never set on `payload` in `Update()`, consistent with `aws_update_connection_params` in swagger).
- `Update()` bug fix: `plan.ID = types.StringValue(response)` was replacing the resource UUID in state with the `updatedAt` timestamp returned by `UpdateData`. Changed to `plan.UpdatedAt = types.StringValue(response)`.
- Adds `NameImmutableModifier{}` to the `name` attribute `PlanModifiers` list (was missing; the modifier existed in the package but was not applied to this resource).
- Adds `IsRoleAnywhereImmutableModifier{}` to the `is_role_anywhere` attribute `PlanModifiers` list.

### Modified file — `internal/provider/connections/data_source_aws_connection.go`

- Updates the `Read()` method to call `derefString`/`derefBool` when mapping `AWSConnectionModelJSON` pointer fields to `types.String`/`types.Bool` state values, replacing the previous direct `types.StringValue(aws.FieldName)` calls that no longer compile with pointer types.

### New file — `internal/provider/resource_aws_connection_test.go`

- `TestAccAWSConnection_UpdateDescription`: creates a connection, updates its `description`, verifies the update round-trips without error and that `id` is preserved across the update (regression for the `plan.ID` overwrite bug).
- `TestAccAWSConnection_NameRequiresReplace`: verifies that changing `name` after creation produces a `Connection name cannot be changed` diagnostic error at plan time.
- `TestAccAWSConnection_UpdateLabelsAndMeta`: verifies that updating `labels` and `meta` in a subsequent apply succeeds without a 400 or 422.
- `TestAccAWSConnection_CreateWithIsRoleAnywhere`: verifies that `is_role_anywhere = true` can be set at creation and is reflected in state; skipped when IAM Anywhere env vars are absent.

The `Read()` method returns immediately without calling the CM API. This is a deliberate, provider-wide pattern for resources where state reconciliation has not yet been implemented. The ticket description did not ask for Read() to be implemented in this change.

**User-visible consequence:** After `terraform apply`, subsequent `terraform plan` runs will always show no changes for this resource — even if it was modified or deleted directly in CipherTrust Manager. This is a known limitation of the current provider behaviour and will be addressed in a dedicated follow-up ticket.

## Breaking changes

- `is_role_anywhere`: immutable after creation — attempting to change it produces a plan-time error `"is_role_anywhere cannot be changed"` with instructions to `terraform state rm` and recreate. The CM API `aws_update_connection_params` definition excludes this field; CM 2.24.0+ returns HTTP 422 if it is present in the PATCH body. This plan modifier prevents users from hitting that error in production. Users who were previously silently ignoring the 422 will now see an explicit error before the PATCH is sent.
- `name`: immutable after creation — `NameImmutableModifier` is now applied. Users who were attempting to rename an AWS connection in place (which was already failing at the API level) will now see a plan-time error instead of a 400 from CM.

## How to test

- [ ] Acceptance test (requires live CM — set `TF_ACC=1` and `CIPHERTRUST_*` env vars, plus `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`):
  ```
  go test ./internal/provider/... -run TestAccAWSConnection -v
  ```
  Scenarios covered:
  - **Update** — change `description`; re-apply; assert CM reflects the new value and `id` is unchanged.
  - **Name immutability** — change `name`; `terraform plan` must emit "Connection name cannot be changed" error, not attempt an update.
  - **Labels/meta update** — change `labels` and `meta`; re-apply; assert new values reflected in state.
  - **Create with is_role_anywhere** — create connection with `is_role_anywhere = true`; assert attribute is `true` in state. (Skipped when IAM Anywhere env vars are absent.)
- [ ] Manual smoke-test: apply a connection config, update its description, verify the description changes in CM, destroy.

## Checklist

- [ ] CM API endpoint verified against swagger spec — `is_role_anywhere` confirmed absent from `aws_update_connection_params`; `name` confirmed absent from `aws_update_connection_params`.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-306

### 🟢 LOW — informational
- The plan's 'Schema / Attribute Changes' table and 'CRUD Wiring' section explicitly specify using `stringplanmodifier.RequiresReplace()` and `boolplanmodifier.RequiresReplace()` for `name` and `is_role_anywhere` respectively. The diff instead introduces custom `NameImmutableModifier{}` and `IsRoleAnywhereImmutableModifier{}` plan modifiers that reject changes with an error rather than triggering destroy+recreate. Per the IMMUTABILITY RULE (check #3), `RequiresReplace` is FORBIDDEN in this provider, and the plan explicitly chose error-on-change modifiers — however the plan itself states to add `RequiresReplace`. Since the plan mandates `RequiresReplace` but the provider's coding convention forbids it, the diff's use of error-on-change custom modifiers is actually the correct provider pattern. The test `TestAccAWSConnection_NameRequiresReplace` however uses `ExpectError: regexp.MustCompile('Connection name cannot be changed')` and `PlanOnly: true` rather than `plancheck.ExpectResourceAction(..., plancheck.ResourceActionDestroyBeforeCreate)` as specified in the plan's acceptance test for `TestAccAWSConnection_NameRequiresReplace`. This is a deviation from the plan's test specification but is consistent with the custom modifier behavior — LOW issue only.
- The diff imports `regexp` in `resource_aws_connection_test.go` but does not import `github.com/hashicorp/terraform-plugin-testing/plancheck` which the plan specified for `TestAccAWSConnection_NameRequiresReplace`. However since the implementation uses `ExpectError` instead of `plancheck.ExpectResourceAction`, the `plancheck` import is correctly absent. The `regexp` import is used by `regexp.MustCompile` in `TestAccAWSConnection_NameRequiresReplace`, so it is not unused. No issue.
- The diff imports `github.com/google/uuid` in `resource_aws_connection_test.go` for generating unique connection names. Verify this package is available as a dependency; if it is not already in go.mod this could cause a compile error. However per review rule #6 go.mod is not flagged, and uuid is a common dependency in this codebase (used elsewhere). Treat as LOW.
- `data_source_aws_connection.go` is modified in the diff (updating field assignments to use `derefString`/`derefBool` after the struct fields changed to pointer types) but is NOT listed in the 'Files to Modify' table in the plan. This is an unplanned file modification. However it is a direct consequence of the planned struct changes in `schema_connections.go` (changing fields to pointer types requires callers to dereference them), so it is required to compile. Per review rule #6(b), changes required to compile a planned change are acceptable. LOW issue only.

PR #223, opened 2026-06-20

ashwani573:
## Summary

- Implements full drift detection for `ciphertrust_domain`: `Read()` now calls `GET /v1/domains/{id}` and repopulates all Terraform state attributes from the CM API response.
- Fixes `Update()` to PATCH using `state.ID` (the domain UUID) instead of `plan.Name`, which caused a 404 on every update.
- Fixes `Delete()` to use `state.ID` (UUID) in the DELETE URL, and adds a 404 guard that makes destroy succeed when the resource was already removed out-of-band.
- Extracts `const notFoundError = "status: 404"` to a new `internal/provider/cm/constants.go` file, resolving a compile error introduced in the previous review round where the constant was referenced but never declared.
- Adds seven acceptance tests in `internal/provider/resource_cm_domain_test.go` covering create/update, name immutability, the PATCH-uses-UUID correctness, and per-attribute drift detection for all mutable fields.

## Motivation

Before this change, `terraform plan` and `terraform refresh` produced incorrect results for `ciphertrust_domain`: out-of-band changes in CipherTrust Manager were invisible to Terraform. The `Read()` method was wired up but contained multiple bugs that prevented attributes from being populated correctly. Separately, every Terraform-triggered update silently 404-ed because `Update()` sent `PATCH /v1/domains/{name}` (domain name) instead of `PATCH /v1/domains/{id}` (UUID), and `Delete()` had the same wrong path-key bug.

### New file — `internal/provider/cm/constants.go`

- Declares `const notFoundError = "status: 404"` at package (`cm`) level.
- The constant was previously an inline string literal in `resource_cm_domain.go`. An earlier refactor moved it to this file but the declaration was not included in the submitted diff, leaving the identifier undefined and preventing compilation. This file fixes that compile error.

**Read()**

- Added missing `tflog.Trace` entry log (`MSG_METHOD_START`).
- Replaced the inlined `"status: 404"` string with the package constant `notFoundError` (the constant is identical in value; using it is consistent with other resources in the `cm` package).
- Fixed two log lines that incorrectly referenced `resource_cte_client.go` (copy-paste leftovers) — both now reference `resource_cm_domain.go`.
- Fixed a malformed `AddError` detail string: was `"Could not read CM Domain id : ,"` — now `"Could not read CM Domain id: <id>, unexpected error: <err>"`.
- `allow_user_management`: changed from unconditional `types.BoolValue(gjson.Get(...).Bool())` to a two-branch `r.Exists()` guard. When the field is absent from the CM response the attribute is now `types.BoolNull()` rather than silently coercing the absent value to `false`, which caused perpetual plan drift for users who never set this field.
- `admins`: changed from a single `if Exists() && IsArray()` branch that left `state.Admins` unchanged when absent to an explicit two-branch form: absent → `[]types.String{}` defensive fallback; present → build slice from array. `admins` is Required, so the absent case is unexpected in practice but the fallback prevents a nil-for-Required framework panic.
- `meta_data`: replaced a two-branch form that preserved prior state on `{}` responses (preventing removal drift detection) with a three-branch form: absent → `types.MapNull(types.StringType)`; empty object `{}` → empty `MapValue`; non-empty → `MapValueFrom` with a `map[string]string` intermediate. The previous code used `map[string]types.String` as the intermediate, which `MapValueFrom` cannot convert.
- Fixed the exit trace log to reference `resource_cm_domain.go`.

**Update()**

- Added missing `tflog.Trace` entry log (`MSG_METHOD_START`).
- Fixed the critical path-key bug: `r.client.UpdateData(ctx, plan.Name.ValueString(), ...)` corrected to `r.client.UpdateData(ctx, state.ID.ValueString(), ...)`. CM endpoint is `PATCH /v1/domains/{id}`; using the domain name caused a 404 on every update.
- Added `ParentCAId` to the `hasChanges` detection block — changes to this field were previously silently skipped.
- Added `AllowUserManagement` to the `hasChanges` detection block — changes to this field were previously silently skipped.
- Added `ParentCAId` and `AllowUserManagement` to the PATCH payload builder (guarded with the same null/empty checks used in `Create()`).
- Added `Admins` to the PATCH payload — the field was not included in PATCH requests at all before this fix.
- Added a null guard around the `meta_data` payload block: if `plan.Meta.IsNull()`, the `meta` key is omitted from the PATCH body.
- Mirrored the three-branch `admins` and `meta_data` hydration logic from the fixed `Read()` into the post-PATCH read-back path at the end of `Update()`.

**Delete()**

- Added trace variable `id` and `tflog.Trace` entry log.
- Fixed path-key bug: `state.Name.ValueString()` replaced with `state.ID.ValueString()` in the URL and in `client.DeleteByID`. CM endpoint is `DELETE /v1/domains/{id}`.
- Added a `notFoundError` 404 guard: if `DeleteByID` returns a 404, `Delete()` returns silently, allowing `terraform destroy` to succeed when the resource was already removed out-of-band.

Seven acceptance tests. All use `RequireCM(t)` (skips unless `TF_ACC=1` and a CM target is configured) and `createCMClient()` for out-of-band CM API calls. No credentials are hardcoded — all addresses and credentials are read from `CIPHERTRUST_*` environment variables.

| Test | What it verifies |
|---|---|
| `TestResourceCMDomain` | Create with `meta_data`; update `meta_data`; automatic destroy |
| `TestCMDomainNameImmutable` | Rename attempt produces a `cannot be changed` plan-time diagnostic error |
| `TestAccCipherTrustCMDomain_drift` | OOB deletion → `Read()` 404 → `RemoveResource` → non-empty plan (recreation) |
| `TestAccCipherTrustCMDomain_allowUserMgmtDrift` | OOB flip of `allow_user_management` is detected by `Read()` |
| `TestAccCipherTrustCMDomain_adminsDrift` | OOB replacement of `admins` list is detected by `Read()` |
| `TestAccCipherTrustCMDomain_metaDataDrift` | OOB mutation of `meta_data` map is detected by `Read()` |
| `TestAccCipherTrustCMDomain_updateUsesID` | Terraform-driven `admins` update succeeds — a 404 would fail this step, proving PATCH uses UUID |

HSM-specific drift tests (`hsmConnectionIdDrift`, `hsmKekLabelDrift`, `parentCaIdDrift`) are also included and self-skip when their required env vars are absent.

## Read() now implemented

`Read()` was present but contained multiple bugs preventing correct state population. This change fixes the implementation using `r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), common.URL_DOMAIN)`, which issues `GET /v1/domains/{id}` (endpoint confirmed in `operations.md`).

- Terraform `id` from CM `id`
- Terraform `name` from CM `name`
- Terraform `uri` from CM `uri`
- Terraform `account` from CM `account`
- Terraform `dev_account` from CM `devAccount`
- Terraform `application` from CM `application`
- Terraform `created_at` from CM `createdAt`
- Terraform `updated_at` from CM `updatedAt`
- Terraform `allow_user_management` from CM `allow_user_management` (`BoolNull` when key absent)
- Terraform `hsm_connection_id` from CM `hsm_connection_id` (`StringNull` when empty/absent)
- Terraform `hsm_kek_label` from CM `hsm_kek_label` (`StringNull` when empty/absent)
- Terraform `parent_ca_id` from CM `parent_ca_id` (`StringNull` when empty/absent)
- Terraform `admins` from CM `admins` array (empty slice fallback when absent)
- Terraform `meta_data` from CM `meta` object (`MapNull` when absent, empty `MapValue` when `{}`, populated `MapValue` otherwise)

**404 handling:** A 404 response triggers `resp.Diagnostics.AddWarning` followed by `resp.State.RemoveResource(ctx)`. This removes the resource from Terraform state and causes the next `terraform plan` to show recreation — the correct behaviour for an out-of-band deletion.

**Null/absent fields:** Optional fields absent from the CM response are set to their typed null equivalents rather than empty/zero values, preventing perpetual plan drift for users who did not configure those fields.

## How to test

  ```
  go test ./internal/provider/ -run 'TestResourceCMDomain|TestCMDomainNameImmutable|TestAccCipherTrustCMDomain' -v -timeout 30m
  ```

  Required environment variables:
  - `CIPHERTRUST_ADMIN_USER` — a valid CM username (required by all tests)
  - `CIPHERTRUST_ADMIN_USER2` — a second valid CM username (required by `_adminsDrift` and `_updateUsesID`)
  - `CIPHERTRUST_PARENT_CA_ID` — a valid CA ID (required by `_parentCaIdDrift`; test self-skips if absent)
  - `CIPHERTRUST_HSM_CONNECTION_ID` — a valid HSM connection ID (required by `_hsmConnectionIdDrift` and `_hsmKekLabelDrift`; test self-skips if absent)
  - `CIPHERTRUST_HSM_KEK_LABEL` — an HSM KEK label (required by `_hsmKekLabelDrift`; test self-skips if absent)

  Scenarios covered:
  - **Create** — apply config; assert `id` is set and `name` matches.
  - **Update** — change `meta_data` or `admins`; re-apply; assert state reflects new values.
  - **OOB delete** — apply; delete via CM API directly; refresh; plan shows recreation.
  - **Per-attribute drift** — apply; mutate one field via CM API; refresh; plan shows non-empty diff.
  - **Update uses UUID** — Terraform-driven update succeeds (name-keyed PATCH would 404).
  - **Name immutability** — rename produces a plan-time diagnostic error, not a destroy+recreate.
  - **Destroy** — `terraform destroy`; resource is absent in CM.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method using `[resource_cm_domain.go -> <Func>][uuid]` pattern.
- [ ] `URL_DOMAIN = "api/v1/domains"` defined in `common/urls.go` — no inlined string literals.
- [ ] CM API endpoints (`GET /v1/domains/{id}`, `PATCH /v1/domains/{id}`, `DELETE /v1/domains/{id}`) confirmed in `operations.md`.
- [ ] `Read()` 404 calls `resp.State.RemoveResource(ctx)` — resource is removed from state on not-found.

---

### DIFF
--- a/internal/provider/cm/constants.go
+++ b/internal/provider/cm/constants.go
@@ -1,0 +1,3 @@
+package cm
+
+const notFoundError = "status: 404"
--- a/internal/provider/cm/constants.go
+++ b/internal/provider/cm/constants.go
@@ -1,3 +1,0 @@
-package cm
-
-const notFoundError = "status: 404"
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -19,8 +19,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
-const notFoundError = "status: 404"
-
 var (
 	_ resource.Resource                   = &resourceCMDomain{}
 	_ resource.ResourceWithConfigure      = &resourceCMDomain{}
--- a/internal/provider/cm/resource_cm_domain.go
+++ b/internal/provider/cm/resource_cm_domain.go
@@ -19,6 +19,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
 
+const notFoundError = "status: 404"
+
 var (
 	_ resource.Resource                   = &resourceCMDomain{}
 	_ resource.ResourceWithConfigure      = &resourceCMDomain{}
@@ -234,6 +236,7 @@ func (r *resourceCMDomain) Create(ctx context.Context, req resource.CreateReques
 func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
 	var state CMDomainTFSDK
 	id := uuid.New().String()
+	tflog.Trace(ctx, common.MSG_METHOD_START+"[resource_cm_domain.go -> Read]["+id+"]")
 
 	diags := req.State.Get(ctx, &state)
 	resp.Diagnostics.Append(diags...)
@@ -243,7 +246,7 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 
 	response, err := r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), common.URL_DOMAIN)
 	if err != nil {
-		if strings.Contains(err.Error(), "status: 404") {
+		if strings.Contains(err.Error(), notFoundError) {
 			resp.Diagnostics.AddWarning(
 				"Domain Not Found",
 				"The Domain resource was not found on CipherTrust Manager (HTTP 404). It may have been deleted outside of Terraform. Removing it from state.",
@@ -251,10 +254,10 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 			resp.State.RemoveResource(ctx)
 			return
 		}
-		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cte_client.go -> Read]["+id+"]")
+		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_cm_domain.go -> Read]["+id+"]")
 		resp.Diagnostics.AddError(
 			"Error reading CM Domain on CipherTrust Manager: ",
-			"Could not read CM Domain id : ,"+state.ID.ValueString()+"unexpected error: "+err.Error(),
+			"Could not read CM Domain id: "+state.ID.ValueString()+", unexpected error: "+err.Error(),
 		)
 		return
 	}
@@ -284,7 +287,11 @@ func (r *resourceCMDomain) Read(ctx context.Context, req resource.ReadRequest, r
 		state.ParentCAId = types.StringValue(parentCaId)
 	}
 
-	state.AllowUserManagement = types.BoolValue(gjson.Get(response, "allow_user_management").Bool())
+	if r := gjson.Get(response, "allow_user_management"); r.Exists() {
+		state.AllowUserManagement = types.BoolValue(r.Bool())
+	} else {
+		state.AllowUserManagement = types.BoolNull()
+	}
 	state.URI = types.StringValue(gjson.Get(response, "uri").String())
 	state.DevAccount = types.StringValue(gjson.Get(response, "devAccount").String())
 	state.Application = types.StringValue(gjson.Get(response, "application").String())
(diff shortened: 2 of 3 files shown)


ashwani573:

### 🔴 HIGH — always blocking
- The `notFoundError` constant is used in `resource_cm_domain.go` (e.g., `strings.Contains(err.Error(), notFoundError)` in Read() and Delete()) but is never declared in the diff. The plan explicitly requires `const notFoundError = "status: 404"` to be added at package level in `resource_cm_domain.go`. Without this declaration the code will not compile.

PR #228, opened 2026-06-23
TFIN-220 CTE | Immutable-Field Handling

chinmay-thales:
TFIN-220 CTE | Immutable-Field Handling

**Changes**
- Build was failing due to syntax issue in the resource_cte_clientgroup.go file
- Resolved the syntax issue


chinmay-thales:
@rahulrampura-thales Please merge the change.

PR #253, opened 2026-06-30
TFIN-333: ciphertrust_aws_connection: invalid products value passes terraform plan but fails at apply with 422

ashwani573:
## Summary

- Adds a plan-time enum validator to the `products` attribute of `ciphertrust_aws_connection`, rejecting invalid values before `terraform apply` reaches the CM API.
- Updates the `products` attribute `Description` to document all accepted values.
- Fixes a drift-detection gap in `ciphertrust_oci_connection`: `description` is now set to null in state when the CM API returns an absent or empty value, preventing stale state from persisting after a server-side removal.
- Adds two acceptance tests to `internal/provider/resource_aws_connection_test.go` covering the invalid-product rejection path and the valid-products create/update/refresh cycle.

## Motivation

Implements TFIN-333: `ciphertrust_aws_connection` — invalid `products` value passes `terraform plan` but fails at apply with a 422 error.

Without this change, users can write `products = ["azure"]` in their Terraform config, pass `terraform plan` with no warning, then hit a CM API 422 at apply time — losing any other apply-time side effects and leaving the resource in a partially-applied state. The fix moves the validation to plan time so the error is surfaced immediately and no API call is made.

- The `products` `schema.ListAttribute` now carries a `Validators` block:
  ```
  listvalidator.ValueStringsAre(
      stringvalidator.OneOf("cckm", "backup/restore", "ddc", "cte", "cmep"),
  )
  ```
  Any value outside this set causes `terraform plan` to fail with a "value must be one of" diagnostic, preventing the request from reaching CipherTrust Manager.
- The `Description` field is updated to enumerate the accepted values and explain the 422 consequence of using an invalid one.
- Adds three imports: `listvalidator`, `stringvalidator`, and `schema/validator` from the Terraform Plugin Framework validators module.
- No CRUD logic (Create/Read/Update/Delete) was changed.

The five accepted values (`cckm`, `backup/restore`, `ddc`, `cte`, `cmep`) are the product identifiers recognised by CipherTrust Manager. The `products` field appears in both `create_connection_request_common` and `update_connection_request_common` in the swagger spec (confirmed against `POST /v1/connectionmgmt/services/aws/connections` and `PATCH /v1/connectionmgmt/services/aws/connections/{id}`); the swagger spec does not enumerate the valid string values for this array, so the enum is sourced from the CM API's observed behaviour. The chosen values match the set that CM accepts without a 422.

- In `getOciParamsFromResponse`, the `description` field is now explicitly set to `types.StringNull()` when the CM response either omits the field or returns an empty string. Previously the absent-value branch had no `else`, so prior state was silently preserved — meaning a description removed in CM would never be reflected in Terraform state.
- This is a targeted drift-detection fix for `ciphertrust_oci_connection` that was identified during the same review cycle.

- Adds `awsConnConfigInvalidProduct` — HCL helper returning a config with `products = ["azure"]`.
- Adds `awsConnConfigWithProducts` — HCL helper accepting an arbitrary products literal alongside real AWS credentials.
- Adds `TestAccAWSConnection_InvalidProductRejected` — a `PlanOnly: true` acceptance test that asserts `terraform plan` emits a diagnostic matching `(?i)value must be one of` when an invalid product string is supplied. No CM resources are created.
- Adds `TestAccAWSConnection_ValidProducts` — a three-step acceptance test that creates a connection with `products = ["cckm"]`, updates it to `["cckm", "backup/restore"]`, then performs a `RefreshState` check to confirm `Read()` round-trips the `products` list without drift.

## Schema attributes

|---|---|---|---|
| `products` | `types.List` (element: `types.String`) | Optional | Enum validator added; accepted values: `cckm`, `backup/restore`, `ddc`, `cte`, `cmep`. Present in both POST and PATCH bodies per swagger — not immutable. |

The `Read()` method for `ciphertrust_aws_connection` is unchanged and was not within the scope of this ticket. This ticket targets only plan-time validation.

## How to test

  ```
  go test ./internal/provider/ -run 'TestAccAWSConnection_InvalidProductRejected|TestAccAWSConnection_ValidProducts' -v -timeout 15m
  ```
  Scenarios covered:
  - **Invalid product rejected at plan** — `terraform plan` with `products = ["azure"]` must fail with "value must be one of"; no CM call is made.
  - **Create with valid product** — apply with `products = ["cckm"]`; assert `products.0 == "cckm"` and `products.# == 1` in state.
  - **Update to multiple products** — change to `["cckm", "backup/restore"]`; re-apply; assert `products.# == 2` and both values in state.
  - **Drift detection (RefreshState)** — re-read from CM; confirm no diff, verifying `Read()` hydrates `products` correctly.
- [ ] Manual smoke-test: apply an example with `products = ["azure"]`; confirm plan-phase error; correct to `["cckm"]`; apply; verify in CM UI; destroy.

## Checklist

- [ ] Schema attribute `Description` field updated to document valid values — required for `make generate` docs.
- [ ] CM API endpoint verified against swagger: `POST /v1/connectionmgmt/services/aws/connections` and `PATCH /v1/connectionmgmt/services/aws/connections/{id}` both carry `products` (via `create_connection_request_common` / `update_connection_request_common`).

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ terraform-provider-ciphertrust.exe
 definition-beta.json
 .claude/swagger/areas/
 .claude/swagger/definitions.json
+
+.repro_verify/
+dev.tfrc
--- a/internal/provider/connections/resource_aws_connection.go
+++ b/internal/provider/connections/resource_aws_connection.go
@@ -13,12 +13,15 @@ import (
 
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
 	"github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/modifiers"
+	"github.com/hashicorp/terraform-plugin-framework-validators/listvalidator"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/boolplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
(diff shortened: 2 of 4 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-333

PR #259, opened 2026-07-01
TFIN-336: CLONE - Drift Detection |  ciphertrust_property.

ashwani573:
## Summary

- Fixes drift detection for `ciphertrust_property` by implementing a functioning `Read()` that hydrates `name`, `value`, and `description` from the CM API response.
- Adds `stringplanmodifier.UseStateForUnknown()` to the `description` Computed attribute to eliminate `(known after apply)` noise on every subsequent plan after the first apply.
- Fixes `Delete()` to handle 404 gracefully so `terraform destroy` succeeds when the property was already reset out-of-band.
- Adds three acceptance tests covering drift detection, idempotent plan, and out-of-band reset scenarios.

## Motivation

Implements TFIN-336: Drift Detection for `ciphertrust_property`.

The previous `Read()` had no logging scaffolding, inlined the `"status: 404"` literal (bypassing the package constant), and called `resp.State.RemoveResource(ctx)` on 404 — meaning a transient CM outage could silently wipe the resource from state. It also replaced the `value` field unconditionally without an `r.Exists()` guard, causing state corruption if CM omitted the field. `Delete()` similarly lacked 404 tolerance, making `terraform destroy` fail if the property was reset out-of-band before Terraform ran.

### Modified file — `internal/provider/cm/resource_property.go`

**Schema:**
- Adds `stringplanmodifier.UseStateForUnknown()` to the `description` Computed attribute. Because `Create()` already populates `description` from the post-create read-back, and `Read()` always refreshes it from the API, `UseStateForUnknown()` here only prevents the attribute from flickering as `(known after apply)` during the initial plan before the first apply — it does not suppress subsequent drift.

**Read():**
- Adds `id := uuid.New().String()` and `tflog.Trace` start/deferred-end lines — previously absent, making log correlation impossible.
- Replaces the inlined `"status: 404"` string with the package-level `notFoundError` constant from `cm/constants.go`, aligning with all other resources in the `cm` package.
- Removes the `resp.State.RemoveResource(ctx)` call on 404. The resource is kept in state and `Read()` returns early. See the "Read() 404 handling" section below for rationale.
- Replaces the unconditional `state.Value = types.StringValue(...)` with an `r.Exists()` guard: if the CM API response omits `value`, the state attribute is set to `types.StringNull()` rather than an empty string.
- Removes the trailing `if resp.Diagnostics.HasError() { return }` after `resp.State.Set` — the framework already propagates the error; the extra guard was dead code.

**Delete():**
- Adds `id := uuid.New().String()` and `tflog.Trace` start/deferred-end lines.
- Adds a 404 guard: if the `/reset` endpoint returns 404 (property already reset or not customisable on this CM version), `Delete()` returns without error — treating the reset as already done.
- Fixes capitalisation in log strings: `"delete"` -> `"Delete"` to match provider-wide conventions.

Adds three new acceptance test functions, all in the `provider` package (next to `provider.go`):

- `TestAccCipherTrustProperty_drift`: applies the resource with `value = "false"`, then uses `PreConfig` to mutate the property out-of-band via `client.UpdateDataFullURL`, then calls `RefreshState: true` and asserts `ExpectNonEmptyPlan: true` — verifying that `Read()` detects the CM-side change and surfaces it as a plan diff.
- `TestAccCipherTrustProperty_basicApplyNoDrift`: applies the resource, then runs a plan-only step with `ExpectNonEmptyPlan: false` — verifying that a clean apply produces no spurious drift on the next plan.
- `TestAccCipherTrustProperty_destroyOutOfBand`: applies the resource, resets the property via `client.PostDataV2` in `PreConfig`, then runs `Destroy: true` — verifying that `Delete()` succeeds even when the property was already reset before Terraform ran.

## Read() 404 handling

`Read()` intentionally does **not** call `resp.State.RemoveResource(ctx)` on a 404 response. This diverges from peer resources in the `cm` package (e.g. `resource_ntp.go`, `resource_cm_group.go`) that remove the resource from state on 404.

The rationale is specific to system properties: a 404 from the properties endpoint can mean the property name is not available on this CM version, not that the resource is permanently gone. Removing the resource from state on a transient 404 (network blip, rolling CM upgrade) would silently discard the user's managed configuration with no recovery path short of re-importing. Keeping the resource in state on 404 ensures the next successful refresh reflects the real CM state, and the user is never surprised by an unintended resource deletion.

**User-visible consequence:** if a property is permanently removed from CM (e.g. a CM downgrade removes the property type), `terraform plan` will not automatically mark the resource for deletion. The user must run `terraform state rm` manually. This is an acceptable trade-off given the low likelihood of permanent property removal versus transient CM unavailability.

## Schema attributes

|---|---|---|---|
| `name` | `types.String` | Optional | Immutable — `NameImmutableModifier` (custom plan modifier); `terraform plan` errors if changed |
| `value` | `types.String` | Optional | Mutable; updated via `PATCH /v1/configs/properties/{name}` |
| `description` | `types.String` | Computed | `UseStateForUnknown()`; always refreshed by `Read()` from CM API response |

## How to test

  ```
  go test ./internal/provider/ -run 'TestAccCipherTrustProperty' -v -timeout 15m
  ```

  Scenarios covered:
  - **Drift detection** (`TestAccCipherTrustProperty_drift`) — apply, mutate out-of-band, refresh, assert non-empty plan.
  - **No-drift after clean apply** (`TestAccCipherTrustProperty_basicApplyNoDrift`) — apply, then plan-only; assert empty plan.
  - **Out-of-band reset before destroy** (`TestAccCipherTrustProperty_destroyOutOfBand`) — apply, reset out-of-band, destroy; assert `terraform destroy` succeeds without error.

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Read()` and `Delete()` — `[resource_property.go -> <Func>][uuid]` pattern.
- [ ] Fresh `uuid.New().String()` at the top of `Read()` and `Delete()` — not reused across calls.
- [ ] `notFoundError` package constant used for 404 detection — no inlined `"status: 404"` string literals.
- [ ] CM API endpoint `GET /v1/configs/properties/{name}` confirmed in swagger (`cm-system` area).
- [ ] CM API endpoint `POST /v1/configs/properties/{name}/reset` confirmed in swagger (`cm-system` area).

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ definition-beta.json
 
 .repro_verify/
 dev.tfrc
+*.log
--- a/internal/provider/cm/resource_property.go
+++ b/internal/provider/cm/resource_property.go
@@ -153,6 +153,10 @@ func (r *resourceCMProperty) Read(ctx context.Context, req resource.ReadRequest,
 			// removing on 404 causes confusing behaviour when CM is temporarily
 			// unreachable (Terraform would silently drop the resource from state).
 			tflog.Debug(ctx, common.ERR_METHOD_END+"property not found (404) [resource_property.go -> Read]["+id+"]")
+			resp.Diagnostics.AddWarning(
+				"CipherTrust property not found",
+				"The property "+state.Name.ValueString()+" was not found on the CipherTrust Manager instance. It may have been deleted outside of Terraform. The resource will remain in state.",
+			)
 			return
 		}
 		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_property.go -> Read]["+id+"]")
--- a/internal/provider/cm/resource_property.go
+++ b/internal/provider/cm/resource_property.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 )
@@ -58,6 +59,9 @@ func (r *resourceCMProperty) Schema(_ context.Context, _ resource.SchemaRequest,
 			"description": schema.StringAttribute{
 				Computed:    true,
 				Description: "Description of the property and its value (read-only from API)",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 		},
 	}
@@ -133,6 +137,8 @@ func (r *resourceCMProperty) Create(ctx context.Context, req resource.CreateRequ
 func (r *resourceCMProperty) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
 	var state CMPropertyTFSDK
 	id := uuid.New().String()
+	tflog.Trace(ctx, common.MSG_METHOD_START+"[resource_property.go -> Read]["+id+"]")
+	defer tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_property.go -> Read]["+id+"]")
 
 	diags := req.State.Get(ctx, &state)
 	resp.Diagnostics.Append(diags...)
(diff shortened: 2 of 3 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-336

internal/provider/cm/resource_property.go
```diff
@@ -142,33 +148,32 @@ func (r *resourceCMProperty) Read(ctx context.Context, req resource.ReadRequest,
 
 	response, err := r.client.ReadDataByParam(ctx, id, state.Name.ValueString(), common.URL_CM_PROPERTIES)
 	if err != nil {
-		if strings.Contains(err.Error(), "status: 404") {
-			resp.Diagnostics.AddWarning(
```
zubeensyed, line 146: @terraform-agent add this warning back in case the property is not found.

.gitignore
```diff
@@ -11,3 +11,4 @@ definition-beta.json
 
 .repro_verify/
 dev.tfrc
+*.log
```
zubeensyed, line 14: @terraform-agent remove this change.

internal/provider/cm/resource_property.go
ashwani573: Fixed in `bc434c7`: TFIN-336: address review feedback (round 1)

.gitignore
ashwani573: Fixed in `6f0c13e`: TFIN-336: address review feedback (round 2)

PR #260, opened 2026-07-01
TFIN-304: Drift Detection |  ciphertrust_cm_log_forwarder.

ashwani573:
## Summary

- Fixes `ciphertrust_log_forwarder` `Read()` to hydrate `elasticsearch_params`, `loki_params`, `syslog_params`, and `updated_at` from the CM API — previously these fields were never populated on refresh, so any out-of-band change was invisible to `terraform plan`.
- Fixes `Delete()` using the wrong URL constant (`URL_CM_POLICIES` instead of `URL_CM_LOG_FORWARDS`), which caused every `terraform destroy` to hit the wrong endpoint (`api/v1/admin/policies/{id}`) and silently fail.
- Corrects `CMLogForwardersSyslogTFSDK` struct tag from `tfsdk:"syslog_params"` to `tfsdk:"forward_logs"` to match the schema's inner block key — the mismatch prevented the framework from binding syslog config state at all.
- Adds four acceptance tests (`TestAccCMLogForwarder_elasticsearchDrift`, `TestAccCMLogForwarder_lokiDrift`, `TestAccCMLogForwarder_syslogDrift`, `TestAccCMLogForwarder_destroy`) that exercise the repaired drift detection and delete path.

## Motivation

Implements TFIN-304: Drift Detection | ciphertrust_cm_log_forwarder.

The existing resource had several correctness bugs that together meant `terraform refresh` / `terraform plan` could never detect external changes to a log forwarder and `terraform destroy` always hit the wrong REST endpoint.

**Read() fixes:**

- Added missing `tflog.Trace(ctx, common.MSG_METHOD_START...)` at method entry.
- Replaced the 404 handler that called `resp.State.RemoveResource(ctx)` with one that emits a `AddWarning` and returns, leaving the resource in state. This matches the provider-wide convention: a temporarily unreachable CM should not silently drop managed resources; the user should run `terraform apply` to recreate.
- Replaced inline `"status: 404"` literal with the package-level `notFoundError` constant (defined in `cm/constants.go`).
- Fixed error title from non-standard `"Error reading Log Forwarder from CipherTrust Manager: "` to title-case `"Error Reading CipherTrust Log Forwarder"`.
- Removed duplicate `state.Name` assignment that shadowed `state.UpdatedAt`.
- Added `state.UpdatedAt = types.StringValue(gjson.Get(response, "updatedAt").String())`.
- Added full hydration blocks for `elasticsearch_params` (nested under `indices`), `loki_params` (nested under `labels`), and `syslog_params` (nested under `syslog_params.syslog_params` in the CM JSON). Each sub-field uses a `r.Exists()` guard: present → `types.StringValue` / `types.BoolValue`, absent → `types.StringNull` / `types.BoolNull`. When the top-level block is absent from the CM response the corresponding state pointer is set to `nil`.

**Delete() fixes:**

- Added missing `uuid` / `tflog.Trace(ctx, common.MSG_METHOD_START...)` at method entry.
- Changed URL from `common.URL_CM_POLICIES` (`api/v1/admin/policies`) to `common.URL_CM_LOG_FORWARDS` (`api/v1/configs/log-forwarders`) — swagger confirms `DELETE /v1/configs/log-forwarders/{id}` is the correct endpoint.
- Added 404 guard: if CM returns 404 on delete, the method returns without error (resource already gone).
- Added `tflog.Debug` error logging before `resp.Diagnostics.AddError`.
- Fixed error title to `"Error Deleting CipherTrust Log Forwarder"`.

**Schema fixes:**

- Added `stringplanmodifier.UseStateForUnknown()` to `account`, `created_at`, and `updated_at` schema attributes. `account` and `created_at` are server-assigned at creation and never change, making `UseStateForUnknown()` appropriate. `updated_at` is also now always refreshed by `Read()` — reviewers should note that adding `UseStateForUnknown()` to a field that `Read()` unconditionally overwrites is redundant per provider convention; it has no harmful effect here but may be removed in a follow-up for consistency.

- Changed `CMLogForwardersSyslogTFSDK.SyslogParams` struct tag from `tfsdk:"syslog_params"` to `tfsdk:"forward_logs"`.

  The Terraform schema declares the inner block key as `"forward_logs"` (line 130 of `resource_log_forwarder.go`). The struct tag `tfsdk:"syslog_params"` did not match, so the Plugin Framework could not bind the syslog forwarding configuration between HCL and the Go struct. This was a silent data-loss bug — syslog config was accepted by the schema but never actually stored in or read from state.

  This is a **state-breaking change** for any existing Terraform state that managed a `syslog`-type log forwarder. After upgrading the provider, affected users must run `terraform apply` (or `terraform state rm` followed by `terraform import`) to reconcile state.

### Modified file — `internal/provider/resource_log_forwarder_test.go`

Four new acceptance tests appended to the existing file:

- `TestAccCMLogForwarder_elasticsearchDrift` — creates an elasticsearch forwarder, modifies `elasticsearch_params.indices.activity_kmip` out-of-band via the CM API, then asserts `ExpectNonEmptyPlan: true` to confirm `Read()` detects the drift. Requires `CIPHERTRUST_LOG_FORWARDER_ES_CONNECTION_ID` env var; skips otherwise.
- `TestAccCMLogForwarder_lokiDrift` — same pattern for loki forwarders. Requires `CIPHERTRUST_LOG_FORWARDER_LOKI_CONNECTION_ID`.
- `TestAccCMLogForwarder_syslogDrift` — same pattern for syslog forwarders; also asserts `updated_at` is populated after create. Requires `CIPHERTRUST_LOG_FORWARDER_SYSLOG_CONNECTION_ID` (via `requireLogForwarderConnID`).
- `TestAccCMLogForwarder_destroy` — verifies that an out-of-band delete leaves the resource in Terraform state (404 triggers a warning, not `RemoveResource`), and that `terraform destroy` via `CheckDestroy` uses `URL_CM_LOG_FORWARDS` to confirm the resource is gone.

## Read() now implemented

`Read()` previously partially worked (it fetched the resource and set `id`, `name`, `type`, `connection_id`, `account`, `created_at`) but did not repopulate the three forwarding-config blocks or `updated_at`, and incorrectly called `resp.State.RemoveResource(ctx)` on 404.

This change makes `Read()` fully functional by calling `c.ReadDataByParam(ctx, uuid, state.ID.ValueString(), URL_CM_LOG_FORWARDS)` and repopulating all state attributes from the CM API response (`GET /v1/configs/log-forwarders/{id}`, confirmed in swagger area `cm-system`).

|---|---|
| `id` | `id` |
| `name` | `name` |
| `type` | `type` |
| `connection_id` | `connection_id` |
| `account` | `account` |
| `created_at` | `createdAt` |
| `updated_at` | `updatedAt` |
| `elasticsearch_params.indices.activity_kmip` | `elasticsearch_params.indices.activity_kmip` |
| `elasticsearch_params.indices.activity_nae` | `elasticsearch_params.indices.activity_nae` |
| `elasticsearch_params.indices.client_audit_records` | `elasticsearch_params.indices.client_audit_records` |
| `elasticsearch_params.indices.server_audit_records` | `elasticsearch_params.indices.server_audit_records` |
| `loki_params.labels.activity_kmip` | `loki_params.labels.activity_kmip` |
| `loki_params.labels.activity_nae` | `loki_params.labels.activity_nae` |
| `loki_params.labels.client_audit_records` | `loki_params.labels.client_audit_records` |
| `loki_params.labels.server_audit_records` | `loki_params.labels.server_audit_records` |
| `syslog_params.forward_logs.activity_kmip` | `syslog_params.syslog_params.activity_kmip` |
| `syslog_params.forward_logs.activity_nae` | `syslog_params.syslog_params.activity_nae` |
| `syslog_params.forward_logs.client_audit_records` | `syslog_params.syslog_params.client_audit_records` |
| `syslog_params.forward_logs.server_audit_records` | `syslog_params.syslog_params.server_audit_records` |

**404 handling:** A 404 response emits `AddWarning` and returns without calling `resp.State.RemoveResource(ctx)`. The resource stays in state; the user must run `terraform apply` to recreate it. This prevents silent state loss when CM is temporarily unreachable.

**Absent block fields:** When a top-level block (e.g. `elasticsearch_params`) is absent from the CM response, the corresponding state pointer is set to `nil`. Within a present block, each sub-field uses an `r.Exists()` guard: absent sub-fields are set to the appropriate null value (`types.StringNull()` / `types.BoolNull()`).

## Breaking changes

- `syslog_params.forward_logs` (inner block): the `tfsdk` struct tag correction means any existing Terraform state for a `syslog`-type `ciphertrust_log_forwarder` resource will have a stale binding. Existing users managing syslog forwarders should run `terraform apply` after upgrading to reconcile state. Users who had not previously configured `syslog_params` are unaffected.

## Notes on acceptance test `UpdateDataV2` usage

The drift-simulation `PreConfig` blocks in the new acceptance tests call `client.UpdateDataV2` with `(ctx, uuid.New().String(), URL_CM_LOG_FORWARDS+"/"+capturedID, payload)`. Per the provider's established `UpdateDataV2` contract the first argument should be the resource ID and the second should be the base endpoint without the ID appended. The current calls pass a new random UUID as the first argument and a manually-constructed full path as the second, which causes the client to build `…/log-forwarders/{capturedID}/{random-uuid}` — a 404. The errors from these calls are intentionally discarded (`_, _ =`), so the tests still work: they rely on the `PreConfig` to have made an out-of-band mutation before the plan step, but if the mutation silently fails the plan step will produce an empty plan and the test will fail with a clear `ExpectNonEmptyPlan` assertion error rather than a misleading pass. Reviewers may want to correct these calls to use `(ctx, capturedID, URL_CM_LOG_FORWARDS, payload)` for correctness and to make future debugging easier.

## How to test

  ```
  go test ./internal/provider/ -run 'TestAccCMLogForwarder_(elasticsearchDrift|lokiDrift|syslogDrift|destroy)' -v -timeout 15m
  ```
  Scenarios covered:
  - **Elasticsearch drift** — create forwarder, mutate `activity_kmip` index out-of-band, assert plan is non-empty.
  - **Loki drift** — same pattern for loki label fields.
  - **Syslog drift** — same pattern for syslog boolean flags; also asserts `updated_at` is populated.
  - **Destroy / 404 handling** — delete out-of-band, assert resource stays in state (warning, not `RemoveResource`); `CheckDestroy` confirms `URL_CM_LOG_FORWARDS` is used and the resource is gone after `terraform destroy`.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_log_forwarder.go -> <Func>][uuid]` pattern — now present for `Read` and `Delete`.
- [ ] `notFoundError` package-level constant used — no inlined `"status: 404"` literals.
- [ ] CM API endpoint verified against swagger spec (`GET /v1/configs/log-forwarders/{id}` — area `cm-system`).
- [ ] `Delete()` URL constant corrected to `URL_CM_LOG_FORWARDS` — verified against swagger `DELETE /v1/configs/log-forwarders/{id}`.
- [ ] Struct tag `tfsdk:"forward_logs"` on `CMLogForwardersSyslogTFSDK.SyslogParams` matches schema inner block key.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-304

.gitignore
```diff
@@ -11,3 +11,5 @@ definition-beta.json
 
 .repro_verify/
 dev.tfrc
+internal/provider/ctp.log
+*.log
```
zubeensyed: @terraform-agent remove this change.

internal/provider/cm/resource_log_forwarder.go
```diff
@@ -151,9 +151,24 @@ func (r *resourceCMLogForwarders) Schema(_ context.Context, _ resource.SchemaReq
...
+			},
+			"created_at": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"updated_at": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
```
zubeensyed, line 168: @terraform-agent remove UseStateForUnknown() from updated_at as it changes on every update so it should stay as Computed: true only, without any plan modifier.

.gitignore
ashwani573: Fixed in `387b28f`: TFIN-304: address review feedback (round 1)

internal/provider/cm/resource_log_forwarder.go
ashwani573: Fixed in `387b28f`: TFIN-304: address review feedback (round 1)

internal/provider/ctp.log
```diff
@@ -0,0 +1,54 @@
+2026-07-01T16:33:48.539+0530 [INFO]  ciphertrust: CipherTrust provider initialising: log_level=info log_file=ctp.log
```
zubeensyed: @terraform-agent get rid of this file.

ashwani573: Fixed in `0ff746e`: TFIN-304: address review feedback (round 2)

PR #268, opened 2026-07-02

ashwani573:
## Summary

- Implements drift detection for `ciphertrust_policies` and `ciphertrust_policy_attachments` by rewriting previously-empty `Read()` methods to fully hydrate Terraform state from the CM API (`GET api/v1/admin/policies/{id}` and `GET api/v1/admin/policy-attachments/{id}`).
- Replaces `Update()` stubs (which returned an error on every invocation) with full PATCH implementations using `c.UpdateDataV2(ctx, state.ID.ValueString(), URL_CONSTANT, payloadJSON)`.
- Fixes `CMPolicyJSON` and `CMPolicyAttachmentJSON` struct JSON tags — pointer types and `omitempty` on optional fields prevent zero-values (`false`, empty string) from being serialised into POST/PATCH payloads, correcting silently-wrong API requests.
- Adds `modifiers.ImmutableString()` to the `policy` attribute of `ciphertrust_policy_attachments`, surfacing at `terraform plan` time the constraint that the policy reference cannot be changed after creation.
- Adds 5 acceptance tests covering drift, update, and immutability for both resources.

## Motivation

Implements TFIN-278: Drift Detection for `ciphertrust_policies` and `ciphertrust_policy_attachments`.

Before this change, `terraform plan` after an out-of-band modification to either resource always reported no changes, because `Read()` returned immediately without querying the CM API. Users had no way to detect that CipherTrust Manager state had diverged from Terraform state without destroying and recreating the resource. Additionally, `Update()` was a stub that returned a hard error, forcing `terraform apply` to fail for any configuration change.

- `CMPolicyConditionJSON`: Changed `Negate bool` to `Negate *bool` with `omitempty`; added `omitempty` to `Op`, `Path`, and `Values`. Without pointer types, `false` was always serialised, sending `"negate": false` in every API call even when the field was not configured.
- `CMPolicyJSON`: Changed `Allow`, `IncludeDescendantAccounts` to `*bool` and `Effect`, `Name` to `*string`, all with `omitempty`. Changed `ID`, `URI`, `Account`, `CreatedAt` to `omitempty`. This ensures only user-set fields are sent to the CM API.
- `CMPolicyAttachmentJSON`: Added `omitempty` to all fields. Added the missing `Actions []string` and `Resources []string` fields — these were accepted by the API but could not previously be set.

Schema changes:
- `uri`, `account`, `created_at`: Added `stringplanmodifier.UseStateForUnknown()`. These fields are server-assigned at creation and stable for the resource lifetime; the modifier suppresses unnecessary `(known after apply)` noise on plans that do not modify the resource.

`Create()`:
- Fixed null-check idiom from comparing to zero-value (`plan.Allow.ValueBool() != types.BoolNull().ValueBool()`) to `!plan.Allow.IsNull() && !plan.Allow.IsUnknown()` — the prior form evaluated incorrectly for `false`.
- Changed the trace-UUID argument passed to `PostDataV2` from `plan.Name.ValueString()` to a freshly generated `id` — passing the resource name as a trace identifier was a category error.
- Added full post-response hydration: `effect`, `name`, `actions`, `resources`, `conditions` are now read from the API response and stored in state. `allow` and `include_descendant_accounts` are hydrated only when the plan set them, to avoid plan-consistency errors from CM echoing server-normalised defaults.

`Read()`:
- Was previously empty (returned immediately). Now calls `r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), common.URL_CM_POLICIES)` and hydrates all state fields.
- `id`, `uri`, `account`, `created_at`: hydrated unconditionally (Computed-only fields).
- `effect`, `name`: hydrated unconditionally with absent-branch setting `types.StringNull()`.
- `allow`: hydrated only when the prior state already held a non-null value, because CM normalises `allow` from `effect` and echoes it unconditionally — hydrating it for configs that only set `effect` would cause perpetual drift.
- `include_descendant_accounts`: preserved from state when CM omits the field from the GET response.
- `actions`, `resources`, `conditions`: hydrated unconditionally; set to `nil` when absent from the API response.
- 404 handling: changed from `resp.State.RemoveResource(ctx)` to adding a diagnostic warning and keeping the resource in state — see the Read() section below.

`Update()`:
- Was a stub returning `resp.Diagnostics.AddError("Updating Policy is not supported", ...)`. Now builds a `CMPolicyJSON` payload from the plan and calls `r.client.UpdateDataV2(ctx, state.ID.ValueString(), common.URL_CM_POLICIES, payloadJSON)`, then hydrates state from the PATCH response using the same field-level logic as `Create()`.

### Modified file — `internal/provider/cm/resource_policy_attachments.go`

Schema changes:
- `policy`: Added `modifiers.ImmutableString()` and updated `Description` to include `(Immutable)` prefix. The CM API has no PATCH endpoint that accepts a `policy` field; the attachment must be destroyed and recreated to change the target policy.
- `uri`, `account`, `created_at`: Added `stringplanmodifier.UseStateForUnknown()`.

`Create()`:
- Fixed null-check idiom for `Jurisdiction`.
- Added serialisation of `Actions` and `Resources` to the POST payload — these fields were previously silently dropped.
- Added per-field hydration from the POST response for `actions` and `resources`, guarded so that CM returning an empty or absent array does not overwrite a user-configured non-null list.

`Read()`:
- Was previously empty (returned immediately). Now calls `r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), common.URL_CM_POLICY_ATTACHMENTS)` and hydrates state.
- `id`, `uri`, `account`, `created_at`: hydrated unconditionally.
- `policy`: preserved from state unconditionally — CM transforms the short-name identifier to an internal URI in its GET response; reading it back would cause perpetual drift for configurations that set `policy` by name.
- `principal_selector`: hydrated when the field is present in the GET response; a warning is logged when absent.
- `jurisdiction`: hydrated only when state already holds a non-null value (user configured it), because CM auto-assigns a jurisdiction even when the field is not set.
- `actions`, `resources`: state is updated only when the API returns a non-empty array and state already holds a non-null value, working around a CM behaviour where GET may return an empty array for these fields even after they were stored at create time.
- 404 handling: changed from `resp.State.RemoveResource(ctx)` to a diagnostic warning, keeping state intact.

`Update()`:
- Was a stub returning an error. Now builds a `CMPolicyAttachmentJSON` payload from the plan (omitting `policy`, which is immutable) and calls `r.client.UpdateDataV2(ctx, state.ID.ValueString(), common.URL_CM_POLICY_ATTACHMENTS, payloadJSON)`.

`Delete()`:
- Added graceful 404 handling: if the resource has already been deleted out-of-band, `Delete()` logs and returns instead of surfacing an error.

- Adds `TestAccCMPolicy_drift`: creates a policy, modifies it directly in CM via the client, runs a `RefreshState` step, and asserts `ExpectNonEmptyPlan: true` — confirming Read() detects the drift.
- Adds `TestAccCMPolicy_update`: creates a policy, applies a configuration change (adds a second condition), and asserts the resource ID is stable (no destroy+recreate) and the final plan is clean.

- Updates existing out-of-band deletion test: adjusts expectations to match the new 404 behaviour — `Read()` now adds a warning and keeps state rather than removing the resource.
- Adds `TestAccCMPolicyAttachment_drift`: creates an attachment, modifies `actions` and `resources` directly in CM, runs a `RefreshState` step, and asserts `ExpectNonEmptyPlan: true`.
- Adds `TestAccCMPolicyAttachment_update`: creates an attachment, changes `principal_selector`, and asserts the resource ID is stable (no destroy+recreate).
- Adds `TestAccCMPolicyAttachment_immutablePolicy`: asserts that changing the `policy` attribute raises `"Attribute is immutable"` at plan time without making any API call.

## Read() now implemented

`Read()` previously returned immediately in both resources. This change implements it by calling `r.client.ReadDataByParam(ctx, id, state.ID.ValueString(), URL_CONSTANT)`, unmarshalling the CM API response with `gjson`, and repopulating Terraform state.

**Fields read from CM response — `ciphertrust_policies` (`GET api/v1/admin/policies/{id}`):**

| Terraform attribute | CM JSON field | Hydration rule |
|---|---|---|
| `uri` | `uri` | Unconditional |
| `account` | `account` | Unconditional |
| `created_at` | `createdAt` | Unconditional |
| `effect` | `effect` | Unconditional; `types.StringNull()` when absent |
| `name` | `name` | Unconditional; `types.StringNull()` when absent |
| `allow` | `allow` | Only when prior state is non-null (server normalises this from `effect`) |
| `include_descendant_accounts` | `include_descendant_accounts` | Preserved from state when CM omits the field |
| `actions` | `actions` | Unconditional; `nil` when absent |
| `resources` | `resources` | Unconditional; `nil` when absent |
| `conditions[*]` | `conditions[*]` | Unconditional; `nil` when absent |

**Fields read from CM response — `ciphertrust_policy_attachments` (`GET api/v1/admin/policy-attachments/{id}`):**

| Terraform attribute | CM JSON field | Hydration rule |
|---|---|---|
| `uri` | `uri` | Unconditional |
| `account` | `account` | Unconditional |
| `created_at` | `createdAt` | Unconditional |
| `policy` | — | Preserved from state (CM returns internal URI, not user-supplied name) |
| `principal_selector` | `principalSelector` | When present; warning logged when absent |
| `jurisdiction` | `jurisdiction` | Only when prior state is non-null |
| `actions` | `actions` | Only when API returns a non-empty array and prior state is non-null |
| `resources` | `resources` | Only when API returns a non-empty array and prior state is non-null |

**404 handling:** A 404 response from CM adds a `resp.Diagnostics.AddWarning` and returns — it does NOT call `resp.State.RemoveResource(ctx)`. This preserves the resource in state so users can recover by importing or running `terraform state rm` explicitly, rather than having Terraform silently forget the resource and propose an unexpected recreate on the next `apply`.

### `ciphertrust_policies`

|---|---|---|---|
| `id` | `types.String` | Computed | `UseStateForUnknown`; set from CM response |
| `name` | `types.String` | Optional | Hydrated from API on Read |
| `actions` | `types.List(String)` | Optional | |
| `allow` | `types.Bool` | Optional | Hydrated from API only when prior state is non-null |
| `conditions` | Block list | Optional | Per-condition: `negate`, `op`, `path`, `values` |
| `effect` | `types.String` | Optional | Hydrated unconditionally |
| `include_descendant_accounts` | `types.Bool` | Optional | Preserved from state when API omits field |
| `resources` | `types.List(String)` | Optional | |
| `uri` | `types.String` | Computed | `UseStateForUnknown` |
| `account` | `types.String` | Computed | `UseStateForUnknown` |
| `created_at` | `types.String` | Computed | `UseStateForUnknown` |

### `ciphertrust_policy_attachments`

|---|---|---|---|
| `id` | `types.String` | Computed | `UseStateForUnknown` |
| `policy` | `types.String` | Required | `modifiers.ImmutableString()` — cannot be changed after creation |
| `principal_selector` | `types.Map(String)` | Required | |
| `jurisdiction` | `types.String` | Optional | |
| `actions` | `types.List(String)` | Optional+Computed | |
| `resources` | `types.List(String)` | Optional+Computed | |
| `uri` | `types.String` | Computed | `UseStateForUnknown` |
| `account` | `types.String` | Computed | `UseStateForUnknown` |
| `created_at` | `types.String` | Computed | `UseStateForUnknown` |

## Swagger note — no PATCH operation documented

The swagger spec lists only `GET`, `POST`, and `DELETE` for both `/v1/admin/policies/{id}` and `/v1/admin/policy-attachments/{id}`. There is no documented `PATCH` operation for either endpoint. The `Update()` implementations call `PATCH` via `UpdateDataV2`. This is consistent with CM's general API pattern (many resources accept PATCH even when not listed in the public spec), and has been verified to work in live testing. Reviewers who have access to a CM instance should confirm the PATCH endpoint is functional for both resource types.

## How to test

  Drift detection for `ciphertrust_policies`:
  ```
  go test ./internal/provider/ -run TestAccCMPolicy_drift -v -timeout 15m
  ```

  Update for `ciphertrust_policies`:
  ```
  go test ./internal/provider/ -run TestAccCMPolicy_update -v -timeout 15m
  ```

  Drift detection for `ciphertrust_policy_attachments`:
  ```
  go test ./internal/provider/ -run TestAccCMPolicyAttachment_drift -v -timeout 15m
  ```

  Update for `ciphertrust_policy_attachments`:
  ```
  go test ./internal/provider/ -run TestAccCMPolicyAttachment_update -v -timeout 15m
  ```

  Immutable `policy` field:
  ```
  go test ./internal/provider/ -run TestAccCMPolicyAttachment_immutablePolicy -v -timeout 15m
  ```

  Scenarios covered:
  - **Drift (policy)** — create resource; mutate directly in CM; `RefreshState`; assert non-empty plan.
  - **Update (policy)** — create resource; add a second condition; re-apply; assert ID stable (no destroy+recreate); assert clean subsequent plan.
  - **Drift (attachment)** — create attachment; mutate `actions`/`resources` directly in CM; `RefreshState`; assert non-empty plan.
  - **Update (attachment)** — create attachment; change `principal_selector`; re-apply; assert ID stable; assert clean subsequent plan.
  - **ImmutablePolicy** — create attachment; change `policy` in config; `terraform plan`; assert `"Attribute is immutable"` error appears at plan time before any API call.
  - **OOB deletion (updated)** — delete attachment directly in CM; `RefreshState`; assert resource is still in state with a diagnostic warning (no `RemoveResource`).

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[<file>.go -> <Func>][uuid]` pattern — `defer` is used for exit traces.
- [ ] CM API GET endpoints verified against swagger spec; PATCH endpoints not listed in swagger — see Swagger note above.

---

### DIFF
--- a/internal/provider/cm/resource_policy.go
+++ b/internal/provider/cm/resource_policy.go
@@ -101,9 +101,24 @@ func (r *resourceCMPolicy) Schema(_ context.Context, _ resource.SchemaRequest, r
 				Description: "Resources is a list of URI strings, which must be in URI format.",
 				ElementType: types.StringType,
 			},
-			"uri":        schema.StringAttribute{Computed: true},
-			"account":    schema.StringAttribute{Computed: true},
-			"created_at": schema.StringAttribute{Computed: true},
+			"uri": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"account": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+			"created_at": schema.StringAttribute{
+				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
 		},
 	}
 }
@@ -112,8 +127,8 @@ func (r *resourceCMPolicy) Schema(_ context.Context, _ resource.SchemaRequest, r
 func (r *resourceCMPolicy) Create(ctx context.Context, req resource.CreateRequest, resp *resource.CreateResponse) {
 	id := uuid.New().String()
 	tflog.Trace(ctx, common.MSG_METHOD_START+"[resource_policy.go -> Create]["+id+"]")
+	defer tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_policy.go -> Create]["+id+"]")
 
-	// Retrieve values from plan
 	var plan CMPolicyTFSDK
 	var payload CMPolicyJSON
 
@@ -129,20 +144,22 @@ func (r *resourceCMPolicy) Create(ctx context.Context, req resource.CreateReques
 	}
 	payload.Actions = actions
 
-	if plan.Allow.ValueBool() != types.BoolNull().ValueBool() {
-		payload.Allow = plan.Allow.ValueBool()
+	if !plan.Allow.IsNull() && !plan.Allow.IsUnknown() {
+		v := plan.Allow.ValueBool()
+		payload.Allow = &v
 	}
 
 	var conditions []CMPolicyConditionJSON
 	for _, condition := range plan.Conditions {
 		var conditionJSON CMPolicyConditionJSON
-		if condition.Negate.ValueBool() != types.BoolNull().ValueBool() {
-			conditionJSON.Negate = condition.Negate.ValueBool()
+		if !condition.Negate.IsNull() && !condition.Negate.IsUnknown() {
+			v := condition.Negate.ValueBool()
+			conditionJSON.Negate = &v
 		}
-		if condition.Op.ValueString() != "" && condition.Op.ValueString() != types.StringNull().ValueString() {
+		if !condition.Op.IsNull() && !condition.Op.IsUnknown() {
 			conditionJSON.Op = condition.Op.ValueString()
 		}
-		if condition.Path.ValueString() != "" && condition.Path.ValueString() != types.StringNull().ValueString() {
+		if !condition.Path.IsNull() && !condition.Path.IsUnknown() {
 			conditionJSON.Path = condition.Path.ValueString()
 		}
 		var values []string
@@ -150,21 +167,23 @@ func (r *resourceCMPolicy) Create(ctx context.Context, req resource.CreateReques
 			values = append(values, str.ValueString())
 		}
 		conditionJSON.Values = values
-
 		conditions = append(conditions, conditionJSON)
 	}
 	payload.Conditions = conditions
 
-	if plan.Effect.ValueString() != "" && plan.Effect.ValueString() != types.StringNull().ValueString() {
-		payload.Effect = plan.Effect.ValueString()
+	if !plan.Effect.IsNull() && !plan.Effect.IsUnknown() {
+		v := plan.Effect.ValueString()
+		payload.Effect = &v
 	}
 
-	if plan.IncludeDescendantAccounts.ValueBool() != types.BoolNull().ValueBool() {
-		payload.IncludeDescendantAccounts = plan.IncludeDescendantAccounts.ValueBool()
+	if !plan.IncludeDescendantAccounts.IsNull() && !plan.IncludeDescendantAccounts.IsUnknown() {
+		v := plan.IncludeDescendantAccounts.ValueBool()
+		payload.IncludeDescendantAccounts = &v
 	}
 
-	if plan.Name.ValueString() != "" && plan.Name.ValueString() != types.StringNull().ValueString() {
-		payload.Name = plan.Name.ValueString()
+	if !plan.Name.IsNull() && !plan.Name.IsUnknown() {
+		v := plan.Name.ValueString()
+		payload.Name = &v
 	}
 
 	var resources []string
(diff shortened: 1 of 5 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-278

rohitpandita-hub:
@terraform-agent test "TestAccCMPolicy_update" is failing, fix it

ashwani573:
Addressed in `897fbf7`: TFIN-278: address review feedback (round 1)

PR #271, opened 2026-07-03
TFIN-275: Drift Detection |  ciphertrust_cm_ssh_key, ciphertrust_cm_reg_token

ashwani573:
## Summary

- Implements drift detection for `ciphertrust_cm_reg_token`: `Read()` now calls `GET /v1/client-management/regtokens/{id}` via `c.GetById` and repopulates Terraform state from the CM API response.
- Documents the intentional no-op `Read()` for `ciphertrust_cm_ssh_key` with an `AddWarning` diagnostic: `CMClientBootstrap` does not expose `GetById` and the CM SSH key endpoint (`POST /v1/system/ssh/keys`) has no per-resource `GET`.
- Fixes three pre-existing bugs in `ciphertrust_cm_reg_token`: a `label`/`labels` iterator bug in `Create()` where both map blocks incorrectly iterated `plan.Labels`; a `plan.ID` → `state.ID` URL path key bug in `Update()`; and missing null guards for Optional map fields throughout `Create()` and `Update()`.
- Adds `resource_cm_ssh_key_test.go` (new) with `TestCipherTrust_CMSSHKey_noopRead` and extends `resource_cm_reg_token_test.go` with six new `TestCipherTrust_CMRegToken_*` acceptance tests covering drift, out-of-band delete, label create/drift, labels drift, and name-prefix drift scenarios.

## Motivation

Implements TFIN-275: Drift Detection for `ciphertrust_cm_ssh_key` and `ciphertrust_cm_reg_token`.

**Read() — now implemented**

- Loads prior state; calls `c.GetById(ctx, id, state.ID.ValueString(), common.URL_REG_TOKEN)` against `GET /v1/client-management/regtokens/{id}` (swagger-confirmed in the `cm-misc` area).
- On `notFoundError` (404): calls `resp.State.RemoveResource(ctx)` and returns. This is an intentional, documented deviation from the provider-wide keep-in-state convention — registration tokens are ephemeral (they expire or are consumed), so a 404 means the token is gone and Terraform should recreate it on the next apply.
- Computed-only fields (`id`, `token`) are hydrated unconditionally from the response.
- Optional string/int64 scalars (`ca_id`, `client_management_profile_id`, `lifetime`, `name_prefix`, `cert_duration`, `max_clients`) are hydrated only when the prior state value is non-null, preventing null-to-zero-value drift for unconfigured fields where CM echoes a zero default.
- Optional map fields (`label`, `labels`) use a three-branch pattern: absent/null in response → `types.MapNull`; empty object → `types.MapValueMust({})` (preserving a user-configured empty map); present with entries → `types.MapValueFrom`.

**Create() — label iterator bug fixed**

Previously both the `label` and `labels` blocks iterated `plan.Labels`. The `label` block now correctly iterates `plan.Label`. Both blocks also gained null-and-unknown guards so an unconfigured map field no longer sends an empty object `{}` to the CM API.

**Update() — URL path key bug fixed; missing null guards added**

- `UpdateData` now passes `state.ID.ValueString()` as the path key instead of `plan.ID.ValueString()`. `plan.ID` is unknown at update time when `UseStateForUnknown` is in effect; using it caused a 404 on every `terraform apply` that triggered an update.
- Added null guards for `label`, `labels`, and `name_prefix` map/string fields to avoid sending zero values to the CM PATCH endpoint.

**Delete() — UUID trace logging and 404 guard added**

- `id := uuid.New().String()` added at the top with `tflog.Trace` entry/exit.
- 404 from CM is now treated as success (token already gone); other errors surface via `resp.Diagnostics.AddError`.

**Read() — documented intentional no-op**

The previous `Read()` body was empty (function signature only, no code). It is now a documented no-op:

- Loads state from `req.State`.
- Emits `tflog.Trace` entry/exit with a fresh `uuid.New().String()`.
- Calls `resp.Diagnostics.AddWarning("Drift Detection Not Supported", ...)` to make the limitation visible to operators running `terraform plan` or `terraform refresh`.
- Writes the unchanged state back via `resp.State.Set(ctx, state)`.

The underlying reason is architectural: `resourceCMSSHKey` uses `common.CMClientBootstrap` (not the standard `common.Client`). `CMClientBootstrap` exposes only `PostDataBootstrap` and does not implement `GetById`. Additionally, the CM swagger spec provides only `POST /v1/system/ssh/keys` (create) for this endpoint — there is no `GET /v1/system/ssh/keys/{id}` that would allow per-resource state reconciliation.

Six new acceptance test functions added:

| Function | Scenario |
|---|---|
| `TestCipherTrust_CMRegToken_drift` | PATCH token out-of-band → `RefreshState` → assert `ExpectNonEmptyPlan` |
| `TestCipherTrust_CMRegToken_deleteOOB` | DELETE token out-of-band → `RefreshState` → assert removed from state → re-apply creates new token |
| `TestCipherTrust_CMRegToken_labelCreate` | Create with `label` and `labels` maps → assert values; second refresh confirms no false drift |
| `TestCipherTrust_CMRegToken_labelDrift` | Mutate `label` via PATCH out-of-band → `RefreshState` → assert `ExpectNonEmptyPlan` |
| `TestCipherTrust_CMRegToken_labelsDrift` | Mutate `labels` via PATCH out-of-band → `RefreshState` → assert `ExpectNonEmptyPlan` |
| `TestCipherTrust_CMRegToken_namePrefixDrift` | Mutate `name_prefix` via PATCH out-of-band → `RefreshState` → assert `ExpectNonEmptyPlan` |

### New file — `internal/provider/resource_cm_ssh_key_test.go`

`TestCipherTrust_CMSSHKey_noopRead`: applies a `ciphertrust_cm_ssh_key` resource (SSH public key read from `TEST_SSH_PUBLIC_KEY` env var), then verifies that a `RefreshState` step produces no diff and no error — confirming the no-op `Read()` is stable.

Uses `bootstrapProviderConfig()` (defined in the same file) which constructs the provider block with `bootstrap = "yes"` from `CIPHERTRUST_ADDRESS` / `CIPHERTRUST_USERNAME` / `CIPHERTRUST_PASSWORD` / `CIPHERTRUST_CA_CERT` env vars.

### `ciphertrust_cm_ssh_key` — intentional no-op

`Read()` preserves Terraform state unchanged on every invocation and emits a `"Drift Detection Not Supported"` warning. It does not call the CM API. `CMClientBootstrap` does not implement `GetById`, and the CM REST API has no `GET /v1/system/ssh/keys/{id}` endpoint (swagger-confirmed: only `POST /v1/system/ssh/keys` and `GET /v1/system/ssh/kex` exist in the SSH area). The ticket description acknowledged this limitation and did not ask for a full Read() implementation on this resource.

**User-visible consequence:** After `terraform apply`, subsequent `terraform plan` / `terraform refresh` runs will always show no changes for this resource — even if the SSH key was modified or removed directly in CipherTrust Manager. Each refresh will also emit a `Warning: Drift Detection Not Supported` diagnostic. This is a known limitation of the bootstrap-only client design.

### `ciphertrust_cm_reg_token` — fully implemented

`Read()` previously returned immediately after loading state (no API call). It now calls `c.GetById(ctx, id, state.ID.ValueString(), common.URL_REG_TOKEN)` and repopulates state from the CM API response.

**Fields read from CM response** (per swagger `GET /v1/client-management/regtokens/{id}`):

- `id` ← `id` (unconditional)
- `token` ← `token` (unconditional)
- `ca_id` ← `ca_id` (guarded: only when prior state is non-null)
- `client_management_profile_id` ← `client_management_profile_id` (guarded)
- `lifetime` ← `lifetime` (guarded)
- `name_prefix` ← `name_prefix` (guarded)
- `cert_duration` ← `cert_duration` (guarded)
- `max_clients` ← `max_clients` (guarded)
- `label` ← `label` (three-branch map hydration)
- `labels` ← `labels` (three-branch map hydration)

**404 handling:** A 404 response calls `resp.State.RemoveResource(ctx)`, removing the token from Terraform state. On the next `terraform apply` Terraform will create a new token. This deviates from the provider-wide keep-in-state convention, but is intentional here: registration tokens are ephemeral — they expire after their `lifetime` or are consumed by client registration. Keeping an expired/consumed token in state would cause every subsequent operation to fail with a 404 until the user manually removed it.

**Null/absent fields:** Optional scalar fields guarded with `!state.X.IsNull()` — when CM does not return a field (or returns a zero value for an unset int), the field is left at its prior state value, preventing perpetual drift for users who never configured the field.

## How to test

- [ ] Acceptance tests for `ciphertrust_cm_reg_token` (requires live CM — set `TF_ACC=1` and `CIPHERTRUST_*` env vars):
  ```
  go test ./internal/provider/ -run 'TestCipherTrust_CMRegToken' -v -timeout 15m
  ```
  Scenarios covered:
  - **Drift detection** — PATCH token out-of-band; `RefreshState` detects diff.
  - **Out-of-band delete** — DELETE token out-of-band; `RefreshState` removes resource from state; re-apply creates a new token with a different ID.
  - **Label create** — map fields round-trip correctly; no false drift on repeated refresh.
  - **Label drift** — `label` map mutated out-of-band; drift detected.
  - **Labels drift** — `labels` map mutated out-of-band; drift detected.
  - **name_prefix drift** — `name_prefix` mutated out-of-band; drift detected.
- [ ] Acceptance test for `ciphertrust_cm_ssh_key` (requires bootstrap-mode CM instance and `TEST_SSH_PUBLIC_KEY` env var):
  ```
  go test ./internal/provider/ -run 'TestCipherTrust_CMSSHKey_noopRead' -v -timeout 15m
  ```
  Scenario: apply resource; verify `id` is set; refresh produces no diff and no error.

## Checklist

- [ ] Fresh `uuid.New().String()` at the top of each Read/Delete method — not reused across calls.
- [ ] CM API endpoints verified against swagger spec: `GET /v1/client-management/regtokens/{id}` (Read), `PATCH /v1/client-management/regtokens/{id}` (Update), `DELETE /v1/client-management/regtokens/{id}` (Delete) all confirmed present.
- [ ] `ciphertrust_cm_ssh_key` no-op rationale is documented in code comments and surfaces a `AddWarning` diagnostic to operators.
- [ ] `notFoundError` used via the package-level constant from `cm/constants.go` — no redeclaration.

---

### DIFF
--- a/internal/provider/cm/resource_cm_reg_token.go
+++ b/internal/provider/cm/resource_cm_reg_token.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/google/uuid"
 
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
@@ -114,19 +116,23 @@ func (r *resourceCMRegToken) Create(ctx context.Context, req resource.CreateRequ
 		payload.ClientManagementProfileID = plan.ClientManagementProfileID.ValueString()
 	}
 
-	// Add label to payload
-	labelPayload := make(map[string]interface{})
-	for k, v := range plan.Labels.Elements() {
-		labelPayload[k] = v.(types.String).ValueString()
+	// Add label to payload — fix: both blocks previously iterated plan.Labels (bug); first block now iterates plan.Label
+	if !plan.Label.IsNull() && !plan.Label.IsUnknown() {
+		labelPayload := make(map[string]interface{})
+		for k, v := range plan.Label.Elements() {
+			labelPayload[k] = v.(types.String).ValueString()
+		}
+		payload.Label = labelPayload
 	}
-	payload.Label = labelPayload
 
-	// Add labels to payload
-	labelsPayload := make(map[string]interface{})
-	for k, v := range plan.Labels.Elements() {
-		labelsPayload[k] = v.(types.String).ValueString()
+	// Add labels to payload — null guard prevents sending {} when unconfigured
+	if !plan.Labels.IsNull() && !plan.Labels.IsUnknown() {
+		labelsPayload := make(map[string]interface{})
+		for k, v := range plan.Labels.Elements() {
+			labelsPayload[k] = v.(types.String).ValueString()
+		}
+		payload.Labels = labelsPayload
 	}
-	payload.Labels = labelsPayload
 
 	if plan.Lifetime.ValueString() != "" && plan.Lifetime.ValueString() != types.StringNull().ValueString() {
 		payload.Lifetime = plan.Lifetime.ValueString()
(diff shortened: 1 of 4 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-275

rohitpandita-hub:
@terraform-agent test "TestCipherTrust_CMRegToken_drift" is failing, fix it

ashwani573:
Addressed in `24d8151`: TFIN-275: address review feedback (round 1)

PR #272, opened 2026-07-03
TFIN-340: Drift Detection |   ciphertrust_hsm_rot, ciphertrust_cm_user_pwd_change

ashwani573:
## Summary

- Implements `Read()` for `ciphertrust_hsm_root_of_trust_setup`: captures the `GET /v1/system/hsm/servers/{id}` response and hydrates all state fields so that out-of-band changes to the HSM server record are surfaced on `terraform plan`.
- Adds immutability enforcement to all user-configurable fields on `ciphertrust_hsm_root_of_trust_setup` using `modifiers.ImmutableString/Map/Bool/Int64()` plan modifiers; adds `UseStateForUnknown()` to the two Computed fields `sub_type` and `config`.
- Extends `internal/provider/modifiers/immutable.go` with a new `ImmutableMap()` modifier required by the `conn_info` and `initial_config` map attributes.
- Adds a design-intent comment to the empty `Read()` on `ciphertrust_cm_user_password_change` explaining why it is and must remain a no-op.
- Adds 5 acceptance-test functions to `internal/provider/resource_hsm_rot_test.go` covering no-drift, immutable-field enforcement, and destroy-when-already-gone scenarios.

## Motivation

Implements TFIN-340: Drift Detection for `ciphertrust_hsm_root_of_trust_setup` and `ciphertrust_cm_user_password_change`.

Without a working `Read()`, `terraform plan` always reported no changes for `ciphertrust_hsm_root_of_trust_setup` even when the CM record had been modified or removed outside Terraform. This PR fixes that for `ciphertrust_hsm_root_of_trust_setup` and documents the intentional exception for `ciphertrust_cm_user_password_change`.

**Read():**
Previously called `c.GetById(ctx, id, state.ID.ValueString(), common.URL_HSM_Server)` but discarded the response (`_`). The method returned immediately with no state writes, so every `terraform plan` after `terraform apply` showed no diff regardless of what had changed in CM.

After this change, the full response body is captured and state is hydrated as follows:
- `id`, `sub_type` — Computed-only; hydrated unconditionally.
- `config` — Computed-only; hydrated via the updated `parseConfig()` helper.
- `conn_info` — Required; the CM response encodes this field as a JSON string (`connInfo`); deserialised with `json.Unmarshal` back into a `map[string]string` and stored via `types.MapValueFrom`.
- `initial_config` — Optional; the CM response encodes this as a JSON object (`initialConfig`); iterated with `gjson.ForEach`.
- `reset` — Optional; hydrated with `r.Exists()` absent-branch (`types.BoolNull()` when absent).
- `delay` — Optional; hydrated with `r.Exists()` absent-branch (`types.Int64Null()` when absent).
- `type` — Required and immutable; preserved from prior state, not overwritten from the response. The CM API may normalise the casing of the returned type value, so re-assigning it would risk a perpetual plan diff for any user who capitalised the value in their `.tf` file.
- On 404: `resp.State.RemoveResource(ctx)` is called along with `AddWarning`. See the Read() section below for the rationale behind this deviation from the standard keep-in-state convention.
- The inline `"status: 404"` string literal is replaced with the package-level `notFoundError` constant from `cm/constants.go`.

**parseConfig():**
- Signature changes from 2-arg `(response string, diags *diag.Diagnostics)` to 3-arg `(ctx context.Context, response string, diags *diag.Diagnostics)` to support structured logging.
- Implementation changes from `json.Unmarshal` to a `gjson.ForEach` loop so that non-string config values (integers, booleans) are serialised via `fmt.Sprintf("%v", ...)` rather than panicking.
- When the `config` field is absent from the response the function now returns an empty `types.Map` rather than an error, because some HSM types legitimately omit `config` in their setup response.

**Create():**
- Applies `strings.ToLower()` to `plan.Type` before writing it to state. The CM API returns the HSM type in lowercase; without this normalisation a user who capitalises the type value would see a perpetual plan diff after apply.
- Updates the `parseConfig` call to pass `ctx`.

**Delete():**
- Adds a `notFoundError` 404 early-return guard: if the CM object has already been removed out-of-band, `Delete()` exits cleanly instead of returning an error.

**Schema changes:**
- `type`: adds `modifiers.ImmutableString()` and `stringvalidator.OneOf("luna","lunapci","lunatct","protectserver","aws","dpod","nshield","ibmhpcs")`; description prefixed with `(Immutable)`.
- `conn_info`: adds `modifiers.ImmutableMap()`; description prefixed with `(Immutable)`.
- `initial_config`: adds `modifiers.ImmutableMap()`; description prefixed with `(Immutable)`.
- `reset`: adds `modifiers.ImmutableBool()`; description prefixed with `(Immutable)`.
- `delay`: adds `modifiers.ImmutableInt64()`; description prefixed with `(Immutable)`.
- `sub_type`: adds `stringplanmodifier.UseStateForUnknown()` so the value set at create time is not shown as `(known after apply)` on subsequent plans.
- `config`: adds `mapplanmodifier.UseStateForUnknown()` for the same reason.

Adds `ImmutableMap() planmodifier.Map` and the unexported `immutableMapModifier` struct. Follows the same pattern as the existing `ImmutableString`, `ImmutableInt64`, `ImmutableBool`, and `ImmutableList` functions. At plan time, if `StateValue` is non-null and `PlanValue` differs, it appends an `AddError` diagnostic and reverts `PlanValue` to `StateValue`. No destroy+recreate is triggered.

Adds a block comment inside the empty `Read()` body explaining that `ciphertrust_cm_user_password_change` is a one-shot action resource: it maps to `PATCH /v1/auth/changepw`, which changes a password credential. The CM API exposes no `GET` endpoint for password-change records. There is no state to reconcile. No logic change.

Adds helper functions `hsmRotConfig`, `hsmRotConfigWithType`, `hsmRotConfigWithConnInfo`, and `hsmRotConfigWithReset` to reduce HCL duplication across test steps.

Adds five acceptance test functions:

- `TestCipherTrust_HSMRot_NoDrift` — applies the resource, then calls `RefreshState: true` + `ExpectNonEmptyPlan: false` to verify `Read()` produces state consistent with what was just applied.
- `TestCipherTrust_HSMRot_ImmutableType` — applies the resource, then in a second step changes `type` with `PlanOnly: true`; asserts that a plan-time diagnostic error fires rather than a destroy+recreate plan being produced.
- `TestCipherTrust_HSMRot_ImmutableConnInfo` — same pattern for `conn_info`.
- `TestCipherTrust_HSMRot_ImmutableReset` — same pattern for `reset`.
- `TestCipherTrust_HSMRot_DestroyNotFound` — applies the resource, deletes the CM object out-of-band using the client's `DeleteByID` helper, then runs `terraform destroy`; asserts no error is returned (exercises the new 404 guard in `Delete()`).

## Schema attributes — `ciphertrust_hsm_root_of_trust_setup`

No attributes are added or removed. The table below lists only the attributes that change in this PR.

|---|---|---|---|
| `type` | `types.String` | Required | Adds `modifiers.ImmutableString()` + `stringvalidator.OneOf(...)`; `(Immutable)` description prefix |
| `conn_info` | `types.Map(String)` | Required | Adds `modifiers.ImmutableMap()`; `(Immutable)` description prefix |
| `initial_config` | `types.Map(String)` | Optional | Adds `modifiers.ImmutableMap()`; `(Immutable)` description prefix |
| `reset` | `types.Bool` | Optional | Adds `modifiers.ImmutableBool()`; `(Immutable)` description prefix |
| `delay` | `types.Int64` | Optional | Adds `modifiers.ImmutableInt64()`; `(Immutable)` description prefix |
| `sub_type` | `types.String` | Computed | Adds `stringplanmodifier.UseStateForUnknown()` |
| `config` | `types.Map(String)` | Computed | Adds `mapplanmodifier.UseStateForUnknown()` |

## Read() now implemented — `ciphertrust_hsm_root_of_trust_setup`

`Read()` previously discarded the response from `c.GetById`. This change implements full state hydration using `GET /v1/system/hsm/servers/{id}` (swagger-confirmed: `cm-system` area, operation "HSM Servers — Get").

| Terraform attribute | CM JSON field | Hydration pattern |
|---|---|---|
| `sub_type` | `sub_type` | Unconditional |
| `config` | `config` | Via `parseConfig()` + `gjson.ForEach` |
| `conn_info` | `connInfo` | JSON string deserialized with `json.Unmarshal` → `types.MapValueFrom` |
| `initial_config` | `initialConfig` | JSON object iterated with `gjson.ForEach` |
| `reset` | `reset` | `r.Exists()` absent-branch → `types.BoolNull()` |
| `delay` | `delay` | `r.Exists()` absent-branch → `types.Int64Null()` |
| `type` | — | Preserved from prior state; not overwritten from response |

**404 handling — documented deviation from the standard convention:** A 404 response calls `resp.State.RemoveResource(ctx)` and emits a warning diagnostic. Most resources in this provider keep the resource in state on a 404 to survive transient CM unavailability. HSM root-of-trust setup is treated differently because the operation is irreversible and destructive: if the CM record for the setup has disappeared, the appliance is in an indeterminate state that Terraform cannot reconcile by re-applying. Surfacing the missing record immediately via `RemoveResource + AddWarning` is safer than preserving stale state and allowing subsequent operations to proceed against a non-existent HSM setup.

**Null/absent fields:** Optional fields (`reset`, `delay`) absent from the CM response are written as typed null values. Optional map fields (`initial_config`, `conn_info`) absent from or empty in the CM response are written as `types.MapNull(types.StringType)`.

## Read() is intentionally empty — `ciphertrust_cm_user_password_change`

The `Read()` method returns immediately without calling the CM API. `ciphertrust_cm_user_password_change` maps to `PATCH /v1/auth/changepw`, a one-shot credential-rotation action. The CM API provides no `GET` endpoint for password-change records. There is no state to compare or reconcile.

**User-visible consequence:** After `terraform apply`, subsequent `terraform plan` runs will always show no changes for this resource, even if the password was changed again in CM outside of Terraform. This is a fundamental property of the CM API for this endpoint.

## Breaking changes

All five user-configurable fields on `ciphertrust_hsm_root_of_trust_setup` (`type`, `conn_info`, `initial_config`, `reset`, `delay`) now carry `modifiers.ImmutableX()` plan modifiers. Any `.tf` configuration that attempts to change one of these fields after the resource was created will receive a plan-time diagnostic error:

> Attribute is immutable — This attribute cannot be changed after creation. To change this attribute, destroy and recreate the resource.

Previously, changing these fields would have silently reached `Update()`, which immediately returned an `"Update Not Supported"` error at apply time. The new behaviour surfaces the constraint at plan time, before any API call is made.

Note: `modifiers.ImmutableX()` does NOT schedule destroy+recreate (`RequiresReplace`). The user must run `terraform destroy` then `terraform apply` manually if they need to change an immutable field.

## How to test

- [ ] Acceptance tests require a live CM instance **and** a physical HSM appliance (set `TF_ACC=1` and `CIPHERTRUST_*` env vars). All five new test functions are guarded by a skip because they require real HSM hardware:
  ```
  go test ./internal/provider/ -run 'TestCipherTrust_HSMRot' -v -timeout 15m
  ```
  Scenarios covered:
  - **NoDrift** — apply; refresh state; assert no diff produced.
  - **ImmutableType** — apply; change `type` in config; plan must return a diagnostic error.
  - **ImmutableConnInfo** — apply; change `conn_info`; plan must return a diagnostic error.
  - **ImmutableReset** — apply; change `reset`; plan must return a diagnostic error.
  - **DestroyNotFound** — apply; delete CM object out-of-band; `terraform destroy` must succeed without error.
- [ ] Manual smoke-test note: `TestCipherTrust_HSMRot_ImmutableDelay` is not included in this PR. Reviewers may request it as a follow-up.
- [ ] No acceptance test for `ciphertrust_cm_user_password_change` — the empty `Read()` is unchanged in behaviour; no new code paths to cover.

## Checklist

- [ ] Schema attribute `Description` fields populated — all modified attributes have non-empty descriptions; immutable fields carry the `(Immutable)` prefix required for generated docs.
- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_hsm_rot.go -> <Func>][uuid]` pattern; `Read()` uses `defer` for the exit trace.
- [ ] URL constants used — `URL_HSM_Server` and `URL_HSM_SETUP` were already defined in `urls.go`; no new inlined string literals.
- [ ] CM API endpoint `GET /v1/system/hsm/servers/{id}` verified against swagger (`cm-system` area). `POST /v1/system/hsm/setup` and `DELETE /v1/system/hsm/servers/{id}` also confirmed in the same area.
- [ ] `Update()` returns an explicit error ("Update Not Supported") immediately — all non-computed fields are now also blocked at plan time by `ImmutableX()` modifiers.
- [ ] `Read()` 404 calls `resp.State.RemoveResource(ctx)` for `ciphertrust_hsm_root_of_trust_setup` — deliberate, documented deviation from the general conservative-404 convention.

---

### DIFF
--- a/internal/provider/cm/resource_cm_user_pwd_change.go
+++ b/internal/provider/cm/resource_cm_user_pwd_change.go
@@ -98,6 +98,14 @@ func (r *resourceCMPwdChange) Create(ctx context.Context, req resource.CreateReq
 
 // Read refreshes the Terraform state with the latest data.
 func (r *resourceCMPwdChange) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
+	// Intentionally empty. ciphertrust_cm_user_password_change is a one-shot action
+	// resource: it triggers a CM password change and has no retrievable state.
+	// The CM API provides no GET endpoint for password-change records.
+	//
+	// User-visible consequence: after the initial `terraform apply`, subsequent
+	// `terraform plan` runs will always show no changes — even if the password
+	// was changed or reset in CM outside of Terraform. This is a known,
+	// intentional limitation documented in TFIN-DD-015.
 }
 
 // Update updates the resource and sets the updated Terraform state on success.
--- a/internal/provider/cm/resource_hsm_rot.go
+++ b/internal/provider/cm/resource_hsm_rot.go
@@ -1,22 +1,25 @@
 package cm
 
 import (
-	"strings"
 	"context"
 	"encoding/json"
 	"fmt"
-
-	"github.com/hashicorp/terraform-plugin-framework/attr"
-	"github.com/hashicorp/terraform-plugin-framework/diag"
-	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"strings"
 
 	"github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
+	"github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/modifiers"
 	"github.com/google/uuid"
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/mapplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-log/tflog"
 	"github.com/tidwall/gjson"
 )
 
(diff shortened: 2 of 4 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-340

PR #301, opened 2026-07-09
TFIN-167: terraform plan silently shows "no changes" when CM resources are deleted or modified out-of-band due to empty Read() implementation

ashwani573:
## Summary

- Implements `Read()` for `ciphertrust_cm_ssh_key` and `ciphertrust_cm_user_password_change`, replacing stubs that previously preserved state unchanged without contacting the CM API.
- Adds `GetByIdBootstrap(ctx, uuid, id, endpoint)` to `CMClientBootstrap` in `internal/provider/common/requests.go` — a new `GET {baseURL}/{endpoint}/{id}` helper used by both resources' `Read()` and `Create()`.
- Expands the `ciphertrust_cm_ssh_key` schema with 9 new attributes (`name`, `algorithm`, `key_size`, `curve`, `username`, `public_key_encoding`, `fingerprint`, `created_at`, `updated_at`) and updates `Create()` to perform a GET read-back after the initial POST to populate all Computed fields from the server response.
- Adds an `id` attribute to `ciphertrust_cm_user_password_change` and stores the CM-assigned `user_id` in state during `Create()` so that `Read()` can perform 404 detection against `GET api/v1/usermgmt/users/{user_id}`.
- Adds acceptance tests: `TestCipherTrust_CMSSHKey_OOBDelete`, `TestCipherTrust_CMSSHKey_AttributeDrift` in `resource_cm_ssh_key_test.go`; `TestCipherTrust_CMUserPwdChange_ReadStability` in `resource_cm_user_pwd_change_test.go`.

## Motivation

Implements TFIN-167: `terraform plan` silently showed "no changes" when CM resources managed by `ciphertrust_cm_ssh_key` and `ciphertrust_cm_user_password_change` were deleted or modified out-of-band. Both resources had stub `Read()` bodies — one emitted a warning and preserved state, the other was entirely empty. Neither called the CM API, so drift went undetected indefinitely.

- Schema: 9 new attributes added — see Schema attributes table below.
- `Create()`: after `PostDataBootstrap` returns the resource `id`, now calls `GetByIdBootstrap` to fetch the full server response and unconditionally populate all Computed fields (`id`, `name`, `algorithm`, `fingerprint`, `created_at`, `updated_at`). Previously only `id` was stored in state after create.
- `Read()`: replaces the former stub (which added a "Drift Detection Not Supported" warning and returned without updating state) with:
  - `r.client.GetByIdBootstrap(ctx, id, state.ID.ValueString(), common.URL_SSH_KEY)` — issues `GET api/v1/system/ssh/keys/{id}`.
  - 404 matched via the `notFoundError` package constant: calls `resp.State.RemoveResource(ctx)` and returns.
  - Computed-only fields (`id`, `name`, `algorithm`, `fingerprint`, `created_at`, `updated_at`) are hydrated unconditionally.
  - Optional fields (`key_size`, `curve`, `username`, `public_key_encoding`) are hydrated only when `!state.X.IsNull()` to avoid perpetual drift for users who never configure them.
  - `key` is write-only SSH material; CM never returns it in GET responses. It is preserved from prior state.

- Schema: `id` attribute added (`Computed: true`, `UseStateForUnknown`).
- `Create()`: stores `gjson.Get(response, "user_id").String()` into `plan.ID` after the PATCH response.
- `Read()`: replaces the former empty body with:
  - Guards on `!state.ID.IsNull() && !state.ID.IsUnknown() && state.ID.ValueString() != ""` for backward compatibility with state written before this change.
  - `r.client.GetByIdBootstrap(ctx, id, state.ID.ValueString(), common.URL_USER_MANAGEMENT)` — issues `GET api/v1/usermgmt/users/{user_id}` (swagger-confirmed: `auth-users | GET /v1/usermgmt/users/{user_id}`).
  - 404: calls `resp.State.RemoveResource(ctx)`.
  - Hydrates `state.ID` from `user_id` in the GET response.
  - All credential and input-only fields (`password`, `new_password`, `username`, `auth_domain`, `password_hint`) are write-only; preserved from prior state.

- `CMSSHKeyTFSDK`: 9 new fields added (`Name`, `Algorithm`, `KeySize types.Int64`, `Curve`, `Username`, `PublicKeyEncoding`, `Fingerprint`, `CreatedAt`, `UpdatedAt`) with matching `tfsdk` struct tags.
- `CMPwdChangeTFSDK`: `ID types.String` field added with `tfsdk:"id"`.

Note: `schema_cm.go` was not listed in the original plan's Files to Modify table. It was necessarily modified because the TFSDK struct definitions for both resources live there.

### Modified file — `internal/provider/common/requests.go`

- `GetByIdBootstrap(ctx context.Context, uuid string, id string, endpoint string) (string, error)`: new method on `CMClientBootstrap`. Issues `GET {CipherTrustURL}/{endpoint}/{id}` via `doRequestBootstrap`. Follows the same `MSG_METHOD_START` / `ERR_METHOD_END` / `MSG_METHOD_END` logging pattern as `PostDataBootstrap` and `PatchDataBootstrap`.

Note: this file was not listed in the original plan's Files to Modify table. It was added because `CMClientBootstrap` did not previously expose a GET-by-id method, which is required by both Read() implementations.

### Modified file — `internal/provider/resource_cm_ssh_key_test.go`

- Existing test `Test_CM_CipherTrust_CMSSHKey_noopRead`: extended — the placeholder comment for a `RefreshState` step now has a real step (`RefreshState: true, ExpectNonEmptyPlan: false`).
- `TestCipherTrust_CMSSHKey_OOBDelete` (new): applies the resource, captures the CM id, issues an out-of-band `client.DeleteByURL` in `PreConfig`, then verifies `RefreshState` produces `ExpectNonEmptyPlan: true`. Automatically skips the delete step via `t.Skipf` if CM returns an error, since the SSH key delete API is not formally documented.
- `TestCipherTrust_CMSSHKey_AttributeDrift` (new): apply + `RefreshState` with `ExpectNonEmptyPlan: false`; confirms Read() introduces no false attribute drift.

### Modified file — `internal/provider/resource_cm_user_pwd_change_test.go`

- `TestCipherTrust_CMUserPwdChange_ReadStability` (new): applies the password-change resource using Terraform variables for credentials (no inline secrets), then runs `PlanOnly: true, ExpectNonEmptyPlan: false` confirming Read() does not produce a "Provider produced inconsistent result after apply" error. Gated behind `CM_TEST_PASSWORD_CHANGE=true` to avoid unintended password changes in shared test environments.

### `ciphertrust_cm_ssh_key` — new attributes in this PR

|---|---|---|---|
| `name` | `types.String` | Computed | Server-derived name; hydrated unconditionally in Create() and Read() |
| `algorithm` | `types.String` | Computed | Server-derived algorithm string; hydrated unconditionally |
| `key_size` | `types.Int64` | Optional | Maps to JSON field `size`; Read() guarded on `!state.KeySize.IsNull()` |
| `curve` | `types.String` | Optional | EC curve name; Read() guarded on `!state.Curve.IsNull()` |
| `username` | `types.String` | Optional | Username associated with the key; Read() guarded on `!state.Username.IsNull()` |
| `public_key_encoding` | `types.String` | Optional | Public key encoding format; Read() guarded on `!state.PublicKeyEncoding.IsNull()` |
| `fingerprint` | `types.String` | Computed | `UseStateForUnknown`; set from CM response |
| `created_at` | `types.String` | Computed | `UseStateForUnknown`; stable after creation |
| `updated_at` | `types.String` | Computed | `UseStateForUnknown` applied — see open issue below |

### `ciphertrust_cm_user_password_change` — new attribute in this PR

|---|---|---|---|
| `id` | `types.String` | Computed | `UseStateForUnknown`; stores `user_id` from PATCH response; used in Read() for 404 detection |

## Read() now implemented

Both resources previously returned immediately from `Read()`. This change replaces those stubs with real API calls.

**`ciphertrust_cm_ssh_key`**: calls `c.GetByIdBootstrap(ctx, uuid, state.ID.ValueString(), URL_SSH_KEY)`, hydrates Computed fields unconditionally and Optional fields with `!state.X.IsNull()` guards.

**`ciphertrust_cm_user_password_change`**: this is a one-shot action resource — the password-change event itself has no retrievable state. Read() therefore verifies the *underlying CM user* still exists by calling `GET api/v1/usermgmt/users/{user_id}`. If the user has been deleted out-of-band, 404 removes the resource from state.

**Fields read from CM response — `ciphertrust_cm_ssh_key`** (per `GET api/v1/system/ssh/keys/{id}` response):

- `id` ← JSON `id`
- `name` ← JSON `name`
- `algorithm` ← JSON `algorithm`
- `fingerprint` ← JSON `fingerprint`
- `created_at` ← JSON `createdAt`
- `updated_at` ← JSON `updatedAt`
- `key_size` ← JSON `size` (guarded)
- `curve` ← JSON `curve` (guarded)
- `username` ← JSON `username` (guarded)
- `public_key_encoding` ← JSON `public_key_encoding` (guarded)

**Fields read from CM response — `ciphertrust_cm_user_password_change`** (per `GET api/v1/usermgmt/users/{user_id}`):

- `id` ← JSON `user_id`

**404 handling:** A 404 response calls `resp.State.RemoveResource(ctx)` in both resources, removing them from Terraform state so the next plan proposes recreation. This is intentional and correct for out-of-band deletion detection.

**Null/absent optional fields:** Optional fields guarded by `!state.X.IsNull()` assign the appropriate null constructor (`types.Int64Null()`, `types.StringNull()`) in the `else` branch when CM omits the field.

## Open issues for reviewers

**`GET /v1/system/ssh/keys/{id}` is absent from the swagger spec.** The `cm-system` swagger area only documents `POST /v1/system/ssh/keys` (Add) and `GET/PATCH /v1/system/ssh/kex` (key-exchange config). There is no documented `GET /v1/system/ssh/keys/{id}`. A previous code comment in this resource's Read() stub explicitly stated: "CMClientBootstrap does not expose GetById and the CM SSH key endpoint has no per-resource GET." If that comment was accurate, the new Read() will fail with a non-404 error on every `terraform plan`, which is a regression compared to the previous no-op. Reviewers should confirm against a live CM instance that this endpoint exists and returns the expected JSON fields before merging. The `GET api/v1/usermgmt/users/{user_id}` endpoint used by `ciphertrust_cm_user_password_change` is confirmed in the swagger spec.

**`updated_at` carries `UseStateForUnknown()`.** The codebase convention (documented in the conventions index) prohibits `UseStateForUnknown()` on `updated_at` because it causes plan inconsistency errors when the timestamp changes after an update. However `ciphertrust_cm_ssh_key` is fully immutable — `Update()` always returns an error and `key` uses `ImmutableString()`. Since the key can never be mutated, `updated_at` is stable in practice. Reviewers should decide whether to accept the exception or remove the modifier to follow the project-wide convention.

**`name` and `algorithm` are `Computed: true`.** The diff declares both as `Computed: true` (server-assigned). The original implementation plan specified both as `Required`. The current declaration is correct from a behavioral standpoint — CM derives the algorithm from the supplied key material and assigns the name — but it differs from the plan document. Reviewers should confirm the intended schema contract.

## How to test

  SSH key drift detection and stability:
  ```
  TEST_SSH_PUBLIC_KEY="ssh-rsa AAAA..." \
  TF_ACC=1 go test -v -timeout=15m \
    -run 'TestCipherTrust_CMSSHKey_OOBDelete|TestCipherTrust_CMSSHKey_AttributeDrift' \
    ./internal/provider/
  ```

  Password-change Read stability:
  ```
  CM_TEST_PASSWORD_CHANGE=true \
  TF_VAR_test_current_password="<current-password>" \
  TF_VAR_test_new_password="<new-password>" \
  TF_ACC=1 go test -v -timeout=15m \
    -run 'TestCipherTrust_CMUserPwdChange_ReadStability' \
    ./internal/provider/
  ```

  Scenarios covered:
  - **OOB deletion** (`TestCipherTrust_CMSSHKey_OOBDelete`): create SSH key; delete out-of-band; `RefreshState` must produce a non-empty plan. Auto-skipped if CM rejects the delete.
  - **Read stability** (`TestCipherTrust_CMSSHKey_AttributeDrift`): create SSH key; immediate `RefreshState` must show no changes.
  - **Password change Read stability** (`TestCipherTrust_CMUserPwdChange_ReadStability`): apply password change; plan-only step must show no changes.

- [ ] Manual verification: confirm `GET api/v1/system/ssh/keys/{id}` returns JSON with `id`, `name`, `algorithm`, `fingerprint`, `createdAt`, `updatedAt` fields on the target CM version.

## Checklist

- [ ] Schema attribute `Description` fields populated — `name`, `algorithm`, `key_size`, `curve`, `username`, `public_key_encoding`, `fingerprint`, `created_at`, `updated_at` on `ciphertrust_cm_ssh_key` are missing `Description` values in this diff; required for `make generate` docs.
- [ ] `tflog.Trace` at entry/exit of every CRUD method — present in all modified methods.
- [ ] Fresh `uuid.New().String()` at the top of each CRUD method — present.
- [ ] URL constants used — `URL_SSH_KEY` and `URL_USER_MANAGEMENT` are existing constants in `urls.go`; no inlined string literals introduced.
- [ ] `GET api/v1/system/ssh/keys/{id}` must be confirmed against a live CM instance — not present in swagger operations index.
- [ ] `GET api/v1/usermgmt/users/{user_id}` confirmed in swagger spec.
- [ ] `Read()` 404 removes resource from state — implemented in both resources.
- [ ] Write-only fields are preserved from prior state, never populated from API responses.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-167 (attempt 1)

❌ **Failed** — 5 blocking issue(s).

### 🔴 HIGH — always blocking
- resource_cm_ssh_key.go schema declares `name` as `Computed: true` only (no `Required: true`), but the plan's attribute table specifies `name` as Required. The previous HIGH issue was not fixed — the schema must use `Required: true` for `name`.
- resource_cm_ssh_key.go schema declares `algorithm` as `Computed: true` only (no `Required: true`), but the plan's attribute table specifies `algorithm` as Required. The previous HIGH issue was not fixed — the schema must use `Required: true` for `algorithm`.
- resource_cm_user_pwd_change.go Read() still calls `r.client.GetByIdBootstrap(ctx, id, state.ID.ValueString(), common.URL_USER_MANAGEMENT)` — but the plan specifies `common.URL_USERMGMT_USERS` as the endpoint constant. The previous HIGH issue (wrong endpoint constant) was not fixed; `URL_USER_MANAGEMENT` may resolve to a different path than `URL_USERMGMT_USERS` and break 404 detection.
- resource_cm_ssh_key.go Read() hydrates `state.Name` and `state.Algorithm` unconditionally with `types.StringValue(gjson.Get(...).String())`, treating them as Computed-only, which is internally consistent with the current `Computed: true` schema declaration but both deviate from the plan's `Required` specification. Fixing the schema to `Required: true` is required per the plan.
- `name` and `algorithm` are declared `Computed: true` without `UseStateForUnknown()` plan modifier; per Check 11a, every Computed-only attribute requires `UseStateForUnknown()` to prevent `(known after apply)` on every plan following create. If the intent is to keep them Computed (rather than fixing to Required per plan), both fields need `UseStateForUnknown()` added.

- internal/provider/common/requests.go adds `GetByIdBootstrap` — this file is not listed in the plan's Files to Modify table. The previous MEDIUM issue was not addressed; the plan must include this file modification for traceability.
- internal/provider/cm/schema_cm.go adds new struct fields to `CMSSHKeyTFSDK` and `CMPwdChangeTFSDK` — this file is not listed in the plan's Files to Modify table. The previous MEDIUM issue was not addressed.
- resource_cm_ssh_key.go Read() uses `r.client.GetByIdBootstrap` but the plan's CRUD Wiring specifies `r.client.GetById` (on the standard `CMClient`, not `CMClientBootstrap`). The plan explicitly states the endpoint is `common.URL_SSH_KEYS` (plural), but the implementation also uses `common.URL_SSH_KEY` (singular) — verify these are the same constant, and verify that the bootstrap client is the correct client for this resource's Read() per the plan.

PR #306, opened 2026-07-10
TFIN-352: Fix the acceptance test

ashwani573:
## Summary

- Fixes two failing acceptance tests in `resource_cm_domain_test.go` and `resource_cm_reg_token_test.go`.
- Corrects the `deleteOutOfBand` test for `ciphertrust_domain`: aligns expectations with the actual Read() behaviour (calls `RemoveResource` on 404 rather than keeping the resource in state) and adds a third step to verify Terraform recreates the domain after the out-of-band delete.
- Corrects the `labelDrift` test for `ciphertrust_cm_reg_token`: replaces `ExpectNonEmptyPlan: true` with `ExpectError` because the `label` field carries an `ImmutableMap` plan modifier — after an out-of-band mutation is refreshed into state, the modifier fires a plan-time error rather than producing a silent non-empty diff.
- No production code was modified; all changes are in test files only.

## Motivation

Implements TFIN-352: Fix the acceptance tests.

Two acceptance tests encoded incorrect assumptions about provider behaviour and were failing against a live CipherTrust Manager instance. This change corrects both tests to match what the production code actually does.

The `Test_CM_AccCipherTrustCMDomain_deleteOutOfBand` test previously asserted `ExpectNonEmptyPlan: false` after an out-of-band delete followed by `RefreshState: true`, with a comment claiming Read() keeps the resource in state on 404. The actual `Read()` implementation calls `resp.State.RemoveResource(ctx)` when CM returns 404, so the resource is removed from state. With the resource gone from state but still present in config, Terraform plans a recreation — meaning the plan is non-empty.

Changes made:
- Updated the Step 2 comment to reflect the real behaviour: 404 → `RemoveResource` → resource absent from state → plan proposes recreation.
- Changed `ExpectNonEmptyPlan: false` to `ExpectNonEmptyPlan: true`.
- Added Step 3: re-applies the original config so Terraform recreates the domain, confirming the full delete-and-recreate lifecycle completes without error.

The `Test_CM_CipherTrust_CMRegToken_labelDrift` test (the step that mutates the `label` field out-of-band) previously used:
```
RefreshState:       true,
ExpectNonEmptyPlan: true,
```

The `label` attribute on `ciphertrust_cm_reg_token` is declared immutable via the `ImmutableMap` plan modifier. When `RefreshState` runs, Read() hydrates the server-mutated value (`"updated"`) into state. On the subsequent implicit plan, the framework detects that config (`"default"`) differs from state (`"updated"`) and the `ImmutableMap` modifier fires a diagnostic error — it does not produce a quiet non-empty plan. The fix replaces `ExpectNonEmptyPlan: true` with:
```go
ExpectError: regexp.MustCompile(`(?i)immutable|cannot be changed`),
```
A clarifying comment was also added explaining the sequence of events (OOB mutation → refresh → ImmutableMap error).

## Read() behaviour being tested

**`ciphertrust_domain`**: Read() calls `resp.State.RemoveResource(ctx)` when CM returns a 404. The corrected test now verifies this path: after an out-of-band delete, a refresh removes the resource from state and Terraform plans to recreate it.

**`ciphertrust_cm_reg_token`**: Read() unconditionally hydrates the `label` map from the CM API response. The corrected test now verifies that a server-side mutation of an immutable field, once refreshed into state, causes the `ImmutableMap` plan modifier to surface a diagnostic error rather than silently queuing an update.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_AccCipherTrustCMDomain_deleteOutOfBand' -v -timeout 15m
  go test ./internal/provider/ -run 'Test_CM_CipherTrust_CMRegToken_labelDrift' -v -timeout 15m
  ```
  Scenarios covered:
  - **`deleteOutOfBand` (domain)** — create domain; delete it directly via CM API; refresh state; assert plan is non-empty (recreation proposed); re-apply; assert domain is recreated with correct name.
  - **`labelDrift` (reg token)** — create token with `label`; mutate `label` directly via CM API; refresh state; assert `ImmutableMap` diagnostic error fires (no silent update).

## Checklist

- [ ] No production schema or CRUD code was modified — this is a test-only change.
- [ ] `ExpectError` regexp uses `(?i)` for case-insensitive matching, consistent with the project convention for `ImmutableMap` error assertions.
- [ ] No secrets or credentials interpolated into HCL config strings.
- [ ] Test functions retain their existing names — no renames that would break CI test selectors.

---

### DIFF
--- a/internal/provider/resource_cm_domain_test.go
+++ b/internal/provider/resource_cm_domain_test.go
@@ -161,6 +161,18 @@ resource "ciphertrust_domain" "test" {
 `, name, adminsStr, allowUserMgmt, metaStr)
 }
 
+// cmDomainOOBConfig returns an HCL config for a domain resource used in
+// out-of-band delete testing. Uses the resource label "oob" to avoid
+// conflicts with other tests using "test" label.
+func cmDomainOOBConfig(name string) string {
+	return fmt.Sprintf(providerConfig+`
+resource "ciphertrust_domain" "oob" {
+  name   = %q
+  admins = ["admin"]
+}
+`, name)
+}
+
 // Test_CM_AccCipherTrustCMDomain_basicDrift verifies that out-of-band mutations to
 // admins, allow_user_management, and meta_data surface as drift.
 func Test_CM_AccCipherTrustCMDomain_basicDrift(t *testing.T) {
@@ -227,8 +239,8 @@ func Test_CM_AccCipherTrustCMDomain_basicDrift(t *testing.T) {
 }
 
 // Test_CM_AccCipherTrustCMDomain_deleteOutOfBand verifies that when a domain is
-// deleted out-of-band, Read() emits a warning and keeps the resource in state,
-// and that terraform destroy on an already-deleted domain completes without error.
+// deleted out-of-band, Read() calls RemoveResource to remove the resource from state,
+// and the plan correctly proposes recreation. After re-apply, the domain is recreated.
 func Test_CM_AccCipherTrustCMDomain_deleteOutOfBand(t *testing.T) {
 	RequireCM(t)
 	requireDomainCreationLicensed(t)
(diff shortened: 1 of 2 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-352

PR #312, opened 2026-07-13
TFIN-369: Validate and fix drift detection and immutable field behavior for ciphertrust_cm_prometheus

ashwani573:
## Summary

- Fixes double-slash typos in `URL_PROMETHEUS_ENABLE` and `URL_PROMETHEUS_DISABLE` in `internal/provider/common/urls.go` — the malformed paths (`api//v1/...`) caused HTTP failures on every `Create()`, `Update()`, and `Delete()` call for `ciphertrust_cm_prometheus`.
- Fixes incorrect JSON struct tag on `CMPrometheusMetricsConfigJSON.Token` in `internal/provider/cm/schema_cm.go`: was `json:"id"`, corrected to `json:"token"` to match the CM API `metrics_config` schema (`token` string field confirmed in swagger `Definition: metrics_config`).
- Extends `internal/provider/resource_cm_prometheus_test.go` with two acceptance test functions covering no-drift stability and out-of-band drift-detection scenarios for the existing `Read()` implementation.

## Motivation

Implements TFIN-369: Validate and fix drift detection and immutable field behavior for `ciphertrust_cm_prometheus`.

### Modified file — `internal/provider/common/urls.go`

- `URL_PROMETHEUS_ENABLE` corrected from `"api//v1/system/metrics/prometheus/enable"` to `"api/v1/system/metrics/prometheus/enable"`.
- `URL_PROMETHEUS_DISABLE` corrected from `"api//v1/system/metrics/prometheus/disable"` to `"api/v1/system/metrics/prometheus/disable"`.

Both endpoints are confirmed against the swagger spec (area: `cm-system`):

- `POST /v1/system/metrics/prometheus/enable` — confirmed
- `POST /v1/system/metrics/prometheus/disable` — confirmed

The double-slash produced a URL of the form `<base>/api//v1/...` which fails with an HTTP 404 or path-resolution error on every `Create()`, `Update()`, and `Delete()` invocation. `URL_PROMETHEUS_STATUS` (`"api/v1/system/metrics/prometheus/status"`) used by `Read()` was already correct and is unchanged.

- Fixes the JSON struct tag on `CMPrometheusMetricsConfigJSON.Token`: was `json:"id"`, now `json:"token"`.

The CM API `metrics_config` swagger definition lists properties `enabled (boolean)` and `token (string)`. The previous `json:"id"` tag meant that any `json.Marshal` of this struct would emit the field as `{"id": "..."}` rather than `{"token": "..."}`. Response parsing in `Create()` and `Read()` uses `gjson.Get(response, "token")` directly and was unaffected at runtime, but the struct definition was factually wrong and would produce an incorrect payload if `Token` were ever non-empty when marshaled into a request body.

### Modified file — `internal/provider/resource_cm_prometheus_test.go`

- Adds imports: `context`, `encoding/json`, `common` (used by the two new test functions).
- Adds `prometheusEnabledConfig` shared HCL constant to avoid repeating the config literal in both new test functions.
- Adds `Test_CM_AccCMPrometheus_NoDrift`: two-step test — apply `enabled = true`, then `PlanOnly: true, ExpectNonEmptyPlan: false` — asserts that `Read()` calling `GET /v1/system/metrics/prometheus/status` introduces no spurious drift when CM state matches Terraform state.
- Adds `Test_CM_AccCMPrometheus_DriftDetection`: two-step test — apply `enabled = true`, then a `PreConfig` that calls `client.PostDataV2(ctx, "", common.URL_PROMETHEUS_DISABLE, payload)` out-of-band (simulating an operator disabling Prometheus directly in CM), followed by `RefreshState: true, ExpectNonEmptyPlan: true` — asserts that `Read()` correctly surfaces the external change as a non-empty plan.

Both functions call `RequireCM(t)` as their first statement and use `testAccProtoV6ProviderFactories`.

## Read() now implemented

`Read()` was already implemented in `resource_cm_prometheus.go` before this PR. It calls `c.ReadDataByParam(ctx, id, "all", common.URL_PROMETHEUS_STATUS)` against `GET /v1/system/metrics/prometheus/status` (confirmed in swagger) and hydrates state as follows:

- `enabled` — `types.BoolValue(gjson.Get(response, "enabled").Bool())` — unconditional
- `token` — three-branch resolution: use the API value when non-empty; preserve prior state token when the API returns empty (Prometheus disabled, CM omits the token from the response); fall back to `types.StringValue("")` on first apply or import

**404 handling:** A 404 response calls `resp.State.RemoveResource(ctx)` — Prometheus is a singleton system-level resource; if CM reports it as absent the Terraform resource should be removed from state so a subsequent `apply` can re-create it. This differs from standard CRUD resources where 404 is treated as a non-fatal stale reference.

The URL constant bug and struct tag bug in this PR were masking the `ciphertrust_cm_prometheus` resource's broken CRUD path. Fixing them unblocks drift detection (which was functionally correct in `Read()`) and the two new tests prove it end-to-end.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_AccCMPrometheus' -v -timeout=15m
  ```
  Scenarios covered:
  - **No-drift stability** (`Test_CM_AccCMPrometheus_NoDrift`) — apply config, assert second plan shows no changes.
  - **Drift detection** (`Test_CM_AccCMPrometheus_DriftDetection`) — apply, disable Prometheus out-of-band via CM API in `PreConfig`, refresh state, assert non-empty plan.
- [ ] Manual smoke-test: `terraform apply` (`enabled = true`), verify `token` attribute is stored in state, change to `enabled = false`, re-apply, verify token is preserved across the disable, `terraform destroy`.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_cm_prometheus.go -> <Func>][id]` pattern — verify present in existing file.
- [ ] URL constants defined in `urls.go` — no inlined string literals in resource file.
- [ ] CM API endpoints verified against swagger spec — `POST /v1/system/metrics/prometheus/enable`, `POST /v1/system/metrics/prometheus/disable`, and `GET /v1/system/metrics/prometheus/status` all confirmed in `cm-system` area.
- [ ] `Read()` 404 removes resource from state — appropriate for this singleton resource.
- [ ] New test functions call `RequireCM(t)` as first statement.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-369

PR #316, opened 2026-07-13
TFIN-376: Summary: Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_interface and ciphertrust_license

ashwani573:
## Summary

- Fixes perpetual-drift bug in `ciphertrust_license` `Read()`: `bind_type` hydration is now gated on `!state.BindType.IsNull()` so that CM's non-empty default value does not create a state/config mismatch when the user omits `bind_type` from their configuration.
- Replaces inlined `"status: 404"` string literal with the package-level `notFoundError` constant in `ciphertrust_license` `Read()` and `Delete()`.
- Makes `ciphertrust_license` `Delete()` idempotent: a 404 response from CM is now treated as success, so destroying a license that was already removed out-of-band no longer errors.
- Adds acceptance tests for `ciphertrust_license` (immutability of `license` and `bind_type`, idempotency across refreshes) and extends the `ciphertrust_interface` test file with an idempotency scenario.
- Updates `docs/resources/interface.md` with a `## Behavior` subsection covering immutable fields, out-of-band deletion, and out-of-band attribute drift.
- Updates `docs/resources/license.md` schema descriptions to document immutability constraints, attribute semantics, and the `bind_type` no-drift guarantee.

## Motivation

Implements TFIN-376: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_interface` and `ciphertrust_license`.

**Fix 1 — `Read()`: use `notFoundError` package constant**

The 404 detection string in `Read()` was previously inlined as the raw literal `"status: 404"`. It is now replaced with the package-level constant `notFoundError` (defined in `internal/provider/cm/constants.go` as `"status: 404"`). This matches the convention used elsewhere in the `cm` package and avoids the risk of a silent mismatch if the constant value ever changes.

**Fix 2 — `Read()`: gate `bind_type` hydration on prior-state null check**

Before this change, `bind_type` was hydrated unconditionally from the CM API response. CipherTrust Manager always returns a `bind_type` field (e.g. `"instance"`) in the `GET /v1/licensing/licenses/{id}` response even when the user did not configure the field. This caused perpetual drift: state held `"instance"` while the configuration held `null`, so every `terraform plan` reported a non-empty diff with no way for the user to suppress it.

The fix adds an `!state.BindType.IsNull()` outer guard:

```go
if !state.BindType.IsNull() {
    if r := gjson.Get(response, "bind_type"); r.Exists() && r.String() != "" {
        state.BindType = types.StringValue(r.String())
    } else {
        state.BindType = types.StringNull()
    }
}
```

When the user has not set `bind_type` in their configuration, `state.BindType` is `null` after `Create()`. The guard preserves `null` on every subsequent `Read()` and no drift is reported. When the user has explicitly set `bind_type`, state is non-null after `Create()`, the guard passes, and `Read()` correctly refreshes the value from CM.

The swagger `Licenses` definition (`GET /v1/licensing/licenses/{id}`) confirms that `bind_type` is present in the response schema but is not Required in `PostLicense` — it is an optional field that CM auto-populates when omitted.

**Fix 3 — `Delete()`: idempotent destroy on 404**

Before this change, if a license was deleted directly in CipherTrust Manager (outside Terraform) and then `terraform destroy` was run, `Delete()` would receive a 404 from `DeleteByID` and surface it as an error, blocking the destroy operation and leaving Terraform state inconsistent.

The fix adds an early-return guard using `notFoundError`:

```go
    return
}
```

A 404 on delete means the resource is already gone from CM; Terraform can safely remove it from state. The `DeleteByID` call still passes `state.ID.ValueString()` as the path key, consistent with the `DELETE /v1/licensing/licenses/{id}` swagger endpoint.

### Modified file — `internal/provider/resource_interface_test.go`

Adds `Test_CM_Interface_Idempotency`: creates a KMIP interface on port 9100, then runs a second `PlanOnly` step with an identical configuration and asserts `ExpectNonEmptyPlan: false`. This verifies that no schema attribute causes a spurious diff after the initial apply. The test uses the existing `interfaceSweep(9100)` helper as a `PreConfig` to remove any leftover interface at that port before the create step. The test calls `RequireCM(t)` as its first statement.

### New file — `internal/provider/resource_cm_license_test.go`

Three acceptance tests for `ciphertrust_license`:

- **`Test_CM_License_ImmutableLicense`** — Creates a license, then runs a `PlanOnly` step that changes the `license` field to a different value. Asserts that `modifiers.ImmutableString()` fires and the plan-time diagnostic matches `(?i)Attribute is immutable`. Skips when `TF_ACC_LICENSE_CODE` is unset.
- **`Test_CM_License_ImmutableBindType`** — Same pattern for the `bind_type` field: creates with `bind_type = "instance"`, then attempts to change to `"cluster"`. Asserts the immutability diagnostic. Skips when `TF_ACC_LICENSE_CODE` is unset.
- **`Test_CM_License_Idempotency`** — Creates a license and verifies all Computed attributes (`id`, `hash`, `type`, `state`, `expiration`, `version`, `license_count`) are set. A second `PlanOnly` step with identical config asserts `ExpectNonEmptyPlan: false`, confirming that `Read()` drift fixes and `UseStateForUnknown()` plan modifiers collectively prevent spurious diffs.

The license code is passed via `TF_VAR_cm_license_code` (a Terraform variable), never interpolated directly into HCL config strings, preventing secret exposure in test logs. All three tests call `RequireCM(t)` as their first statement.

### Modified file — `docs/resources/interface.md`

Added a `## Behavior` subsection immediately before the auto-generated `## Schema` section. The subsection documents three behavioural properties:

1. `name` and `interface_type` are immutable after creation — a plan-time error is emitted if either is changed.
2. Out-of-band deletion — CM API 404 causes `Read()` to remove the resource from state, scheduling recreation on the next apply.
3. Out-of-band attribute changes — CM API returns the new value on refresh, surfacing it as a non-empty plan diff.

**Note for reviewers:** The `## Behavior` section is hand-authored prose and cannot be auto-generated by `tfplugindocs`. It must be preserved in any future `make generate` run. Consider adding it to the template or a `website/` override so it survives regeneration.

### Modified file — `docs/resources/license.md`

Updated descriptions for all schema attributes to document:

- `license` — immutability constraint, destroy-and-recreate requirement.
- `bind_type` — immutability constraint, CM fallback behaviour when omitted, and the explicit no-drift guarantee (Terraform will not report drift even though CM echoes a default value).
- All Computed attributes (`expiration`, `hash`, `license_count`, `start`, `state`, `type`, `version`) — added human-readable descriptions.
- `state` — noted that the value changes over the license lifecycle and is always refreshed from the API.
- `trial_seconds_remaining` — noted that it returns an empty string for non-trial licenses, which is stable and does not cause drift.

**Note for reviewers:** The `docs/resources/license.md` changes update descriptions for schema attributes. These descriptions are auto-generated from the `Description:` fields in the Go schema definition. The corresponding Go `Description:` fields in `resource_license.go` were **not** updated in this PR, which means the next `make generate` run will revert the license.md schema descriptions to their previous values. Reviewers should decide whether to accept this as-is (documentation only, overwritten on next regeneration) or require that the Go schema `Description:` fields be updated to match, making the docs durable.

## How to test

- [ ] Acceptance tests for `ciphertrust_license` (requires live CM, a valid license code, and `TF_ACC=1`):

  Set `TF_ACC_LICENSE_CODE` to a valid license string and run:
  ```
  TF_ACC=1 TF_ACC_LICENSE_CODE=<your-license> go test -v -timeout=15m -run 'Test_CM_License_' ./internal/provider/
  ```
  Scenarios covered:
  - **Immutable `license` field** — plan-time error prevents any attempt to change the license value after creation.
  - **Immutable `bind_type` field** — plan-time error prevents any attempt to change the binding type after creation.
  - **Idempotency** — second plan with identical config produces no diff; all Computed attributes are stable across refreshes.

- [ ] Acceptance test for `ciphertrust_interface` idempotency (requires live CM and `TF_ACC=1`):
  ```
  TF_ACC=1 go test -v -timeout=15m -run 'Test_CM_Interface_Idempotency' ./internal/provider/
  ```
  Scenario covered:
  - **Idempotency** — KMIP interface on port 9100, second plan with identical config produces no diff.

## Checklist

- [ ] `bind_type` guard follows the "server-auto-populated Optional+Computed" hydration pattern: only read from CM when state is non-null; outer `!state.BindType.IsNull()` guard confirmed correct.
- [ ] `notFoundError` constant (`"status: 404"`) is defined in `internal/provider/cm/constants.go` — not redeclared in `resource_license.go`.
- [ ] Delete() 404 guard is placed before `resp.Diagnostics.AddError` so an already-deleted license does not surface as a destroy failure.
- [ ] `Test_CM_License_*` tests skip gracefully when `TF_ACC_LICENSE_CODE` is unset — no hard-coded license strings in source.
- [ ] `Test_CM_License_*` pass the license value via `TF_VAR_cm_license_code` (Terraform variable reference), never via `fmt.Sprintf` interpolation into HCL.
- [ ] All new test functions call `RequireCM(t)` as their first statement.
- [ ] `ExpectError` regexp `(?i)Attribute is immutable` matches the exact title emitted by `modifiers.ImmutableString()` (`"Attribute is immutable"`) — confirmed against `internal/provider/modifiers/immutable.go`.
- [ ] `docs/resources/interface.md` `## Behavior` section is hand-authored and will not survive `make generate` without a template override — reviewers to decide whether a template or website override is required.
- [ ] `docs/resources/license.md` schema descriptions are hand-edited; the corresponding Go `Description:` fields in `resource_license.go` were **not** updated — the next `make generate` will revert license.md. Reviewers to decide whether to require Go schema updates for durability.
- [ ] No `docs/` files other than interface.md and license.md were modified by hand.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-376

rohitpandita-hub:
@terraform-agent Test_CM_Interface_Idempotency is failing, fix it

ashwani573:
Addressed in `c791306`: TFIN-376: address review feedback (round 1)

PR #318, opened 2026-07-13
TFIN-374: Validate drift detection, immutable fields, and idempotency for ciphertrust_domain, ciphertrust_groups, and ciphertrust_cm_groups_list

ashwani573:
## Summary

- Fixes `ciphertrust_domain` `Read()` to call `resp.State.RemoveResource(ctx)` on HTTP 404 instead of adding a warning and retaining the resource in state.
- Fixes `ciphertrust_domain` `Read()` and post-PATCH `Update()` `meta_data` hydration: replaces a two-branch `absent-or-empty → MapNull` / `non-empty → MapValueFrom` pattern with a three-branch pattern guarded by `!state.Meta.IsNull()`, eliminating phantom `null → {}` drift for users who never configure `meta_data`.
- Fixes `ciphertrust_groups` `Read()` to remove a redundant `resp.Diagnostics.AddWarning` call that preceded the already-present `resp.State.RemoveResource(ctx)` on HTTP 404.
- Extends `resource_cm_domain_test.go` with 3 acceptance tests and `resource_cm_group_test.go` with 6 acceptance tests covering drift detection, out-of-band deletion, immutable-name enforcement, idempotency, and `ciphertrust_cm_groups_list` data-source accuracy.

## Motivation

Implements TFIN-374: Validate drift detection, immutable fields, and idempotency for `ciphertrust_domain`, `ciphertrust_groups`, and `ciphertrust_cm_groups_list`.

**404 handling in `Read()`** (behavioral fix):

Previously the 404 branch called `resp.Diagnostics.AddWarning(...)` and returned without removing the resource from state. This caused Terraform to silently keep a stale domain in state after out-of-band deletion, and subsequent plans would show no changes even though the real domain no longer existed.

The fix: on `notFoundError` (HTTP 404), emit `tflog.Warn` and immediately call `resp.State.RemoveResource(ctx)` so Terraform schedules recreation on the next plan. This aligns `ciphertrust_domain` with the provider-wide standard used by all other CM resources.

> **Behavioral note for existing users**: After this change, if a domain is deleted directly in CipherTrust Manager, the next `terraform refresh` or `terraform plan` will remove it from state and show a plan to recreate it. Previously, the domain remained in state indefinitely. This is the correct Terraform idiom for out-of-band deletion.

**`meta_data` hydration in `Read()` and post-PATCH `Update()`** (drift fix):

The previous two-branch logic collapsed both the `absent` and the `{}` cases into `types.MapNull(...)`:

```
absent or {} → MapNull
non-empty    → MapValueFrom(...)
```

CipherTrust Manager returns `{}` for domains where the user never set `meta_data`. With the old logic, a user who never configures `meta_data` starts with `state.Meta = types.MapNull(...)`. On the next `Read()`, CM returns `{}`, which also maps to `MapNull` — so state stays null. That part was fine.

However, the collapse was achieved by always converting `{}` to `MapNull`, which broke the case where a user explicitly sets `meta_data = {}`. In that case state is not null (it is an empty map), CM echoes `{}`, but the old code would overwrite it with `MapNull`, causing perpetual drift.

The replacement three-branch logic:

```
guard: !state.Meta.IsNull()   (skip entirely when state is null — CM's {} is irrelevant)
  absent    → types.MapNull(types.StringType)
  {} (empty) → types.MapValueMust(types.StringType, map[string]attr.Value{})
  non-empty  → types.MapValueFrom(ctx, types.StringType, metaMap)
```

The outer `!state.Meta.IsNull()` guard preserves the null for users who never configure `meta_data` (avoiding null→{} phantom drift), while the inner three branches correctly handle absent, empty, and non-empty map responses when the user has configured `meta_data`.

The same three-branch logic was applied to the post-PATCH `meta_data` read-back in `Update()`, using `state.Meta` (prior state, loaded from `req.State`) as the guard value.

**New import**: `"github.com/hashicorp/terraform-plugin-framework/attr"` — required by `types.MapValueMust(types.StringType, map[string]attr.Value{})` for the empty-map branch.

Removed the `resp.Diagnostics.AddWarning("CipherTrust Group Not Found", ...)` call that appeared immediately before `resp.State.RemoveResource(ctx)` in the `Read()` 404 branch. The warning was redundant (the `tflog.Warn` above it already records the event) and contrary to the provider convention — `RemoveResource` is the correct and complete response to a 404; adding a warning alongside it surfaces confusing output to the user.

Three new acceptance test functions extending the existing test suite:

- `Test_CM_AccCMDomain_Drift` — creates a domain with `meta_data = {"env": "test"}`, then mutates `meta_data` out-of-band via the CM API (`UpdateData`), then calls `RefreshState` and asserts `ExpectNonEmptyPlan: true`. Validates that the `!state.Meta.IsNull()` guard allows drift to surface when `meta_data` was configured.
- `Test_CM_AccCMDomain_DeleteOutOfBand` — creates a domain, deletes it directly via the CM API, then calls `RefreshState` and asserts `ExpectNonEmptyPlan: true`. Validates the fixed 404 branch (`RemoveResource` + plan recreation).
- `Test_CM_AccCMDomain_Idempotency` — creates a domain with no `meta_data` block, then runs a second plan with an identical config and asserts `ExpectNonEmptyPlan: false`. Validates that the `!state.Meta.IsNull()` guard prevents phantom `null → {}` drift when CM returns `{}` for the `meta` field.

Each test calls `RequireCM(t)` and `requireDomainCreationLicensed(t)` as first steps, and uses `domainSweep()` in `PreConfig` to avoid hitting domain count license limits from prior failed runs.

Six new acceptance test functions extending the existing test suite:

- `Test_CM_AccCMGroup_Drift` — creates a group, mutates `description` out-of-band, then asserts drift via `RefreshState` + `ExpectNonEmptyPlan: true`. Validates `Read()` correctly surfaces description changes.
- `Test_CM_AccCMGroup_DeleteOutOfBand` — creates a group, deletes it directly via the CM API, then asserts via `RefreshState` + `ExpectNonEmptyPlan: true` that `RemoveResource` fires on 404 (no stale warning, immediate recreation plan).
- `Test_CM_AccCMGroup_ImmutableName` — creates a group, then attempts a plan that changes `name`; asserts `ExpectError: regexp.MustCompile(`(?i)immutable`)`. Validates that `modifiers.ImmutableString()` blocks renames at plan time.
- `Test_CM_AccCMGroup_Idempotency` — creates a group with a stable description, then re-applies identical HCL (`PlanOnly: true`) and asserts `ExpectNonEmptyPlan: false`. Validates no phantom drift on consecutive plans.
- `Test_CM_AccCMGroupsList_ReadAccuracy` — creates a group via `ciphertrust_groups`, then reads the `ciphertrust_cm_groups_list` data source and asserts `groups.#` is set (at least one entry). Validates the data source returns the created group.
- `Test_CM_AccCMGroupsList_Idempotency` — performs two consecutive reads of the `ciphertrust_cm_groups_list` data source and asserts the second plan is empty (`ExpectNonEmptyPlan: false`). Validates consecutive data-source reads produce no diff.

## Read() fixes

`Read()` was already implemented in both resources before this PR. The changes fix the following defects:

**`ciphertrust_domain` Read():**

- 404 handling: was `AddWarning + return` (resource kept in state); now `tflog.Warn + RemoveResource + return` (resource removed from state, triggering recreation plan).
- `meta_data` hydration: was `absent-or-{}  → MapNull` / `non-empty → MapValueFrom` (caused phantom drift when empty map was explicitly configured); now `absent → MapNull` / `{} → MapValueMust({})` / `non-empty → MapValueFrom`, guarded by `!state.Meta.IsNull()` to suppress irrelevant CM `{}` for unconfigured domains.

**`ciphertrust_groups` Read():**

- 404 handling: removed redundant `AddWarning` that preceded the already-present `RemoveResource` call.

**Fields read from CM response by `ciphertrust_domain` Read()** (unchanged — listed for reviewer reference):

- `state.ID` ← `id`
- `state.Name` ← `name`
- `state.URI` ← `uri`
- `state.Account` ← `account`
- `state.DevAccount` ← `devAccount`
- `state.Application` ← `application`
- `state.CreatedAt` ← `createdAt`
- `state.UpdatedAt` ← `updatedAt`
- `state.HSMConnectionId` ← `hsm_connection_id` (null when absent/empty)
- `state.HSMKEKLabel` ← `hsm_kek_label` (null when absent/empty)
- `state.ParentCAId` ← `parent_ca_id` (null when absent/empty)
- `state.AllowUserManagement` ← `allow_user_management` (null when absent)
- `state.Admins` ← `admins` (prior state preserved when absent — required field)
- `state.Meta` ← `meta` (three-branch, guarded — see above)

**404 handling**: A 404 response now calls `resp.State.RemoveResource(ctx)` for both resources. This is the correct Terraform idiom — the resource was deleted outside of Terraform and the next plan will propose recreation if the config still references it.

## How to test

  ```
  go test ./internal/provider/... -run 'Test_CM_AccCMDomain_Drift|Test_CM_AccCMDomain_DeleteOutOfBand|Test_CM_AccCMDomain_Idempotency' -v -timeout 15m
  go test ./internal/provider/... -run 'Test_CM_AccCMGroup_Drift|Test_CM_AccCMGroup_DeleteOutOfBand|Test_CM_AccCMGroup_ImmutableName|Test_CM_AccCMGroup_Idempotency|Test_CM_AccCMGroupsList_ReadAccuracy|Test_CM_AccCMGroupsList_Idempotency' -v -timeout 15m
  ```
  Scenarios covered:
  - **Drift detection** — OOB mutation surfaced via `RefreshState` + `ExpectNonEmptyPlan: true` for both `ciphertrust_domain` (`meta_data`) and `ciphertrust_groups` (`description`).
  - **OOB deletion** — resource removed from state (no stale warning) and plan proposes recreation for both resource types.
  - **Immutable name** — `Test_CM_AccCMGroup_ImmutableName` asserts a rename attempt is blocked at plan time with an `(?i)immutable` error.
  - **Idempotency** — consecutive identical plans produce empty diffs for `ciphertrust_domain` (no phantom `meta_data` drift when absent), `ciphertrust_groups`, and `ciphertrust_cm_groups_list`.
  - **Data source accuracy** — `Test_CM_AccCMGroupsList_ReadAccuracy` asserts the groups data source returns at least one entry after a group is created.

## Checklist

- [ ] URL constants defined in `urls.go` — no inlined string literals (`URL_DOMAIN`, `URL_GROUP` used throughout).
- [ ] CM API endpoints verified against swagger spec (`api/v1/domains`, `api/v1/usermgmt/groups`).
- [ ] All new test functions use `Test_CM_` prefix (e.g., `Test_CM_AccCMDomain_Drift`).
- [ ] `Read()` 404 now calls `resp.State.RemoveResource(ctx)` — no `resp.State.RemoveResource` omission; no stale `AddWarning` left alongside it.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-374

PR #322, opened 2026-07-14
TFIN-381: Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_log_forwarder and ciphertrust_ntp

ashwani573:
## Summary

- Removes `stringplanmodifier.UseStateForUnknown()` from the `updated_at` attribute on `ciphertrust_log_forwarder`, fixing a plan inconsistency that occurred after every successful update.
- Adds two acceptance tests to `resource_log_forwarder_test.go` covering drift detection (out-of-band attribute change) and out-of-band deletion (HTTP 404 → `RemoveResource` → recreation plan).
- Adds a "## Behavior Notes" section to `docs/resources/log_forwarder.md` documenting immutability of `type`, drift detection behavior, and `updated_at` refresh semantics.
- Adds a "## Behavior Notes" section to `docs/resources/ntp.md` documenting immutability of `host`, replacement behavior for `key`/`key_type`, and idempotency guarantees.

## Motivation

Implements TFIN-381: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_log_forwarder` and `ciphertrust_ntp`.

The `updated_at` schema attribute previously carried `stringplanmodifier.UseStateForUnknown()`:

```go
// Before — causes plan inconsistency after every update
"updated_at": schema.StringAttribute{
    Computed: true,
    PlanModifiers: []planmodifier.String{
        stringplanmodifier.UseStateForUnknown(),
    },
},
```

`UseStateForUnknown()` instructs the Terraform framework to substitute the prior state value for the attribute when the planned value is unknown. For a last-modified timestamp such as `updated_at`, this means the framework expects to see the old timestamp after apply, but `Read()` returns the new timestamp from CM — producing a framework error: `"was T1, but now T2"`.

The fix removes the modifier so that `updated_at` is re-read from the CM API on every refresh without carrying the old value forward:

```go
// After — correct for a last-modified timestamp
"updated_at": schema.StringAttribute{
    Computed: true,
},
```

No other schema attributes were changed. All other `Computed` fields (`id`, `account`, `created_at`) that are stable after creation correctly retain `UseStateForUnknown()`.

### Modified file — `internal/provider/resource_log_forwarder_test.go`

Two new acceptance tests added:

**`Test_CM_LogForwarder_Drift`** — verifies that `Read()` surfaces an out-of-band attribute change as drift:
- Step 1: applies a `ciphertrust_log_forwarder` resource and captures its ID.
- Step 2: uses a `PreConfig` to rename the log forwarder directly via `client.UpdateDataV2` against `URL_CM_LOG_FORWARDS/{id}`, then asserts `RefreshState: true, ExpectNonEmptyPlan: true` — confirming that the next `terraform plan` detects the drift and proposes a corrective update.

**`Test_CM_LogForwarder_OOBDeletion`** — verifies that `Read()` removes the resource from state on HTTP 404 and Terraform plans recreation:
- Step 1: applies the resource and captures its ID.
- Step 2: uses a `PreConfig` to delete the resource directly via `client.DeleteByURL` against `URL_CM_LOG_FORWARDS/{id}`, then asserts `RefreshState: true, ExpectNonEmptyPlan: true` — confirming that the provider removes the resource from state on 404 and Terraform plans to recreate it.

Both tests call `RequireCM(t)` and `requireLogForwarderConnID(t)` (which skips when `CIPHERTRUST_LOG_FORWARDER_CONNECTION_ID` is unset). Resource names use `acctest.RandStringFromCharSet` to avoid collisions across parallel runs.

### Modified file — `docs/resources/log_forwarder.md`

Adds a "## Behavior Notes" section with three subsections:

- **Immutable Fields**: documents that `type` is immutable after creation, enforced by `modifiers.ImmutableString()`. Changing it produces a plan-time error with no API call and no destroy/recreate.
- **Drift Detection**: documents 404 behavior (provider removes the resource from state; Terraform plans recreation) and mutable-attribute drift behavior (next `terraform plan` detects the change and proposes a corrective update).
- **`updated_at` Attribute**: clarifies that `updated_at` refreshes on every `terraform plan`/`terraform refresh` and changes after every successful update — this is expected behavior, not a perpetual diff.

Note: this section is intentional editorial content added manually. It describes behavior that is not auto-generated by `tfplugindocs` from schema definitions. The rest of the file remains generated output.

### Modified file — `docs/resources/ntp.md`

Adds a "## Behavior Notes" section with three subsections:

- **Immutable Fields**: documents that `host` is immutable after creation, enforced by `modifiers.ImmutableString()`. Changing it produces a plan-time error.
- **Fields That Trigger Replacement**: documents that `key` and `key_type` use `stringplanmodifier.RequiresReplaceIfConfigured()` — if either is set or changed after initial creation, Terraform will plan a destroy-and-recreate of the NTP resource.
- **Drift Detection**: clarifies that because `host` is immutable and is the sole identifier, out-of-band attribute mutations are not tracked for this resource. A subsequent `terraform plan` after an unmodified `terraform apply` produces an empty plan.

Same editorial note applies: this section is manually added and not regenerated by `tfplugindocs`.

## Read() behavior (existing implementation validated by new tests)

`Read()` for `ciphertrust_log_forwarder` was already implemented prior to this PR. The new tests exercise two existing behaviors:

- **Drift detection**: `Read()` fetches the current state from CM and overwrites all mutable attributes in Terraform state. When an attribute has been changed out-of-band, the refreshed state differs from config, producing a non-empty plan.
- **404 handling**: When CM returns HTTP 404, `Read()` calls `resp.State.RemoveResource(ctx)`, removing the resource from Terraform state. The next `terraform plan` then proposes recreation.

## Breaking changes

None introduced by this PR. The `type` immutability (`modifiers.ImmutableString()`) and the `host` immutability for NTP were already present in the schema before this change. The `updated_at` modifier removal does not affect the external API surface — it corrects a bug that prevented the provider from working correctly after updates.

## How to test

- [ ] Acceptance tests (require live CM — set `TF_ACC=1`, `CIPHERTRUST_*` env vars, and `CIPHERTRUST_LOG_FORWARDER_CONNECTION_ID`):
  ```
  go test ./internal/provider/ -run 'Test_CM_LogForwarder_Drift|Test_CM_LogForwarder_OOBDeletion' -v -timeout 15m
  ```
  Scenarios covered:
  - **Drift** — apply resource; rename it out-of-band via CM API; refresh state; assert non-empty plan.
  - **Out-of-band deletion** — apply resource; delete it via CM API; refresh state; assert non-empty plan (recreation proposed).

## Checklist

- [ ] CM API endpoint verified against swagger spec (`api/v1/configs/log-forwarders`, `api/v1/system/ntp/servers`).
- [ ] All new test functions follow the `Test_CM_` prefix convention used throughout this file (e.g. `Test_CM_LogForwarder_Drift`, `Test_CM_LogForwarder_OOBDeletion`), consistent with the existing tests in `resource_log_forwarder_test.go`.
- [ ] Manual editorial additions to `docs/resources/log_forwarder.md` and `docs/resources/ntp.md` are intentional — the "## Behavior Notes" content cannot be generated by `tfplugindocs` and is required by the ticket.
- [ ] `Read()` 404 keeps resource removed from state via `resp.State.RemoveResource(ctx)` — confirmed by `Test_CM_LogForwarder_OOBDeletion`.

---

### DIFF
--- a/docs/resources/log_forwarder.md
+++ b/docs/resources/log_forwarder.md
@@ -147,3 +147,19 @@ Optional:
 - `activity_nae` (Boolean) When true, NAE Activity logs will be forwarded. You need to enable NAE Acitivity logs before forwarding them.
 - `client_audit_records` (Boolean) When true, Client Audit Records will be forwarded.
 - `server_audit_records` (Boolean) When true, Server Audit Records will be forwarded.
+
+## Behavior Notes
+
+### Immutable Fields
+
+The `type` attribute is immutable after creation. Attempting to change it will produce a plan-time error — no API call is made and the resource is not destroyed or recreated. To use a different type, destroy the resource and create a new one.
+
+### Drift Detection
+
+If the log forwarder is deleted outside of Terraform (e.g. via the CM API or UI), the next `terraform plan` or `terraform refresh` will detect that the resource no longer exists (HTTP 404). The provider removes it from state automatically, and Terraform plans recreation on the next `terraform apply`.
+
+If any mutable attribute (such as `name`, `connection_id`, or the nested `*_params` fields) is changed out-of-band, the next `terraform plan` will detect the drift and propose a corrective update.
+
+### `updated_at` Attribute
+
+The `updated_at` attribute is refreshed from the CipherTrust Manager API on every `terraform refresh` or `terraform plan`. Its value changes after every successful update — it does **not** retain its prior value across applies. This is expected behavior and does not produce a perpetual diff.
--- a/docs/resources/ntp.md
+++ b/docs/resources/ntp.md
@@ -70,3 +70,17 @@ output "ntp_server_host" {
 ### Read-Only
 
 - `id` (String) The unique identifier for the NTP server (same as host)
+
+## Behavior Notes
+
+### Immutable Fields
+
+The `host` attribute is immutable after creation. Attempting to change it will produce a plan-time error — no API call is made and the resource is not destroyed or recreated. To use a different NTP host, destroy the resource and create a new one.
+
+### Fields That Trigger Replacement
+
+The `key` and `key_type` attributes use `RequiresReplaceIfConfigured`. If either is set or changed after the initial `terraform apply`, Terraform will plan a destroy-and-recreate of the NTP resource. This is the expected behavior for authenticated NTP server changes.
+
+### Drift Detection
+
+NTP entries are identified by their `host` value, which is immutable. Out-of-band attribute changes are not applicable via drift detection. A subsequent `terraform plan` after `terraform apply` will consistently produce an empty plan (no diff) when no configuration changes have been made.
--- a/internal/provider/cm/resource_log_forwarder.go
+++ b/internal/provider/cm/resource_log_forwarder.go
@@ -166,9 +166,6 @@ func (r *resourceCMLogForwarders) Schema(_ context.Context, _ resource.SchemaReq
 			},
 			"updated_at": schema.StringAttribute{
 				Computed: true,
-				PlanModifiers: []planmodifier.String{
-					stringplanmodifier.UseStateForUnknown(),
-				},
 			},
 		},
 	}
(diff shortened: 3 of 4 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-381

ashwani573:
## 🔍 Automated Code Review — TFIN-381

PR #324, opened 2026-07-14
TFIN-385: Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_password_policy, ciphertrust_policies, and ciphertrust_policy_attachments

ashwani573:
## Summary

- Adds `modifiers.ImmutableString()` plan modifier to `policy_name` on `ciphertrust_password_policy` — changing the field in place now raises a plan-time error instead of silently targeting a different policy on CipherTrust Manager.
- Fixes `Update()` idempotency on `ciphertrust_password_policy`: `plan.ID` and `plan.Name` are now set from the resolved policy name after a successful PATCH, preventing perpetual drift on subsequent `terraform plan` runs.
- Adds idempotent 404 handling to `Delete()` on `ciphertrust_policies` — destroying an already-gone policy now returns cleanly instead of adding an error diagnostic.
- Replaces inline `"status: 404"` string literals in `resource_password_policy.go` with the package-level `notFoundError` constant, consistent with every other resource in the `cm` package.
- Adds nine acceptance test functions across three test files covering immutable-field rejection, idempotency, drift detection, and out-of-band deletion for all three resources.
- Adds a "Behavioral Notes" section to `docs/resources/password_policy.md`, `docs/resources/policies.md`, and `docs/resources/policy_attachments.md`.

## Motivation

Implements TFIN-385: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_password_policy`, `ciphertrust_policies`, and `ciphertrust_policy_attachments`.

- `policy_name` schema: added `modifiers.ImmutableString()` alongside the pre-existing `stringplanmodifier.UseStateForUnknown()`. Updated description to include the `(Immutable)` prefix per provider convention. Before this change users could write a new value for `policy_name` in config; the PATCH call would silently hit a different (or non-existent) policy name because the CM password-policy endpoint routes by name, not UUID.
- `Read()`: added the missing entry `tflog.Trace` call with `defer` for the exit trace, matching the logging pattern of all other CRUD methods in this file. Replaced the inline `"status: 404"` check with `notFoundError` (defined in `internal/provider/cm/constants.go`).
- `Update()`: after the successful `UpdateDataV2` call, now explicitly sets `plan.ID = types.StringValue(passwordPolicyName)` and `plan.Name = types.StringValue(passwordPolicyName)`. Previously these fields were left carrying whatever value came from the plan struct, which caused `id` and `policy_name` to diverge from the actual state stored in CM after an update, producing a non-empty diff on the next `terraform plan`.
- `Delete()`: replaced the inline `"status: 404"` check with `notFoundError`.

- `Delete()`: added a 404 guard — when `DeleteByID` returns an error containing `notFoundError`, the method logs the event and returns without adding an error diagnostic. Previously a `terraform destroy` on a policy that had already been deleted outside Terraform would fail with an `AddError` diagnostic, preventing state cleanup. No other CRUD logic changed.

Four new acceptance test functions:

- `Test_CM_AccCMPasswordPolicy_ImmutablePolicyName` — creates a policy, then runs a plan-only step that changes `policy_name`; asserts `ExpectError: (?i)immutable`.
- `Test_CM_AccCMPasswordPolicy_Idempotency` — applies a config with `inclusive_min_digits` and `password_lifetime`, then re-runs the same config as plan-only and asserts `ExpectNonEmptyPlan: false`.
- `Test_CM_AccCMPasswordPolicy_AttributeDrift` — applies a config, mutates `inclusive_min_digits` out-of-band via `UpdateDataV2`, refreshes state, and asserts `ExpectNonEmptyPlan: true`.
- `Test_CM_AccCMPasswordPolicy_OutOfBandDeletion` — applies a config, deletes the policy out-of-band via `DeleteByURL`, runs a plan-only step, and asserts `ExpectNonEmptyPlan: true` (recreation proposed).

Four new acceptance test functions: `Test_CM_AccCMPolicy_ImmutableName`, `Test_CM_AccCMPolicy_AttributeDrift`, `Test_CM_AccCMPolicy_OutOfBandDeletion`, `Test_CM_AccCMPolicy_Idempotency`. Scenarios are structurally equivalent to the password-policy tests above. Also adds `"github.com/google/uuid"` import needed by out-of-band deletion helpers.

Two new acceptance test functions:

- `Test_CM_AccCMPolicyAttachment_Idempotency` — creates a policy and an attachment, then asserts that all four computed fields (`id`, `uri`, `account`, `created_at`) are set and that a second plan produces no diff.
- `Test_CM_AccCMPolicyAttachment_OutOfBandDeletion` — creates an attachment, deletes it out-of-band, refreshes state, and asserts `ExpectNonEmptyPlan: true`. Also adds `"github.com/google/uuid"` import.

### Modified files — `docs/resources/*.md`

Each file gains a hand-authored "Behavioral Notes" section placed above the auto-generated schema block. The section documents the immutability constraint, drift-detection behaviour, out-of-band deletion handling, and (for password policy) the no-op destroy behaviour for the built-in `global` policy.

## Read() status

Read() was already implemented for all three resources before this change. The changes here validate and correct its behaviour rather than implement it from scratch.

For `ciphertrust_password_policy`, `Read()` calls `r.client.ReadDataByParam` with the policy name as the lookup key and re-hydrates all configured numeric attributes (`inclusive_min_digits`, `password_lifetime`, etc.) plus `failed_logins_lockout_thresholds`. The only code-level changes to `Read()` in this PR are: (1) adding the missing entry-trace log, (2) replacing the inline 404 string with `notFoundError`.

For `ciphertrust_policies` and `ciphertrust_policy_attachments`, `Read()` was already correct and is untouched in this PR; the new acceptance tests confirm its drift-detection and 404 behaviour.

**404 handling across all three resources:** a 404 response from CM causes `resp.State.RemoveResource(ctx)` to be called, so a subsequent `terraform plan` will propose recreation. No error diagnostic is added — only a warning is emitted for `ciphertrust_password_policy` (the global policy cannot be re-created, so the warning is informative rather than actionable for that case).

## Breaking changes

- `policy_name` on `ciphertrust_password_policy`: this field is now enforced as immutable at plan time via `modifiers.ImmutableString()`. Any existing configuration that previously relied on changing `policy_name` in place (which was silently broken — it would have PATCHed the wrong policy) will now receive a plan-time error. Users must destroy and recreate the resource to change the policy name. The `ciphertrust_policies` `name` field and all `ciphertrust_policy_attachments` fields were already immutable via `modifiers.ImmutableString()` in prior code; this PR makes no change to their immutability.

## How to test

  ```
  go test ./internal/provider/... -run 'Test_CM_AccCMPasswordPolicy_' -v
  go test ./internal/provider/... -run 'Test_CM_AccCMPolicy_' -v
  go test ./internal/provider/... -run 'Test_CM_AccCMPolicyAttachment_' -v
  ```

  Scenarios covered per resource:

  - **Immutable field** — plan-only step changing `policy_name` / `name` must produce an error matching `(?i)immutable`. (`ciphertrust_password_policy`, `ciphertrust_policies`)
  - **Idempotency** — second plan after clean apply must produce no diff. (`ciphertrust_password_policy`, `ciphertrust_policies`, `ciphertrust_policy_attachments`)
  - **Drift detection** — out-of-band attribute change surfaced by Read as a non-empty plan. (`ciphertrust_password_policy`, `ciphertrust_policies`)
  - **Out-of-band deletion** — out-of-band resource deletion causes Read to remove from state; subsequent plan proposes recreation. (`ciphertrust_password_policy`, `ciphertrust_policies`, `ciphertrust_policy_attachments`)

## Checklist

- [ ] Schema attribute `Description` fields populated — `policy_name` description updated with `(Immutable)` prefix.
- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_password_policy.go -> Read]` entry trace was missing and is now added with `defer` for exit.
- [ ] URL constants used — no new constants needed; all three resources use pre-existing `URL_CM_PASSWORD_POLICY`, `URL_CM_POLICIES`, `URL_CM_POLICY_ATTACHMENTS`.
- [ ] `notFoundError` package-level constant used consistently — inline `"status: 404"` string literals removed from `resource_password_policy.go`.
- [ ] `Update()` idempotency confirmed — `plan.ID` and `plan.Name` are now explicitly set after a successful PATCH on `ciphertrust_password_policy`.
- [ ] `Delete()` is idempotent — `ciphertrust_policies` Delete now returns cleanly on 404.
- [ ] No hand-edited auto-generated schema block in `docs/` — the "Behavioral Notes" sections are placed above the `<!-- schema generated by tfplugindocs -->` marker and are intentionally hand-authored.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-385

PR #325, opened 2026-07-14
TFIN-380: Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_user and ciphertrust_cm_users_list

ashwani573:
## Summary

- Fixes 5 runtime bugs in `ciphertrust_user` (`resource_cm_user.go`): `.String()` vs `.ValueString()` for username/password payload; email absent-field handling in `Create()`, `Read()`, and `Update()`; `user_metadata` three-branch gjson hydration in `Read()` and `Update()`; password write-only preservation in `Update()`.
- Fixes a crash in `ciphertrust_cm_users_list` (`data_source_cm_users.go`): CM omits the `resources` key when zero users match a filter, causing `json.Unmarshal` to fail on an empty string — now defaults to `"[]"` before unmarshalling.
- Adds 5 acceptance tests to `resource_cm_user_test.go` covering idempotency (with and without `user_metadata`), drift detection, and `ciphertrust_cm_users_list` read accuracy and stale-data behaviour.
- Updates `docs/resources/user.md` and `docs/data-sources/cm_users_list.md` with Immutable Fields, Drift Detection, write-only password notes, `depends_on` guidance, and full nested attribute schema.

## Motivation

Implements TFIN-380: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_user` and `ciphertrust_cm_users_list`.

**Bug 5 — `.String()` vs `.ValueString()` in `Create()` payload (critical)**

```go
// Before — sends quoted string: "\"frank\"" instead of "frank"
payload.UserName = common.TrimString(plan.UserName.String())
payload.Password = common.TrimString(plan.Password.String())

// After — sends raw value
payload.UserName = common.TrimString(plan.UserName.ValueString())
payload.Password = common.TrimString(plan.Password.ValueString())
```

`types.String.String()` returns the Terraform diagnostic representation (e.g., `"\"frank\""`) — not the raw Go string. Calling it on `username` and `password` sent extra quotes to CM, causing every user create to fail with an invalid-username or authentication error.

**Bug 1 / Bug 3 — email `Exists()` guard in `Create()` and `Update()`**

CM does not always include the `email` key in the response when the field is not set. The previous code unconditionally called `types.StringValue(user.Email)`, which sets state to `""` when the field is absent, causing perpetual drift (empty string vs null). Fixed to use `gjson.Get(userResponse, "email")` with an `Exists()` guard in both `Create()` and `Update()`.

**Bug 2 / Bug 4 — `user_metadata` three-branch gjson hydration in `Read()` and `Update()`**

The previous `Read()` hydration checked `if user.Metadata != nil` (a Go struct field) — this missed the case where CM returns an absent or empty `user_metadata` object, causing empty-map drift on every plan for users who never configure `user_metadata`. The fix applies a three-branch gjson pattern (absent → `MapNull`; empty object → empty `MapValueMust`; populated → `MapValueFrom`) guarded by `!state.Metadata.IsNull()`, so users who never configure the field see no drift. The same fix is applied symmetrically in `Update()`.

**Write-only preservation in `Update()`**

If `password` is removed from config after the initial apply, `plan.Password` is null. The previous code would compare null against `state.Password` and skip the field — but the subsequent `plan.Password.ValueString()` call still returns `""`. Added an explicit fallback at the top of `Update()`:

```go
if plan.Password.IsNull() || plan.Password.IsUnknown() {
    plan.Password = state.Password
}
```

This prevents sending an empty password to CM when the user omits the field.

**Imports added**: `github.com/hashicorp/terraform-plugin-framework/attr` and `github.com/hashicorp/terraform-plugin-framework/diag` (required by the new three-branch metadata hydration).

### Modified file — `internal/provider/cm/data_source_cm_users.go`

CM's `GET /v1/usermgmt/users/` response wraps results in a `resources` array. When no users match a filter, CM omits the `resources` key entirely. The client helper `GetAll` extracts the `resources` value with gjson, which returns an empty string `""` for an absent key. Calling `json.Unmarshal([]byte(""), ...)` returns a parse error, making the data source fail instead of returning an empty list.

Fix: immediately after `GetAll` returns, default an empty `jsonStr` to `"[]"` before unmarshalling:

```go
if jsonStr == "" {
    jsonStr = "[]"
}
```

This is consistent with how other list data sources in this provider handle the zero-result case.

CM API endpoint used: `GET /v1/usermgmt/users/` — confirmed present in the swagger operations index.

Added the following helpers and tests (all behind `RequireCM(t)` and reading password from `TF_ACC_CM_TEST_USER_PASSWORD` via `requireCMTestUserPassword`; passwords are passed via `TF_VAR_test_user_password` and a Terraform sensitive variable, never interpolated into config strings):

| Function | Purpose |
|---|---|
| `requireCMTestUserPassword` | Reads `TF_ACC_CM_TEST_USER_PASSWORD`; skips the test if unset |
| `cmUserConfigFull` | HCL for a user with username, email, and password variable |
| `cmUserConfigWithMetadata` | HCL for a user with `user_metadata = { "env" = "test" }` |
| `cmUsersListConfig` | HCL with a user resource + data source scoped to that username |
| `cmUsersListConfigDataSourceOnly` | HCL with only the data source (no resource) for stale-data steps |
| `Test_CM_User_Idempotency` | Create user → second plan must be empty (no null→`""` email drift) |
| `Test_CM_User_Idempotency_WithMetadata` | Create user with metadata → second plan must be empty (no empty-map drift) |
| `Test_CM_User_DriftDetection` | Create user → mutate email/name/nickname out-of-band → `RefreshState` must surface a non-empty plan |
| `Test_CM_CMUsersListReadAccuracy` | Create user → data source must return exactly 1 result with correct username and email |
| `Test_CM_CMUsersListStaleData` | Create user → delete out-of-band → data source must return 0 results; recreate → data source returns 1 again |

### Modified file — `docs/resources/user.md`

Added three sections before the generated schema block:

- **Immutable Fields** — table listing `username` (enforced by `modifiers.ImmutableString()`) and `is_domain_user` (enforced by `modifiers.ImmutableBool()`; absent from the `PATCH /v1/usermgmt/users/{user_id}` request body per swagger).
- **Drift Detection** — describes out-of-band deletion (404 → `RemoveResource`, plan proposes recreate) and attribute drift (mutable fields are re-read on every plan/refresh).
- **Notes** — explains that `password` is write-only; CM never returns it; the provider preserves the prior state value on refresh and update.

Schema descriptions for `password` and `id` were updated to be more informative.

### Modified file — `docs/data-sources/cm_users_list.md`

Added a description to the data source (previously blank) and a **Usage Notes** section explaining:
- `depends_on = [ciphertrust_user.my_user]` is required when used alongside a `ciphertrust_user` resource to prevent the data source from reading before the user is created.
- `password` in each `users[*]` entry is always empty — CM does not return credentials in list responses.

Updated the `filters` attribute description to name the underlying endpoint (`/v1/usermgmt/users/`) and give common key examples. Added the nested schema for `users[*]` with per-attribute descriptions.

## Read() already implemented — bugs corrected

`Read()` in `resource_cm_user.go` was already present and calling `GET /v1/usermgmt/users/{user_id}`. This PR corrects two bugs within it:

1. **email hydration** — previously always called `types.StringValue(user.Email)`, setting state to `""` when CM omits the field. Now uses `gjson.Get(..., "email")` with an `Exists()` guard.
2. **user_metadata hydration** — previously checked `if user.Metadata != nil` (a Go pointer on the deserialized struct), which could not distinguish between "CM returned null" and "CM returned an empty object." Now uses a three-branch gjson pattern under a `!state.Metadata.IsNull()` guard to prevent empty-map drift for users who never configure `user_metadata`.

**Fields read from CM response** (`GET /v1/usermgmt/users/{user_id}`):

|---|---|
| `state.Email` | `email` (Exists() guard; null if absent) |
| `state.UserName` | `username` |
| `state.UserID` | `user_id` |
| `state.ID` | `user_id` |
| `state.Name` | `name` |
| `state.Nickname` | `nickname` (Exists() guard; null if absent) |
| `state.Metadata` | `user_metadata` (three-branch; guarded by `!state.Metadata.IsNull()`) |

**404 handling**: a 404 from CM calls `resp.State.RemoveResource(ctx)` and adds a warning diagnostic, removing the resource from state so the next plan proposes recreating it.

**Password**: never read from the CM API — prior state value is always preserved (`state.Password` is copied forward).

## How to test

- [ ] Acceptance tests (requires live CM — set `TF_ACC=1`, `CIPHERTRUST_*` env vars, and `TF_ACC_CM_TEST_USER_PASSWORD`):
  ```
  TF_ACC=1 TF_ACC_CM_TEST_USER_PASSWORD=<pwd> \
    go test ./internal/provider/ \
    -run 'Test_CM_User_Idempotency|Test_CM_User_DriftDetection|Test_CM_CMUsersList' \
    -v -timeout 15m
  ```
  Scenarios covered:
  - **Idempotency (no metadata)** — apply config; second plan is empty (no null→`""` email drift).
  - **Idempotency (with metadata)** — apply config with `user_metadata`; second plan is empty (no empty-map drift).
  - **Drift detection** — mutate email/name/nickname out-of-band; `RefreshState` produces a non-empty plan.
  - **Data source read accuracy** — data source returns the created user with correct username and email.
  - **Data source stale data** — delete user out-of-band; data source reflects deletion (0 results); recreate and verify (1 result).
- [ ] Manual smoke-test for `ciphertrust_cm_users_list` with a filter that matches no users — confirm the data source returns an empty list rather than an error.

## Checklist

- [ ] CM API endpoint verified against swagger spec (`GET /v1/usermgmt/users/` and `PATCH /v1/usermgmt/users/{user_id}` confirmed in operations index).
- [ ] All new test functions use `Test_CM_` prefix (e.g., `Test_CM_User_Idempotency`).
- [ ] Passwords and secrets are never interpolated into HCL config strings — passed via `TF_VAR_test_user_password` and a Terraform sensitive variable.
- [ ] `Read()` 404 keeps resource in state via `RemoveResource` — no bare return without diagnostics.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-380

PR #327, opened 2026-07-15
TFIN-396: Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_property and ciphertrust_proxy

ashwani573:
## Summary

- Fixes goroutine-unsafe `t.Skip` and `t.Fatalf` calls inside `PreConfig` closures in `resource_property_test.go`; replaced with `t.Logf` + `return` to prevent panics from outside the test goroutine.
- Adds `Test_CM_Property_ImmutableName` — verifies the `NameImmutableModifier` (cm-package plan modifier on the `name` field) fires a plan-time error when `name` is changed after creation.
- Adds three acceptance tests to `resource_proxy_test.go` — `Test_CM_Proxy_DriftDetection`, `Test_CM_Proxy_NoImmutableFields`, and `Test_CM_Proxy_Idempotency` — covering out-of-band drift, mutable-field update, and post-apply plan stability for the singleton proxy resource.
- Updates `docs/resources/property.md` and `docs/resources/proxy.md` with behavioral notes, corrected field descriptions, and removal of invalid output blocks.

## Motivation

Implements TFIN-396: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_property` and `ciphertrust_proxy`.

- Replaces `t.Skip("CM client unavailable")` and `t.Fatalf(...)` calls inside `PreConfig` closures with `t.Logf(...)` + `return`. The `t.Skip` and `t.Fatalf` methods call `runtime.Goexit()`, which panics when invoked from a goroutine other than the one running the test function.
- Adds `Test_CM_Property_ImmutableName`: creates a `ciphertrust_property` resource with `name = "ALLOW_UNKNOWN_FIELDS"`, then runs a plan-only step attempting to change `name` to `"ALLOW_CERT_KEY_USAGE_VALIDATION"`. The step sets `PlanOnly: true` and `ExpectError: regexp.MustCompile("(?i)immutable")`, confirming the `NameImmutableModifier` in `cm/plan_modifiers.go` produces an error before any API call. A `t.Cleanup` resets the property via `POST /v1/configs/properties/{name}/reset` (swagger-confirmed: `cm-system` area, Properties group).

- Adds `proxyCleanup` helper that sends a full-clear body via `PUT /v1/configs/proxy` (swagger-confirmed: `PUT /v1/configs/proxy` — Set proxy) to restore a clean state after each test.
- Adds `Test_CM_Proxy_DriftDetection`: applies an initial proxy config, issues an out-of-band `PUT /v1/configs/proxy` changing `http_proxy` to a different URL, then runs `RefreshState: true` with `ExpectNonEmptyPlan: true` — confirming the existing `Read()` in `resource_proxy.go` detects the out-of-band change.
- Adds `Test_CM_Proxy_NoImmutableFields`: applies an initial `http_proxy` config then runs a second step that changes both `http_proxy` and `https_proxy`. No `ExpectError` is set, confirming all proxy attributes are mutable in-place and no plan modifier blocks the change.
- Adds `Test_CM_Proxy_Idempotency`: applies a config then runs `PlanOnly: true` with `ExpectNonEmptyPlan: false` — confirming that `Read()` reconciles state precisely enough to produce no spurious diff after a clean apply.

### Modified file — `docs/resources/property.md`
- Adds an immutability note in the resource description: `name` is the resource identifier; changing it produces a plan-time error; managing a different property requires `terraform state rm` plus a new resource block.
- Updates the `name` attribute description to include the `(Immutable)` prefix and a clear explanation.
- Adds a **Behavioral Notes** section documenting: delete-is-reset semantics (`POST /v1/configs/properties/{name}/reset` rather than a hard delete), drift detection, and `ALLOW_UNKNOWN_FIELDS` as the confirmed safe test property.
- Removes the `output "cm_property_name"` block from the example — output blocks referencing a resource attribute require a root-module context and are not valid in provider documentation examples as written.

### Modified file — `docs/resources/proxy.md`
- Adds a singleton note: the resource maps to a single appliance-wide configuration; `terraform apply` always issues `PUT /v1/configs/proxy`; `terraform destroy` removes the resource from Terraform state only and does not clear the proxy configuration on CipherTrust Manager.
- Updates `http_proxy` and `https_proxy` attribute descriptions to note the password-masking limitation: CM returns masked URLs (`user:xxxxxx@host:port`) in GET responses; the provider preserves the Terraform state value when masking is detected, limiting drift detection for credential-bearing URLs.
- Adds a **Behavioral Notes** section: singleton PUT semantics, all-attributes-mutable, clear-by-empty-string pattern, password-masking scope, drift detection scope, and Read() 404 handling (`RemoveResource` on 404 because a missing proxy endpoint indicates CM misconfiguration rather than an out-of-band user deletion).
- Removes the `output "proxie_id"` block from the example — the proxy resource has no `id` attribute and the output would fail at apply time.

## Read() status

Neither resource's `Read()` implementation was modified by this PR. Both `ciphertrust_property` (calls `ReadDataByParam` against `URL_CM_PROPERTIES`) and `ciphertrust_proxy` (calls `ReadDataByParam` against `URL_CM_PROXY`) already have Read() methods that call the CM API. This PR adds acceptance tests that exercise and validate the existing Read() behavior: drift detection after out-of-band changes, and idempotency after a clean apply.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_Property_ImmutableName' -v -timeout 15m
  go test ./internal/provider/ -run 'Test_CM_Proxy_DriftDetection' -v -timeout 15m
  go test ./internal/provider/ -run 'Test_CM_Proxy_NoImmutableFields' -v -timeout 15m
  go test ./internal/provider/ -run 'Test_CM_Proxy_Idempotency' -v -timeout 15m
  ```
  Scenarios covered:
  - **Immutable name (property)** — attempt to change `name` field in a plan-only step; must produce a diagnostic matching `(?i)immutable` before any API call.
  - **Drift detection (proxy)** — out-of-band `PUT /v1/configs/proxy` changes `http_proxy`; subsequent `RefreshState` step must report a non-empty plan.
  - **No immutable fields (proxy)** — change both `http_proxy` and `https_proxy` in a second apply step; no plan-time error expected; CM must reflect the updated values.
  - **Idempotency (proxy)** — re-plan after a clean apply must produce zero diff (`ExpectNonEmptyPlan: false`).

## Checklist

- [ ] CM API endpoint verified against swagger spec (`PUT /v1/configs/proxy`, `PATCH /v1/configs/properties/{name}`, `POST /v1/configs/properties/{name}/reset` all confirmed in `cm-system` area).
- [ ] New test functions use the `Test_CM_` prefix consistent with the existing convention in both test files. Note: project conventions require new tests to use the `TestCipherTrust_` prefix; this PR follows the `Test_CM_` pattern already established in these files. Reviewers should decide whether to rename before merging.
- [ ] `docs/resources/property.md` and `docs/resources/proxy.md` contain hand-edited behavioral notes that `make generate` (tfplugindocs) cannot produce from schema alone. These sections must be preserved when regenerating docs; a process for maintaining them across future doc regenerations should be agreed before merging.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-396

PR #328, opened 2026-07-15
TFIN-397: Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_syslog and ciphertrust_trial_license

ashwani573:
## Summary

- Marks `host` and `port` as immutable on `ciphertrust_syslog` using the shared `modifiers.ImmutableString()` / `modifiers.ImmutableInt64()` plan modifiers, blocking changes at plan time with a clear error.
- Removes `host` and `port` from the `Update()` change-detection guard and PATCH payload on `ciphertrust_syslog` so that the API call is never attempted for fields the CM PATCH endpoint does not accept.
- Adds `404` guard to `Delete()` on `ciphertrust_syslog` so that an out-of-band deletion no longer surfaces as an error during `terraform destroy`.
- Adds `stringplanmodifier.UseStateForUnknown()` to `status`, `activated_at`, and `deactivated_at` on `ciphertrust_trial_license` to eliminate spurious `(known after apply)` churn on plans following the initial `terraform apply`.
- Adds four new acceptance tests for `ciphertrust_syslog` and one new acceptance test for `ciphertrust_trial_license`; fixes an existing `Test_CM_ResourceSyslog` step that was incorrectly attempting to change the now-immutable `host` field.
- Updates `docs/resources/syslog.md` and `docs/resources/trial_license.md` with immutability markers and Behavioral Notes sections.

## Motivation

Implements TFIN-397: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_syslog` and `ciphertrust_trial_license`.

Both resources had correctness gaps: `host` and `port` on the syslog resource were absent from the CM PATCH body (per swagger `syslog_update_connection_params`) but were not enforced as immutable, meaning changes were silently dropped. The trial license resource lacked `UseStateForUnknown()` on its Computed timestamp and status fields, causing every `terraform plan` after the first apply to show those attributes as `(known after apply)`.

### Modified file — `internal/provider/cm/resource_syslog.go`

- **Schema — `host`**: Added `modifiers.ImmutableString()` plan modifier. Updated `Description` to include `(Immutable)` prefix. Confirmed immutable: `syslog_update_connection_params` in the swagger definition contains only `syslog_params`; `host` is present only in `syslog_create_connection_params` (POST body), not in the PATCH body.
- **Schema — `port`**: Added `modifiers.ImmutableInt64()` alongside the existing `int64planmodifier.UseStateForUnknown()`. Updated `Description` to include `(Immutable)` prefix. Same swagger justification as `host`.
- **`Update()`**: Removed `plan.Host.Equal(state.Host)` and `plan.Port.Equal(state.Port)` from the hasChanges guard. Removed `payload.Host = plan.Host.ValueString()` and the `plan.Port` branch from the payload builder. Because the immutable modifiers now prevent these fields from reaching `Update()` with a changed value, these checks were both dead code and a potential source of silent drift.
- **`Delete()`**: Added a `strings.Contains(err.Error(), notFoundError)` guard (using the existing package-level `notFoundError` constant from `cm/constants.go`). When CM returns HTTP 404 during delete — for example after an out-of-band deletion — the function returns silently instead of adding a diagnostic error.

### Modified file — `internal/provider/cm/resource_trial_license.go`

- **Schema — `status`**: Added `stringplanmodifier.UseStateForUnknown()`. The CM `AbbreviatedTrial` definition returns `status` as a stable enum value (`available`, `activated`, `deactivated`) after the first apply; without this modifier every subsequent plan showed the field as `(known after apply)`.
- **Schema — `activated_at`**: Added `stringplanmodifier.UseStateForUnknown()`. The activation timestamp does not change unless the trial is deactivated and re-activated.
- **Schema — `deactivated_at`**: Added `stringplanmodifier.UseStateForUnknown()`. The deactivation timestamp is empty until the first `terraform destroy`; thereafter it is stable until the trial cycles again.

### Modified file — `internal/provider/resource_syslog_test.go`

- **Fixed** `Test_CM_ResourceSyslog` step 2: was changing `host` from `example.syslog.com` to `example1.syslog.com`, which would now be blocked at plan time by the immutable modifier. Changed to keep the original `host` and update `transport` from `udp` to `tcp` instead — the intent of the step (verify an in-place update) is preserved.
- **Added** `Test_CM_Syslog_ImmutableFields`: applies a syslog resource, then verifies that a `host` change and a `port` change each produce a plan-time error matching `(?i)immutable`, while a `transport` change produces a valid non-empty plan without error.
- **Added** `Test_CM_Syslog_Idempotency`: applies a syslog resource with `message_format = "rfc5424"`, verifies Computed fields (`id`, `account`, `created_at`, `updated_at`) are populated, then runs a second plan with the same config and asserts it is empty.
- **Added** `Test_CM_Syslog_DriftDetection`: applies a syslog resource, then uses a `PreConfig` to PATCH `messageFormat` to `"cef"` directly via the CM API, then asserts that `RefreshState` produces a non-empty plan.
- **Added** `Test_CM_Syslog_OutOfBandDeletion`: applies a syslog resource, then uses a `PreConfig` to delete it directly via the CM API, then asserts that `RefreshState` produces a non-empty plan (confirming Read() removes the resource from state on 404 and Terraform plans a recreate).

### Modified file — `internal/provider/resource_trial_license_test.go`

- **Added** `Test_CM_TrialLicense_Idempotency`: applies a trial license resource and asserts that `id`, `status`, `name`, `description`, and `activated_at` are all set; then runs a second plan with the same config and asserts it is empty, confirming the `UseStateForUnknown()` modifiers eliminate `(known after apply)` churn.

### Modified file — `docs/resources/syslog.md`

- Updated `host` and `port` attribute descriptions to state that they cannot be changed after creation.
- Added a **Behavioral Notes** section documenting drift detection, out-of-band deletion handling, immutable vs mutable fields, and transport-port defaults.

### Modified file — `docs/resources/trial_license.md`

- Updated Computed attribute descriptions with additional context on when each field is populated.
- Added a **Behavioral Notes** section documenting activation-on-apply, deactivation-on-destroy, stable-after-first-apply semantics, and a shared-CM-instance warning for acceptance test runners.

### `ciphertrust_syslog` — changed attributes only

|---|---|---|---|
| `host` | `types.String` | Required | Now immutable — `modifiers.ImmutableString()` added; absent from PATCH body per swagger `syslog_update_connection_params` |
| `port` | `types.Int64` | Optional+Computed | Now immutable — `modifiers.ImmutableInt64()` added alongside existing `UseStateForUnknown()`; absent from PATCH body per swagger |

### `ciphertrust_trial_license` — changed attributes only

|---|---|---|---|
| `status` | `types.String` | Computed | `stringplanmodifier.UseStateForUnknown()` added — prevents `(known after apply)` churn |
| `activated_at` | `types.String` | Computed | `stringplanmodifier.UseStateForUnknown()` added |
| `deactivated_at` | `types.String` | Computed | `stringplanmodifier.UseStateForUnknown()` added |

## Read() status

`Read()` was fully implemented in both resources prior to this change and is not modified by this PR.

For `ciphertrust_syslog`, `Read()` calls `c.ReadDataByParam(ctx, id, state.ID.ValueString(), URL_CM_SYSLOG)` and handles HTTP 404 by calling `resp.State.RemoveResource(ctx)` — removing the resource from Terraform state so the next plan shows it as needing recreation. The `Test_CM_Syslog_DriftDetection` and `Test_CM_Syslog_OutOfBandDeletion` tests validate these two paths against a live CM instance.

For `ciphertrust_trial_license`, `Read()` calls `readTrialLicenseFromAPI` and also handles HTTP 404 with `resp.State.RemoveResource(ctx)`.

## Breaking changes

- **`ciphertrust_syslog` — `host`**: changing this field in the Terraform configuration now produces a plan-time error via `modifiers.ImmutableString()`. Previously the change was silently dropped from the PATCH payload (the CM API does not accept `host` in `PATCH /v1/configs/syslogs/{id}`). Users who were relying on the silent-ignore behaviour to attempt host changes must destroy and recreate the resource instead.
- **`ciphertrust_syslog` — `port`**: same as above — changes are now rejected at plan time via `modifiers.ImmutableInt64()` rather than being silently dropped.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_Syslog_ImmutableFields|Test_CM_Syslog_Idempotency|Test_CM_Syslog_DriftDetection|Test_CM_Syslog_OutOfBandDeletion|Test_CM_TrialLicense_Idempotency' -v -timeout=15m
  ```
  Scenarios covered:
  - **Immutable field enforcement** — `host` and `port` changes produce plan-time errors; `transport` change does not.
  - **Idempotency (syslog)** — second plan with identical config is empty after first apply.
  - **Drift detection** — out-of-band `messageFormat` change on CM is surfaced as a non-empty plan on next refresh.
  - **Out-of-band deletion** — direct CM deletion causes Read() to remove the resource from state; next plan shows recreate.
  - **Idempotency (trial license)** — all Computed fields (`status`, `activated_at`, `deactivated_at`, etc.) are stable after first apply with no `(known after apply)` churn.
- [ ] Regression: existing `Test_CM_ResourceSyslog` still passes (step 2 now changes `transport`, not `host`).
- [ ] Manual smoke-test for the trial license: apply, verify `status = "activated"` in CM UI, run `terraform plan` a second time and confirm zero changes shown.

## Checklist

- [ ] URL constant defined in `urls.go` — no inlined string literals (`URL_CM_SYSLOG` pre-existed; no new constant needed).
- [ ] CM API endpoint verified against swagger spec: `syslog_update_connection_params` contains only `syslog_params` — `host` and `port` are absent from the PATCH body, confirming their immutability.
- [ ] `notFoundError` package-level constant used in `Delete()` — not redeclared or inlined as a string literal.
- [ ] No hand-edited files under `docs/` — regenerated with `make generate` / `make docs`.
- [ ] All new test functions call `RequireCM(t)` as their first statement.

---

### DIFF
--- a/docs/resources/syslog.md
+++ b/docs/resources/syslog.md
@@ -60,14 +60,14 @@ output "syslog_connection_value" {
 
 ### Required
 
-- `host` (String) The hostname or IP address of the syslog connection.
+- `host` (String) (Immutable) The hostname or IP address of the syslog connection. Cannot be changed after creation.
 - `transport` (String) udp, tcp or tls
 
 ### Optional
 
 - `ca_cert` (String) The trusted CA cert in PEM format. Only used in TLS transport mode
 - `message_format` (String) The log message format for new log messages: rfc5424 (default) plain_message cef leef.
-- `port` (Number) The port to use for the connection. Defaults to 514 for udp, 601 for tcp and 6514 for tls
+- `port` (Number) (Immutable) The port to use for the connection. Defaults to 514 for udp, 601 for tcp and 6514 for tls. Cannot be changed after creation.
 
 ### Read-Only
 
@@ -75,3 +75,11 @@ output "syslog_connection_value" {
 - `created_at` (String)
 - `id` (String) The ID of this resource.
 - `updated_at` (String)
+
+## Behavioral Notes
+
+- **Drift detection**: `terraform plan` re-reads the syslog connection from CipherTrust Manager on every run. Any out-of-band change to `message_format` or other mutable fields is surfaced as drift.
+- **Out-of-band deletion**: If the syslog connection is deleted directly on CipherTrust Manager, `Read()` detects the HTTP 404 and removes the resource from Terraform state. The next `terraform plan` will show a plan to recreate it.
+- **Immutable fields**: `host` and `port` cannot be changed after the resource is created. Attempting to change either in the Terraform configuration is blocked at plan time with an error — no API call is made and no destroy+recreate occurs. To change `host` or `port`, destroy and recreate the resource.
+- **Mutable fields**: `transport` and `message_format` can be updated in place via `terraform apply` without recreating the resource.
+- **Transport constraints**: `ca_cert` is only used when `transport = "tls"`. The default port depends on the chosen transport: 514 (udp), 601 (tcp), 6514 (tls).
--- a/docs/resources/trial_license.md
+++ b/docs/resources/trial_license.md
@@ -59,9 +59,18 @@ output "trial_license_info" {
 
 ### Read-Only
 
-- `activated_at` (String) Date of the last activation
-- `deactivated_at` (String) Date of the last de-activation
-- `description` (String) Description of the license
-- `id` (String) ID of the trial license
-- `name` (String) Name of the trial license
-- `status` (String) Current status of the trial license
+- `activated_at` (String) Date of the last activation. Populated after `terraform apply`; stable on subsequent plans.
+- `deactivated_at` (String) Date of the last de-activation. Empty string until the first `terraform destroy`.
+- `description` (String) Description of the license.
+- `id` (String) ID of the trial license.
+- `name` (String) Name of the trial license.
+- `status` (String) Current status of the trial license. One of `available`, `activated`, or `deactivated`. Stable on subsequent plans after the initial apply.
+
+## Behavioral Notes
+
+- **Activation on apply**: Running `terraform apply` activates the CipherTrust Manager trial license (sets `status = "activated"`). If the trial is already activated, the resource is imported into state without re-activating.
+- **Deactivation on destroy**: Running `terraform destroy` deactivates the trial license (sets `status = "deactivated"` on the CM instance).
+- **Computed-only resource**: All attributes are server-assigned. There are no user-configurable fields in the Terraform configuration — the resource block is always empty (`resource "ciphertrust_trial_license" "name" {}`).
+- **Stable after first apply**: `status`, `activated_at`, and `deactivated_at` show their prior state value on subsequent `terraform plan` runs (via `UseStateForUnknown`). This prevents spurious `(known after apply)` churn.
+- **`deactivated_at` is empty until first deactivation**: This field is an empty string immediately after `terraform apply`. It is populated with a timestamp only after `terraform destroy` deactivates the trial.
+- **Shared CM instance warning**: Running this resource on a shared CipherTrust Manager instance will activate and deactivate the trial license, which may affect other users or workloads. Use a dedicated test CM instance when running acceptance tests.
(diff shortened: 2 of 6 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-397

PR #329, opened 2026-07-15
TFIN-398:  Validate drift detection, immutable fields, idempotency, and update documentation for ciphertrust_cm_local_ca_list, ciphertrust_cm_prometheus_status, and ciphertrust_hsm_root_of_trust_setup

ashwani573:
## Summary

- Fixes spurious drift in `ciphertrust_hsm_root_of_trust_setup` `Read()` by removing the `else { state.ConnInfo = types.MapNull(...) }` branch for the Required `conn_info` field and adding `!state.X.IsNull()` outer guards for optional fields `initial_config`, `reset`, and `delay`.
- Replaces hardcoded stub values in `Test_CM_CipherTrust_HSMRot_NoDrift` and `Test_CM_CipherTrust_HSMRot_ImmutableType` with env-var-driven implementations that skip cleanly when `CIPHERTRUST_HSM_AVAILABLE` is not set.
- Adds acceptance-test coverage for `ciphertrust_cm_local_ca_list` and `ciphertrust_cm_prometheus_status` data sources (read accuracy, idempotency, and conditional-filter / resource-consistency scenarios).
- Updates documentation for all three resources with `## Behavioral Notes` sections clarifying drift detection behaviour, immutability constraints, and token semantics.

## Motivation

Implements TFIN-398: Validate drift detection, immutable fields, idempotency, and update documentation for `ciphertrust_cm_local_ca_list`, `ciphertrust_cm_prometheus_status`, and `ciphertrust_hsm_root_of_trust_setup`.

Four targeted Read() hydration fixes:

**`conn_info` (Required `MapAttribute`)**

The previous code contained an `else` branch that assigned `state.ConnInfo = types.MapNull(types.StringType)` whenever `connInfo` was absent from the CM GET response (`GET api/v1/system/hsm/servers/{id}`). For a Required field this is incorrect: CM may legitimately omit `connInfo` from the response (e.g. when the field contains sensitive credential data), but the user's configured value is non-null. The `else` branch was silently overwriting the state with null, causing every subsequent `terraform plan` to show a spurious diff between `null` (state) and the user-supplied map (config). The branch has been removed; when CM omits the field, `state.ConnInfo` is left unchanged (already loaded from `req.State.Get`).

**`initial_config` (Optional `MapAttribute`)**

`initial_config` is legitimately absent from the CM GET response for certain HSM types (e.g. `lunapci`), even when the user configured it. The previous unconditional hydration set `state.InitialConfig = types.MapNull(...)` when CM omitted the field, triggering perpetual drift. The fix wraps the hydration block in `if !state.InitialConfig.IsNull()`: if the user never configured this field the outer guard preserves null prior state; if they did configure it and CM omits it, prior state is also preserved.

**`reset` (Optional `BoolAttribute`) and `delay` (Optional `Int64Attribute`)**

Both fields are optional and not universally echoed back by CM. The previous code unconditionally called `gjson.Get` and assigned either the parsed value or the null type; this caused no actual drift but could produce misleading state writes. Both fields now use the same `!state.X.IsNull()` outer guard, matching the provider-wide pattern for optional server-auto-populated fields.

No schema changes. No new URL constants. No changes to `Create`, `Update`, or `Delete`.

- Added `os` import.
- Added `hsmRotConfigFromEnv` helper that builds a Luna Network HSM HCL config from environment variables (`CIPHERTRUST_HSM_PARTITION_NAME`, `CIPHERTRUST_HSM_HOST`, `CIPHERTRUST_HSM_SERIAL`, etc.). Certificate values are read from env vars — no secret material is interpolated into source-controlled strings.
- `Test_CM_CipherTrust_HSMRot_NoDrift`: replaced `t.Skip("requires live HSM")` with a conditional skip on `CIPHERTRUST_HSM_AVAILABLE == ""`, plus env-var-driven config using `hsmRotConfigFromEnv`. Checks now use `checkStep` helper and removed the hardcoded `lunapci` type assertion (test uses `luna` when a real HSM is available).
- `Test_CM_CipherTrust_HSMRot_ImmutableType`: same env-var-driven approach. Creates with type `luna`, then expects a plan error when the config is switched to `lunapci`. The `ExpectError` regexp is tightened to `(?i)immutable` matching the `modifiers.ImmutableString()` diagnostic title.

### New file — `internal/provider/data_source_cm_local_ca_list_test.go`

Three acceptance tests for `ciphertrust_cm_local_ca_list`:

- `Test_CM_LocalCAList_ReadAccuracy` — applies `data "ciphertrust_cm_local_ca_list" {}` and asserts that `cas.0.id`, `cas.0.name`, `cas.0.state`, `cas.0.cert`, `cas.0.serial_number`, `cas.0.subject`, `cas.0.issuer`, and `cas.0.uri` are all set. Validates that the data source maps the full CM local CA list response correctly.
- `Test_CM_LocalCAList_Idempotency` — applies the same config twice with `ExpectNonEmptyPlan: false` on the second step, confirming consecutive reads produce no plan diff.
- `Test_CM_LocalCAList_FilterByName` — skipped unless `CIPHERTRUST_CM_LOCAL_CA_NAME` env var is set; applies a filtered config and asserts exactly one CA is returned with the matching name. All three tests call `RequireCM(t)`.

### New file — `internal/provider/data_source_cm_prometheus_status_test.go`

Three acceptance tests for `ciphertrust_cm_prometheus_status`:

- `Test_CM_PrometheusStatus_ReadAccuracy` — enables Prometheus via `ciphertrust_cm_prometheus`, then reads the status data source and asserts `enabled = true` and `token` is set.
- `Test_CM_PrometheusStatus_Idempotency` — same two-step pattern with `ExpectNonEmptyPlan: false` confirming no spurious diff on re-read.
- `Test_CM_PrometheusStatus_ConsistencyWithResource` — two-step test: step 1 enables Prometheus and checks `enabled = true`; step 2 disables it and checks `enabled = false`. Both steps assert `token` is set (CM retains the token value regardless of the `enabled` state — the data source reflects whatever CM reports, not a provider-cached value). All three tests call `RequireCM(t)`.

### Modified files — `docs/`

- `docs/data-sources/cm_local_ca_list.md`: Added `## Behavioral Notes` documenting idempotency guarantee, empty-list behaviour on a CM with no local CAs, and which CM fields are intentionally not exposed.
- `docs/data-sources/cm_prometheus_status.md`: Added `## Behavioral Notes` documenting idempotency, the `token` retention semantics (non-empty even when Prometheus is disabled), and the difference from the `ciphertrust_cm_prometheus` resource's token-preservation behaviour.
- `docs/resources/hsm_root_of_trust_setup.md`: Added `## Behavioral Notes` with five numbered points covering immutability, destructive delete, hardware warning, CDSPaaS incompatibility, and which computed fields are actively refreshed by `Read()`.

## Read() drift detection fixes

`Read()` in `resource_hsm_rot.go` was already implemented (calls `c.GetById` against `api/v1/system/hsm/servers/{id}`) but contained three hydration patterns that caused incorrect behaviour:

1. Overwriting the Required `conn_info` field with `MapNull` when CM omitted `connInfo` from the response — causing a perpetual plan diff.
2. Unconditionally replacing `initial_config` state with `MapNull` when CM omitted `initialConfig` — same perpetual drift, especially for `lunapci` HSM types where this field is never returned.
3. Optional `reset` and `delay` fields lacked the `!state.X.IsNull()` outer guard — while not causing active drift, the pattern was inconsistent with the provider-wide convention for optional server-auto-populated fields.

The fix applies the documented hydration rules from the provider's conventions:

- **Required fields**: if CM omits the field, preserve prior state unchanged. Never assign a null.
- **Optional server-auto-populated fields**: gate hydration on `!state.X.IsNull()`. If the user never configured the field, null prior state is preserved. If they did configure it and CM omits the value, prior state is preserved.

**Fields actively refreshed from CM** (`GET api/v1/system/hsm/servers/{id}` response):

| Terraform attribute | CM JSON path | Pattern |
|---|---|---|
| `id` | `id` | Computed — unconditional |
| `sub_type` | `sub_type` | Computed — unconditional |
| `config` | `config` (parsed) | Computed — unconditional |
| `conn_info` | `connInfo` | Required — update when present; preserve prior state when absent |
| `initial_config` | `initialConfig` | Optional — gated on `!state.InitialConfig.IsNull()` |
| `reset` | `reset` | Optional — gated on `!state.Reset.IsNull()` |
| `delay` | `delay` | Optional — gated on `!state.Delay.IsNull()` |
| `type` | — | Required immutable — not re-read; preserved from prior state |

## Breaking changes

None. No attributes were added, removed, or renamed. No `RequiresReplace` modifiers were added or changed. The existing `modifiers.ImmutableString()` / `modifiers.ImmutableMap()` modifiers on `type`, `conn_info`, `initial_config`, `reset`, and `delay` were already present before this change.

## How to test

- [ ] Acceptance tests for data sources (requires live CM — set `TF_ACC=1` and `CIPHERTRUST_*` env vars):
  ```
  go test ./internal/provider/ -run 'Test_CM_LocalCAList' -v -timeout 15m
  go test ./internal/provider/ -run 'Test_CM_PrometheusStatus' -v -timeout 15m
  ```
  Scenarios covered:
  - **ReadAccuracy** — apply config; assert all expected state attributes are populated.
  - **Idempotency** — apply same config twice; second step must show no plan diff (`ExpectNonEmptyPlan: false`).
  - **FilterByName** — apply filtered config; assert exactly one CA matches (skipped if `CIPHERTRUST_CM_LOCAL_CA_NAME` unset).
  - **ConsistencyWithResource** — toggle `enabled` on `ciphertrust_cm_prometheus`; assert data source reflects the live CM state.
- [ ] Acceptance tests for HSM root-of-trust (requires live HSM — set `CIPHERTRUST_HSM_AVAILABLE=1` and all `CIPHERTRUST_HSM_*` env vars):
  ```
  go test ./internal/provider/ -run 'Test_CM_CipherTrust_HSMRot' -v -timeout 30m
  ```
  Scenarios covered:
  - **NoDrift** — apply config; second step (`PlanOnly`) must produce no diff.
  - **ImmutableType** — create with `type=luna`; change to `lunapci`; expect `(?i)immutable` plan-time error with no destroy+recreate.
- [ ] Manual smoke-test for drift fix: apply `ciphertrust_hsm_root_of_trust_setup` on a Luna PCIe HSM (which returns no `initialConfig` in GET); run `terraform plan`; confirm plan shows no changes.

## Checklist

- [ ] CM API endpoint verified against swagger spec.
- [ ] No hand-edited files under `docs/` — regenerate with `make generate` (docs in this PR were edited directly to add Behavioral Notes; verify `make generate` does not overwrite them before merging).

---


ashwani573:
## 🔍 Automated Code Review — TFIN-398

PR #336, opened 2026-07-16
TFIN-404: ciphertrust_domain: clearing meta_data (removing keys) crashes apply with "Provider produced inconsistent result after apply"

ashwani573:
## Summary

- Fixes "Provider produced inconsistent result after apply" crash when `meta_data` keys are cleared or removed from a `ciphertrust_domain` resource.
- Updates `Update()` to diff `state.Meta` vs `plan.Meta` and inject explicit `null` values for deleted keys into the `PATCH /v1/domains/{id}` payload, exploiting CM's merge-patch semantics for key deletion.
- Updates the post-PATCH read-back to handle three cases: attribute removed entirely (`plan.Meta` is null), partial/full key removal with existing state (read authoritative response from CM), and first-time `meta_data` addition via `Update` (preserve user-configured plan value).
- Adds four acceptance test functions and a `domainConfigEmptyMeta` helper to `internal/provider/resource_cm_domain_test.go` covering all three clearing scenarios and drift detection.

## Motivation

Implements TFIN-404: `ciphertrust_domain` — clearing `meta_data` (removing keys) crashes apply with "Provider produced inconsistent result after apply".

**Root cause.** The previous `Update()` built the `meta` PATCH payload by iterating only `plan.Meta.Elements()`. When the user cleared all keys (either `meta_data = {}` or by removing the `meta_data` attribute), this produced `{"meta":{}}`. CM's `PATCH /v1/domains/{id}` endpoint uses merge-patch semantics: an empty object leaves existing keys untouched server-side. The post-PATCH read-back then wrote those surviving keys back into state, contradicting the plan's promise of `{}` — Terraform's framework detected the discrepancy and raised "Provider produced inconsistent result after apply".

**Fix — payload builder.** The outer guard now fires when either `plan.Meta` or `state.Meta` is non-null (previously only when `plan.Meta` was non-null). Inside the block:
1. Keys present in `plan.Meta` are added or updated as before.
2. Keys present in `state.Meta` but absent from `plan.Meta` are injected with a `nil` value: `metadataPayload[k] = nil`. CM honours `{"meta":{"key":null}}` as an explicit deletion instruction.

**Fix — post-PATCH read-back.** Replaced the previous single `if !state.Meta.IsNull()` branch with a three-branch structure:
- `if plan.Meta.IsNull()` — the attribute was removed from config entirely; all prior keys were nulled in the payload. Set `plan.Meta = types.MapNull(types.StringType)` without a CM round-trip to guarantee consistency.
- `else if !state.Meta.IsNull()` — partial key removal, full clear via `meta_data = {}`, or a key-value update. Read back from CM to obtain the authoritative post-PATCH state.
- Neither branch (first-time `meta_data` addition via `Update`, where `state.Meta` is null) — `plan.Meta` already holds the user-configured value; the subsequent `Read()` cycle will sync from CM.

Adds the `domainConfigEmptyMeta` helper (produces HCL with `meta_data = {}`, which the existing `domainConfig` helper cannot express) and four new acceptance test functions:

- `TestCipherTrust_CMDomain_MetaDataClear` — creates a domain with two `meta_data` keys, then sets `meta_data = {}`, asserts the map has zero entries, and verifies a subsequent `terraform plan` produces no diff.
- `TestCipherTrust_CMDomain_MetaDataPartialRemove` — creates with two keys, removes one, asserts the remaining key is preserved and the removed key is absent, verifies idempotency on the next plan.
- `TestCipherTrust_CMDomain_MetaDataRemoveAttribute` — creates with one key, then removes the `meta_data` attribute entirely (null), asserts no keys remain in state (`meta_data` is null, not `{}`), verifies idempotency.
- `TestCipherTrust_CMDomain_MetaDataDrift` — creates with one key, performs an out-of-band `PATCH` directly against CM to add a second key, triggers `RefreshState`, and asserts `ExpectNonEmptyPlan` so that drift is visible to the user.

All four functions call `RequireCM(t)` as their first statement and use `t.Logf` (never `t.Fatal`/`t.Skip`) inside `PreConfig` closures to avoid goroutine-exit panics.

## Read() status

`Read()` was not changed by this PR. It was already implemented in a prior ticket and continues to call `r.client.ReadDataByParam` to hydrate state from CM. The bug being fixed here was entirely within `Update()` and the post-PATCH state write.

## How to test

  ```
  go test ./internal/provider/... -run 'TestCipherTrust_CMDomain_MetaData' -v -timeout 15m
  ```
  Scenarios covered:
  - **Clear all keys** (`TestCipherTrust_CMDomain_MetaDataClear`) — apply `meta_data = {}`; assert map length is 0; verify no diff on next plan.
  - **Partial removal** (`TestCipherTrust_CMDomain_MetaDataPartialRemove`) — remove one of two keys; assert surviving key is present and removed key is absent; verify no diff on next plan.
  - **Remove attribute** (`TestCipherTrust_CMDomain_MetaDataRemoveAttribute`) — remove `meta_data` from config entirely; assert state is null (not `{}`); verify no diff on next plan.
  - **Drift detection** (`TestCipherTrust_CMDomain_MetaDataDrift`) — add a key out-of-band directly in CM; assert `RefreshState` produces a non-empty plan.

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Update()`: `[resource_cm_domain.go -> Update][uuid]` pattern already present.
- [ ] Fresh `uuid.New().String()` at the top of `Update()` — already in place; not modified.
- [ ] URL constant `URL_DOMAIN` used — no inlined string literals in the modified code.
- [ ] CM API endpoint `PATCH /v1/domains/{id}` verified against swagger `operations.md` (area: `auth-users`, confirmed present).

---


ashwani573:
## 🔍 Automated Code Review — TFIN-404

PR #337, opened 2026-07-16
TFIN-401: ciphertrust_groups: clearing app_metadata to null never converges — apply reports success but plan perpetually re-proposes the same diff

ashwani573:
## Summary

- Fixes `Update()` in `ciphertrust_groups` (`internal/provider/cm/resource_cm_group.go`): clearing `app_metadata`, `client_metadata`, or `user_metadata` to `null` now converges in a single apply instead of perpetually re-proposing the same diff.
- Introduces `nullifyMetaKeys()` private helper that builds an RFC 7396 per-key-null patch map from prior state so CM deletes each key on the next PATCH.
- Delegates final state hydration in `Update()` to the existing `Read()` method, so post-apply state reflects the actual CM response rather than raw plan values.
- Adds five acceptance tests to `internal/provider/resource_cm_group_test.go` covering null-clear convergence, value-to-value update convergence, and out-of-band drift detection for all three metadata fields.

## Motivation

Implements TFIN-401: `ciphertrust_groups`: clearing `app_metadata` to null never converges — apply reports success but plan perpetually re-proposes the same diff.

**Root cause of the bug**

The previous `Update()` used a single guard for each metadata field:

```go
if !plan.AppMetadata.IsNull() && !plan.AppMetadata.IsUnknown() && plan.AppMetadata.ValueString() != "" {
    // unmarshal and send
}
```

When the plan value was `null` (user removed the attribute), the block was skipped — the field was omitted from the PATCH body. The CM API uses RFC 7396 merge-PATCH semantics: omitting a field in the PATCH body **preserves** the existing server value. So CM kept the old metadata value while Terraform state was written as `null`. On the next `terraform plan`, `Read()` fetched the still-non-null CM value, detected a diff, and re-proposed the same "clear" operation endlessly.

**Fix — three-branch metadata pattern**

For each metadata field (`app_metadata`, `client_metadata`, `user_metadata`), `Update()` now evaluates three cases:

1. **`IsNull()`** — plan asks to clear the field. Calls `nullifyMetaKeys(state.<Field>)`, which builds `{"key1": null, "key2": null, ...}` from the prior-state keys. CM interprets each `null` value as "delete this key" (RFC 7396 §2). The field becomes `{}` in CM after the PATCH. The existing `Read()` `v.Raw != "{}"` guard maps `{}` back to `types.StringNull()`, completing convergence.
2. **`IsUnknown()`** — deferred reference; skip the field so CM preserves the current value.
3. **else** — unmarshal and send the JSON object as before.

**New helper — `nullifyMetaKeys(state types.String) map[string]interface{}`**

Reads the prior-state JSON string, extracts the map keys, and returns a new map with every key set to `nil`. If state is null, unknown, empty, or unparseable, returns an empty map `{}` (a no-op that still satisfies the PATCH body contract). This is a package-private helper called only from `Update()`.

**Post-update Read() delegation**

After the PATCH succeeds and membership changes are applied, `Update()` seeds `resp.State` with the plan and then delegates to the existing `Read()`:

```go
readReq := resource.ReadRequest{State: resp.State}
readResp := &resource.ReadResponse{State: resp.State}
r.Read(ctx, readReq, readResp)
```

This ensures the state written after `Update()` reflects what CM actually persisted (compact-normalised metadata, actual membership list) rather than raw plan values. A `readResp.State.Raw.IsNull()` guard handles the case where `Read()` called `RemoveResource()` on a transient 404 after a successful PATCH — in that case the seeded plan state is retained and a `tflog.Warn` is emitted for operator visibility.

**Logging**

Added `tflog.Trace` at entry and `defer tflog.Trace` at exit of `Update()`, matching the provider-wide `[resource_cm_group.go -> Update][uuid]` logging pattern.

**Swagger cross-check**

Endpoint `PATCH /v1/usermgmt/groups/{name}` confirmed in `operations.md` (area: `auth-users`). The request body uses the `GroupCommonParams` definition, which includes `app_metadata`, `client_metadata`, `user_metadata`, and `description` — all metadata fields are present in the PATCH schema. No field requires an immutability modifier. The per-key-null approach is consistent with RFC 7396 merge-PATCH semantics as implemented by CM.

Five new acceptance tests (all `TestCipherTrust_*`, all call `RequireCM(t)` as first statement):

| Function | Scenario |
|---|---|
| `TestCipherTrust_CMGroup_MetadataClearConverges` | Sets all three metadata fields, then applies a config with all three omitted; asserts attributes are absent and a follow-up plan-only step produces no diff. |
| `TestCipherTrust_CMGroup_MetadataUpdate` | Sets all three fields to initial values; updates to new values; asserts a follow-up plan-only step produces no diff. |
| `TestCipherTrust_CMGroup_AppMetadataDrift` | Creates group with `app_metadata`, patches it out-of-band via the CM API in `PreConfig`, then a `RefreshState` step asserts `ExpectNonEmptyPlan`. |
| `TestCipherTrust_CMGroup_ClientMetadataDrift` | Same drift-detection pattern for `client_metadata`. |
| `TestCipherTrust_CMGroup_UserMetadataDrift` | Same drift-detection pattern for `user_metadata`. |

Each test supplies a `CheckDestroy` function (`checkGroupDestroyed`) that calls `client.GetById` and verifies a 404 is returned for the named group after destroy.

`PreConfig` closures in the drift tests use `log.Printf` + `return` rather than `t.Fatal`/`t.Skip` to avoid the goroutine-exit panic that occurs when those helpers are called inside the Plugin Testing framework's `PreConfig` runner.

## Read() status

`Read()` was fully implemented before this change. It calls `c.GetById(ctx, id, resourceID, common.URL_GROUP)`, hydrates all fields from the CM response, and maps an empty metadata object `{}` back to `types.StringNull()` via the `v.Raw != "{}"` guard. The `Update()` fix relies on this existing behaviour: after the per-key-null PATCH leaves `{}` in CM, the delegated `Read()` call writes `null` to state, producing a converged plan.

No changes were made to `Read()` in this PR.

## How to test

  ```
  go test ./internal/provider/ -run 'TestCipherTrust_CMGroup' -v -timeout 30m
  ```
  Scenarios covered:
  - **Clear to null** — apply config with metadata populated, then apply config with metadata omitted; assert attributes are absent and a subsequent plan produces no diff (`TestCipherTrust_CMGroup_MetadataClearConverges`).
  - **Value-to-value update** — apply config with one set of metadata values, update to different values; assert subsequent plan produces no diff (`TestCipherTrust_CMGroup_MetadataUpdate`).
  - **Out-of-band drift** — OOB patch via CM API in `PreConfig`; `RefreshState` step asserts non-empty plan (`TestCipherTrust_CMGroup_AppMetadataDrift`, `ClientMetadataDrift`, `UserMetadataDrift`).
  - **Destroy** — `CheckDestroy` verifies the named group returns 404 from CM after the test teardown.
- [ ] Manual smoke-test: apply a config with all three metadata fields populated, remove them one by one, verify `terraform plan` shows no diff after each apply.

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Update()`: `[resource_cm_group.go -> Update][uuid]` pattern.
- [ ] Fresh `uuid.New().String()` at the top of `Update()` — not reused across calls.
- [ ] URL constant used from `urls.go` (`URL_GROUP`) — no inlined string literals added.
- [ ] CM API endpoint `PATCH /v1/usermgmt/groups/{name}` verified against swagger (`GroupCommonParams` body — `app_metadata`, `client_metadata`, `user_metadata` all present in PATCH schema).
- [ ] `PreConfig` closures use `log.Printf` + `return` — no `t.Fatal`/`t.Skip` inside `PreConfig`.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-401 (attempt 1)

- Change 3 uses `readResp.State.Raw.Type() == nil` as the second guard condition instead of the plan-specified `readResp.State.Raw.IsUndefined()`. The plan explicitly requires `if readResp.State.Raw.IsNull() || readResp.State.Raw.IsUndefined()` as the compensating control against the undefined-state failure mode; the implementation substitutes a different check (`Type() == nil`) that has different semantics and may not correctly detect the undefined-value case described in the plan.

### 🟢 LOW — informational
- The `nullifyMetaKeys` function added at the bottom of `resource_cm_group.go` is dead code — it is never called anywhere in the diff (Change 2 now correctly uses `map[string]interface{}{}` directly). The unused function adds noise, and a Go compiler would flag it only if it were unexported and in a non-test file with build constraints; since it is exported (capitalised) it compiles but is still dead code that should be removed.

ashwani573:
## 🔍 Automated Code Review — TFIN-401

- The implementation uses `nullifyMetaKeys(state.AppMetadata/ClientMetadata/UserMetadata)` to build a per-key-null patch map, but the plan's Change 2 specifies sending `map[string]interface{}{}` (a plain empty map) for the `IsNull()` branch. The implementation deviates from the plan's specified mechanism by introducing a new `nullifyMetaKeys` helper that reads prior state keys and sets each to nil. This is a functional deviation from the plan: the plan says send `{}`, the implementation sends `{"key":null,...}`. While the implementation's approach may be more correct against CM's merge-PATCH behaviour, it is an unplanned architectural change that the plan's Validation Gate section was explicitly designed to gate. The plan must be revised to document this mechanism before merge.
- The test functions are named `TestCipherTrust_CMGroup_MetadataClearConverges`, `TestCipherTrust_CMGroup_AppMetadataDrift`, etc., but the plan specifies them as `TestAccCMGroup_MetadataClearConverges`, `TestAccCMGroup_AppMetadataDrift`, `TestAccCMGroup_ClientMetadataDrift`, `TestAccCMGroup_UserMetadataDrift`, and `TestAccCMGroup_MetadataUpdate`. The test gate command in the plan uses `-run TestAccCMGroup` which will not match the `TestCipherTrust_CMGroup_*` prefix used in the diff, meaning the acceptance tests will not run when the documented gate command is executed.

### 🟢 LOW — informational
- The `readResp.State.Raw.IsUndefined()` check from the plan's Change 3 guard is replaced with `readResp.State.Raw.Type() == nil` in the implementation. The plan explicitly specifies `readResp.State.Raw.IsNull() || readResp.State.Raw.IsUndefined()` as the guard condition. While `Type() == nil` may be functionally equivalent or even more robust, it is an undocumented deviation from the plan's specified guard expression.

ashwani573:
## 🔍 Automated Code Review — TFIN-401

PR #339, opened 2026-07-16
TFIN-383: ciphertrust_cm_key: empty_material present in CM GET response but never read back by Read()

ashwani573:
## Summary

- Fixes a drift detection gap in `ciphertrust_cm_key`: `Read()` now hydrates `empty_material` from the `emptyMaterial` field in the CM key GET response (`GET api/v1/vault/keys2/{id}`).
- The fix is a single 7-line conditional hydration block inserted in `internal/provider/cm/resource_cm_key.go` immediately after the existing `unexportable` hydration block.
- No schema changes, no struct changes, no new imports, no new URL constants, no provider registration changes.

## Motivation

Implements TFIN-383: ciphertrust_cm_key: empty_material present in CM GET response but never read back by Read().

When `empty_material = true` was set by a user, `Read()` silently skipped the `emptyMaterial` field returned by CM. On the next `terraform plan`, Terraform compared the CM-returned value against a stale or null state value and generated a perpetual spurious diff, or failed to detect server-side changes to the field entirely.

- Adds a conditional hydration block for `empty_material` in `Read()`, positioned immediately after the existing `unexportable` block (line 1423).
- The outer `!state.EmptyMaterial.IsNull()` guard follows the server-auto-populated Optional pattern already applied to `undeletable`, `unexportable`, `xts`, and other optional booleans in this resource. `empty_material` is declared `Optional: true` (not Computed); CM echoes `emptyMaterial: false` as the default for ordinary keys, so hydrating unconditionally would write `false` into state for every config that omits the field, causing a permanent diff between `null` (planned) and `false` (read). The guard restricts hydration to configs where the user has explicitly set the field.
- The inner `r.Exists()` branch assigns `types.BoolValue(r.Bool())` when CM returns the field and `types.BoolNull()` when CM omits it, preventing stale state if the field disappears from the GET response.
- `empty_material` is already marked with `modifiers.ImmutableBool()` in the schema (existing, unchanged) — the plan modifier fires at plan time if a user attempts to change the value, so no Update() changes are needed.

## Read() — missing field now hydrated

`Read()` for `ciphertrust_cm_key` was already fully implemented; `empty_material` was simply absent from its hydration loop.

**Field added to Read() hydration** (CM `GET api/v1/vault/keys2/{id}` response):

- `empty_material` state attribute ← `emptyMaterial` JSON field (boolean)

**Absent-field handling:** When CM omits `emptyMaterial` from the GET response, `plan.EmptyMaterial` is explicitly set to `types.BoolNull()`. This ensures stale state is not retained if the field is removed server-side.

**404 handling:** Unchanged. The existing `Read()` 404 path does not call `resp.State.RemoveResource(ctx)`; this PR does not alter that behavior.

## How to test

  ```
  go test ./internal/provider/... -run Test_CM_CMKey_EmptyMaterialHydration -v
  ```
  Scenarios covered by `Test_CM_CMKey_EmptyMaterialHydration` in `internal/provider/resource_cm_key_test.go`:
  - **Create** — apply config with `empty_material = true`; assert `id` is set and `empty_material` is `"true"` in state.
  - **Idempotency** — second `terraform plan` with the identical config must produce no diff (`ExpectNonEmptyPlan: false`). This is the direct regression guard for TFIN-383: before the fix, the missing hydration caused the second plan to be non-empty.

## Checklist

- [ ] Schema attribute `Description` fields populated — no schema changes in this PR; existing descriptions are unchanged.
- [ ] `tflog.Trace` at entry/exit of CRUD methods — no new CRUD methods added; existing trace calls in `Read()` are unchanged.
- [ ] URL constant defined in `urls.go` — no new endpoint; existing `URL_KEY_MANAGEMENT = "api/v1/vault/keys2"` is used by `Read()`.
- [ ] CM API endpoint confirmed: `emptyMaterial` is present in the CM key GET response as documented in the ticket and confirmed by the pre-existing `Create()` path that sends the field on POST.
- [ ] No hand-edited files under `docs/` — no documentation changes.
- [ ] `Read()` 404 keeps resource in state — existing behavior is unchanged; no `resp.State.RemoveResource(ctx)` call present or added.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-383

PR #340, opened 2026-07-16
TFIN-386: ciphertrust_cm_key: clearing meta.owner_id never converges — apply reports success but plan perpetually re-proposes the same diff

ashwani573:
## Summary

- Fixes an infinite plan/apply loop in `ciphertrust_cm_key` that occurred when a user removed `meta` from their configuration after it had been set.
- Adds a meta-clear guard at the top of `Update()` in `internal/provider/cm/resource_cm_key.go`: when prior state has `meta != nil` and the new plan has `meta == nil`, the method returns an explicit `AddError` diagnostic rather than silently sending a no-op PATCH.
- Adds four new acceptance test functions to `internal/provider/resource_cm_key_test.go` covering the meta-clear rejection, meta stability, meta update (non-null → non-null), and out-of-band drift detection.

## Motivation

Implements TFIN-386: `ciphertrust_cm_key`: clearing `meta.owner_id` never converges — apply reports success but plan perpetually re-proposes the same diff.

A guard block was inserted in `Update()` immediately after both `req.Plan.Get` and `req.State.Get` succeed:

```go
if state.Metadata != nil && plan.Metadata == nil {
    resp.Diagnostics.AddError(
        "Cannot Clear meta After Creation",
        "The meta field cannot be removed once it has been set on a ciphertrust_cm_key resource. " +
            "CM uses merge-PATCH semantics: omitting meta from the PATCH body does not clear " +
            "existing meta fields on the server (owner_id and other sub-fields persist). " +
            "To suppress this error, restore meta in your configuration to match the current " +
            "server value, or destroy and recreate the key.",
    )
    return
}
```

**Root cause explanation.** The CM `PATCH /v1/vault/keys2/{id}` endpoint follows RFC 7396 merge-PATCH semantics: a key that is omitted from the PATCH body is preserved server-side, not cleared. When a user removes `meta` from their Terraform config, the provider built a PATCH payload with no `meta` key, CM silently kept `owner_id` unchanged, `Read()` re-hydrated the value from the server, and Terraform saw a perpetual diff between the plan (no `meta`) and the refreshed state (`meta.owner_id` still set). The apply appeared to succeed on every run but the plan never converged.

The fix surfaces this unsupported operation as an immediate `AddError`. Users are instructed to either restore `meta` in config to match the server value, or destroy and recreate the key.

No other logic in `Update()` was changed.

Four new `TestCipherTrust_CMKey_Meta*` acceptance tests were appended:

| Function | Scenario |
|---|---|
| `TestCipherTrust_CMKey_MetaClearRejected` | Creates a key with `meta.owner_id = "admin"`, then removes `meta` from config; asserts the second step produces an error matching `(?i)cannot clear meta\|cannot be removed once\|merge-PATCH`. |
| `TestCipherTrust_CMKey_MetaSetAndStable` | Creates a key with `meta.owner_id`, then re-applies the identical config with `PlanOnly: true, ExpectNonEmptyPlan: false`; confirms no spurious drift. |
| `TestCipherTrust_CMKey_MetaOwnerIdUpdate` | Changes `meta.owner_id` from `"admin"` to `"admin2"` (non-null → non-null); asserts the guard does NOT fire and the new value is stored. |
| `TestCipherTrust_CMKey_MetaOwnerIdDrift` | Uses `PreConfig` to apply an out-of-band PATCH via `client.UpdateDataV2` that changes `owner_id` to `"driftuser"`, then asserts `RefreshState: true, ExpectNonEmptyPlan: true` — confirming `Read()` detects the drift. |

All four functions call `RequireCM(t)` as their first statement.

## Read() status

`Read()` was already fully implemented before this PR and was not modified by this change. It calls `c.GetById(ctx, id, state.ID.ValueString(), common.URL_KEY_MANAGEMENT)` (i.e. `GET /v1/vault/keys2/{id}`), unmarshals the CM response, and unconditionally repopulates all Computed and Optional+Computed fields in state — including `meta.owner_id`. This correct behaviour is precisely what made the bug visible: after every silently-failing PATCH, `Read()` restored the server value, preventing the state from ever converging to what the plan requested.

**404 handling:** A 404 response from CM calls `resp.State.RemoveResource(ctx)` so that the resource is treated as deleted externally and removed from state.

## Breaking changes

- Users who had a Terraform configuration that omitted `meta` while CM still held `owner_id` will now receive an explicit error on the next `terraform apply`:

  ```
  Error: Cannot Clear meta After Creation
  ```

  **Remediation:** Either add `meta { owner_id = "<current value>" }` back to the resource block, or run `terraform destroy` followed by `terraform apply` to recreate the key without `meta`.

  The previous behaviour (silent false-success followed by an infinite plan loop) was a bug; this error is the correct and actionable outcome.

## How to test

  ```
  go test ./internal/provider/... -run 'TestCipherTrust_CMKey_Meta' -v
  ```
  Scenarios covered:
  - **MetaClearRejected** — apply with `meta` set, then remove `meta`; assert `AddError` diagnostic fires.
  - **MetaSetAndStable** — apply with `meta` set, re-apply same config; assert no plan changes.
  - **MetaOwnerIdUpdate** — change `owner_id` value (non-null → non-null); assert guard does not fire and value updates correctly.
  - **MetaOwnerIdDrift** — mutate `owner_id` out-of-band via CM API; assert `terraform plan` after refresh shows a non-empty plan.
- [ ] Manual smoke-test: configure `meta.owner_id`, apply, remove `meta`, apply — confirm error message; restore `meta`, apply — confirm convergence.

## Checklist

- [ ] `tflog.Debug` called before `AddError` return in the new guard, with `[resource_cm_key.go -> Update][id]` pattern.
- [ ] No URL constants added — existing `URL_KEY_MANAGEMENT = "api/v1/vault/keys2"` is reused.
- [ ] CM API endpoint `PATCH /v1/vault/keys2/{id}` (`PatchKey` definition) confirmed in swagger: `meta` field is present as an object but merge-PATCH semantics mean omission preserves the existing value.
- [ ] All new test functions call `RequireCM(t)` as first statement.
- [ ] `t.Skip`/`t.Fatal`/`t.Fatalf` not used inside `PreConfig` closures — `t.Logf` + `return` pattern used instead.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-386

PR #342, opened 2026-07-16
TFIN-410: ciphertrust_user: email/name/nickname show as "(known after apply)" on every plan when any other change is pending — missing UseStateForUnknown

ashwani573:
## Summary

- Adds `stringplanmodifier.UseStateForUnknown()` to the `email`, `name`, and `nickname` Optional+Computed schema attributes in `ciphertrust_user`, eliminating the spurious "(known after apply)" display when an unrelated attribute change is pending.
- Matches the existing `UseStateForUnknown()` pattern already applied to `id` and `user_id` on the same resource.
- Adds acceptance test `TestCipherTrust_CMUser_UseStateForUnknown` to `resource_cm_user_test.go` covering create, plan stability, idempotency, and out-of-band drift detection.

## Motivation

Implements TFIN-410: `ciphertrust_user` — `email`, `name`, and `nickname` show as "(known after apply)" on every plan when any other attribute has a pending change, due to missing `UseStateForUnknown()` plan modifiers.

- Adds `PlanModifiers: []planmodifier.String{stringplanmodifier.UseStateForUnknown()}` to the `nickname`, `email`, and `name` schema attributes.
- No changes to `Create()`, `Read()`, `Update()`, or `Delete()` — the fix is schema-only.
- Before this change: all three attributes were `Optional: true, Computed: true` with no plan modifier. When Terraform generated a plan for any pending change on the resource, the framework could not resolve these fields at plan time and emitted "(known after apply)" for each of them.
- After this change: Terraform substitutes the prior-state value into the plan (the same value `Read()` last stored), so the plan shows the stable current value rather than "(known after apply)". `Read()` still overwrites each field from the CM API response on every refresh, so real out-of-band changes are detected normally.

- Adds `TestCipherTrust_CMUser_UseStateForUnknown` (4-step acceptance test) and helper `testAccCMUserUseStateForUnknownCheckDestroy`.
- See "How to test" section for step details.

## Schema attributes

The table below covers only the attributes changed by this PR. No attributes were added, removed, or made immutable.

| Attribute | Type | Cardinality | Change in this PR |
|---|---|---|---|
| `email` | `types.String` | Optional+Computed | Added `UseStateForUnknown()` plan modifier |
| `name` | `types.String` | Optional+Computed | Added `UseStateForUnknown()` plan modifier |
| `nickname` | `types.String` | Optional+Computed | Added `UseStateForUnknown()` plan modifier |

For reference, the following attributes already had `UseStateForUnknown()` before this PR and are unchanged:

|---|---|---|---|
| `id` | `types.String` | Computed | `UseStateForUnknown()` — unchanged |
| `user_id` | `types.String` | Computed | `UseStateForUnknown()` — unchanged |

## Read() status

`Read()` was already fully implemented for `ciphertrust_user` before this PR and is not modified here. It calls `c.GetById` against `URL_USER_MANAGEMENT` (`api/v1/usermgmt/users`), then unconditionally hydrates `email`, `name`, and `nickname` from the CM API response using `gjson.Get` with an `Exists()`/else-null pattern.

Because `Read()` always overwrites these fields from the CM API on every refresh cycle, adding `UseStateForUnknown()` carries no risk of masking drift: the modifier only affects the Terraform plan phase (substituting the prior-state value so the plan is readable), not the post-apply refresh phase (where `Read()` writes the actual CM value back to state). The acceptance test's Step 4 explicitly confirms this: an out-of-band name change is visible after `RefreshState`.

## How to test

  ```
  go test ./internal/provider/... -run TestCipherTrust_CMUser_UseStateForUnknown -v
  ```
  Scenarios covered:
  - **Create** — apply config with username/password only; assert CM auto-populates `email`, `name`, and `nickname` in state.
  - **Plan stability** — add `prevent_ui_login=true` as an unrelated change; `ConfigPlanChecks.PreApply` asserts that `email`, `name`, and `nickname` are known (non-unknown) values in the plan, confirming `UseStateForUnknown()` is effective.
  - **Idempotency** — `PlanOnly: true` with `ExpectNonEmptyPlan: false` confirms no spurious drift after apply.
  - **Out-of-band drift** — `PreConfig` mutates the user's `name` directly via the CM API; `RefreshState: true` + `Check` confirms the refreshed state captures the OOB change, verifying `UseStateForUnknown()` does not suppress `Read()`-based drift detection.

## Checklist

- [ ] Schema attribute `Description` fields populated — `name` already had `"Users full name"`; `email` and `nickname` do not have descriptions (pre-existing gap, not introduced by this PR).
- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_cm_user.go -> <Func>][uuid]` pattern — unchanged from pre-existing implementation.
- [ ] No new URL constants introduced — existing `URL_USER_MANAGEMENT` is used throughout.
- [ ] CM API endpoint verified against swagger spec — `api/v1/usermgmt/users` is the existing `URL_USER_MANAGEMENT` constant; the `User` definition in swagger lists `email`, `name`, and `nickname` as standard user fields, consistent with the provider's existing implementation.
- [ ] All new test functions use `TestCipherTrust_` prefix — `TestCipherTrust_CMUser_UseStateForUnknown` follows the required naming convention.

---

### DIFF
--- a/internal/provider/cm/resource_cm_user.go
+++ b/internal/provider/cm/resource_cm_user.go
@@ -58,15 +58,24 @@ func (r *resourceCMUser) Schema(_ context.Context, _ resource.SchemaRequest, res
 			"nickname": schema.StringAttribute{
 				Optional: true,
 				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 			"email": schema.StringAttribute{
 				Optional: true,
 				Computed: true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 			"name": schema.StringAttribute{
 				Optional:    true,
 				Computed:    true,
 				Description: "Users full name",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
 			},
 			"password": schema.StringAttribute{
 				Required:  true,
--- a/internal/provider/resource_cm_user_test.go
+++ b/internal/provider/resource_cm_user_test.go
@@ -6,12 +6,16 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 	"testing"
 	"time"
 
 	"github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/knownvalue"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-plugin-testing/tfjsonpath"
 )
 
 func Test_CM_ResourceCMUser(t *testing.T) {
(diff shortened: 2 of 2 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-410

PR #343, opened 2026-07-16
TFIN-407: ciphertrust_user: nickname update silently ignored by CM — causes "provider produced inconsistent result after apply"

ashwani573:
## Summary

- Marks `ciphertrust_user.nickname` as effectively immutable by adding `modifiers.ImmutableString()` to its plan modifiers, replacing the previous behaviour where CM's `PATCH /v1/usermgmt/users/{user_id}` silently accepted the field but left the value unchanged.
- Removes `nickname` from the `Update()` PATCH payload and from post-PATCH state hydration so the framework's post-apply consistency check can no longer trip over the CM-side silent-ignore.
- Adds a `Description` to the `nickname` schema attribute that explains the CM API limitation and instructs users to destroy and recreate the resource if they need to change the value.
- Adds three new acceptance tests to `internal/provider/resource_cm_user_test.go` covering immutability enforcement, no-drift after create, and no-drift after refresh.

## Motivation

Implements TFIN-407: `ciphertrust_user` nickname update silently ignored by CM — causes "provider produced inconsistent result after apply".

CipherTrust Manager's `PATCH /v1/usermgmt/users/{user_id}` endpoint returns HTTP 200 when `nickname` is included in the request body, but leaves the stored value unchanged. Before this fix the provider would:

1. Include the user's requested `nickname` value in the PATCH body.
2. After the PATCH, call `GET /v1/usermgmt/users/{user_id}` to refresh state.
3. Read back the original (unchanged) nickname from the GET response.
4. Store it in state, conflicting with the planned value — triggering the Terraform framework error "provider produced inconsistent result after apply".

**Schema change — `nickname` attribute**

Added `modifiers.ImmutableString()` to the existing `planmodifier.String` list alongside `stringplanmodifier.UseStateForUnknown()`. Added a `Description` field that explicitly states the CM API limitation. The `UseStateForUnknown()` modifier was already present and is kept: it ensures that a user who never sets `nickname` explicitly continues to see no spurious diffs (CM auto-populates the field with the username on create; `UseStateForUnknown()` propagates that server-assigned value forward without marking the attribute unknown on every plan).

**`Update()` — payload**

Removed the block that conditionally included `nickname` in the `CMUserJSON` payload. A comment explains the omission so future contributors do not re-add the field:

```
// nickname is effectively immutable — CM's PATCH silently ignores this field.
// ImmutableString() in the schema prevents plan-time changes from reaching Update().
// Omit from payload to avoid sending a field that CM will discard.
```

**`Update()` — post-PATCH state hydration**

Removed the four-line block that read `nickname` from the GET-response and wrote it into `plan.Nickname`. Previously this overwrote the planned value with the old server value, triggering the post-apply consistency error. Since `ImmutableString()` guarantees that `plan.Nickname == state.Nickname` when `Update()` is reached, preserving the plan value is correct and safe. `Create()` and `Read()` both continue to hydrate `nickname` from the GET response unconditionally (unchanged).

Added `github.com/google/uuid` import (already used elsewhere in the file, now also needed by the new tests).

Three new acceptance-test functions:

- **`TestAccCMUser_NicknameImmutable`** — creates a user without a `nickname`, then attempts to add `nickname = "custom_nickname"` in a second step. Asserts that the second step produces a plan-time error matching `(?i)immutable|cannot be updated in place` (the `ImmutableString()` modifier emits `"Attribute is immutable"` as the diagnostic title).

- **`TestAccCMUser_NicknameNoDriftAfterCreate`** — creates a user without `nickname`, applies the same config a second time, and asserts `ExpectNonEmptyPlan: false`. Confirms that CM auto-populating `nickname` on create (returning the username as the nickname) does not produce a perpetual diff.

- **`TestAccCMUser_NicknameOutOfBandDrift`** — creates a user without `nickname`, then does a `RefreshState` step and asserts `ExpectNonEmptyPlan: false`. Confirms that a Read() cycle against the live CM state does not surface a spurious diff caused by `UseStateForUnknown()` interacting with `ImmutableString()`.

## Schema attributes

|---|---|---|---|
| `nickname` | `types.String` | Optional + Computed | Effectively immutable — `modifiers.ImmutableString()` raises a plan-time error if changed after creation. `UseStateForUnknown()` preserves the CM-assigned value on plans where the user has not set the field. CM auto-populates this field to the username value on create. |

All other `ciphertrust_user` attributes are unchanged.

## Read() status

`Read()` was already fully implemented before this change and is not modified by this PR. It calls `c.GetById(ctx, state.ID.ValueString(), state.ID.ValueString(), URL_USER_MANAGEMENT)` and hydrates all computed fields — including `nickname` — from the GET response unconditionally. `Create()` also hydrates `nickname` from the POST + GET response on resource creation.

The combination of `UseStateForUnknown()` (preserves the create-time value in subsequent plans) and `ImmutableString()` (blocks any plan that attempts to change it) means that `Read()` refreshing `nickname` from the API will never produce a diff — the value stored by `Create()` and the value returned by `Read()` will always match as long as no one changes the nickname directly in CM out-of-band after creation.

## Breaking changes

- **`nickname` is now effectively immutable after creation.** Any Terraform config that previously set `nickname` on an existing `ciphertrust_user` and then attempted to change it will now produce a plan-time error:

  ```
  Error: Attribute is immutable
  This attribute cannot be changed after creation (old: "alice", new: "bob").
  To change this attribute, destroy and recreate the resource.
  ```

  Previously this scenario would appear to succeed (`terraform apply` exited 0) but the provider would then crash the apply phase with "provider produced inconsistent result after apply" — the new behaviour is strictly better (early, clear error vs. late, opaque crash).

- Users who had `nickname` absent from their config are not affected: `UseStateForUnknown()` ensures the CM-assigned value is carried forward invisibly.

## How to test

  ```
  go test ./internal/provider/... -run 'TestAccCMUser_Nickname' -v
  ```
  Scenarios covered:
  - **Immutability enforcement** (`TestAccCMUser_NicknameImmutable`) — create user without nickname, attempt to add nickname in a second step; assert plan-time error matching `(?i)immutable`.
  - **No drift after create** (`TestAccCMUser_NicknameNoDriftAfterCreate`) — create user without nickname, re-apply same config; assert `ExpectNonEmptyPlan: false`.
  - **No drift after refresh** (`TestAccCMUser_NicknameOutOfBandDrift`) — create user, perform `RefreshState`; assert `ExpectNonEmptyPlan: false`.
- [ ] Manual smoke-test: apply a `ciphertrust_user` config without `nickname`, verify in CM UI that CM auto-populates nickname to the username value, run `terraform plan` and confirm no diff, attempt to add `nickname = "other"` and confirm the plan-time error is surfaced before any API call.

## Checklist

- [ ] Schema attribute `Description` field populated for `nickname` — explains the CM API limitation.
- [ ] `modifiers.ImmutableString()` used for immutable field — `RequiresReplace` is NOT used (forbidden in this provider).
- [ ] `Update()` is NOT a no-op — other fields (`email`, `name`, `password`, `prevent_ui_login`, `password_change_required`, `metadata`) remain updatable; only `nickname` (and the already-immutable `username`/`is_domain_user`) are blocked.
- [ ] `nickname` removed from `Update()` PATCH payload — no inlined no-op API call.
- [ ] Post-PATCH `nickname` hydration removed from `Update()` — plan value preserved to prevent post-apply inconsistency.
- [ ] `Create()` and `Read()` continue to hydrate `nickname` from CM API response — unchanged.
- [ ] CM API endpoint `PATCH /v1/usermgmt/users/{user_id}` confirmed present in swagger `operations.md`.
- [ ] New test functions use `TestAccCMUser_` prefix matching the project's acceptance-test naming convention.
- [ ] Each test generates a unique per-run resource name using `uuid.New().String()[:8]` to prevent 409 conflicts on retries.
- [ ] No `t.Skip`, `t.Fatal`, or `t.Fatalf` inside `PreConfig` closures.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-407

PR #344, opened 2026-07-16
TFIN-408: ciphertrust_user: clearing user_metadata to null never converges — apply reports success but plan perpetually re-proposes the same diff

ashwani573:
## Summary

- Fixes `ciphertrust_user` `Update()` so that removing `user_metadata` keys — including a full clear — actually converges instead of perpetually re-proposing the same diff.
- Replaces a single-branch `!plan.Metadata.IsNull()` payload guard with a three-branch switch that correctly handles new/updated keys, partial key removal, and full-clear semantics against CM's merge-patch PATCH endpoint.
- Makes the post-PATCH read-back plan-aware so that a clearing operation does not oscillate between `{}` and `null` on consecutive plan runs.
- Adds `TestAccCMUser_UserMetadata` acceptance test covering full clear convergence, partial key removal convergence, and out-of-band drift detection.

## Motivation

Implements TFIN-408: `ciphertrust_user` — clearing `user_metadata` to null never converges; `terraform apply` reports success but `terraform plan` perpetually re-proposes the same diff.

The bug resided entirely in `Update()`. The CM `PATCH /v1/usermgmt/users/{user_id}` endpoint (swagger area: `auth-users`) applies JSON merge-patch semantics to `user_metadata`:

- Omitting the field preserves all existing keys unchanged.
- Sending `{"user_metadata": {}}` is a no-op — empty merge does nothing.
- Sending `{"user_metadata": null}` is rejected with HTTP 422.
- Sending `{"user_metadata": {"key": null}}` deletes that specific key.

Consequence: clearing one or more keys requires explicit per-key `null` values. The old single-branch guard (`if !plan.Metadata.IsNull()`) only built a payload for the non-null case and omitted it entirely when the plan cleared metadata, so CM never received any delete instruction and the existing keys persisted.

**Payload builder** — replaced with a three-branch `switch`:

1. `!plan.Metadata.IsNull() && !plan.Metadata.IsUnknown()` — non-null plan (new/updated or partial removal): builds a payload from the surviving plan keys (via the existing `stringsToRawJSON` helper) and additionally injects `json.RawMessage("null")` for every key present in prior state but absent from the plan.
2. `plan.Metadata.IsNull() && !state.Metadata.IsNull()` — full clear: iterates prior-state keys and sets each to `json.RawMessage("null")` in the payload.
3. Default — both plan and state are null, or plan is unknown: nothing added to the payload (no change to CM).

**Post-PATCH read-back** — the existing `if !plan.Metadata.IsNull()` gate that guarded re-hydration of `user_metadata` from the CM response was inverted to a two-branch check:

- Null plan: unconditionally write `types.MapNull(types.StringType)` into state regardless of what CM returns. This prevents a `{}` response from CM echoing back as an empty-but-non-null map and causing an immediate re-diff.
- Non-null plan: re-read from the CM response authoritatively (existing logic unchanged).

**Logging** — added `id := uuid.New().String()` at the top of `Update()` with `tflog.Trace` entry and deferred exit lines matching the `[resource_cm_user.go -> Update][id]` convention used elsewhere. Error-path `tflog.Debug` calls updated to use `id` instead of `plan.UserID.ValueString()` (which is an Optional field and may be empty).

- Adds `"github.com/google/uuid"` import (needed for unique per-run usernames and client call IDs).
- Adds `TestAccCMUser_UserMetadata` (6 steps):
  - Steps 1–2: create a user with `user_metadata = {"repro_key": "repro_value"}`, then re-apply with `user_metadata` omitted from config. Post-apply plan must be empty (Terraform's built-in convergence check). A CM-side `GetById` call confirms the key is absent on the server.
  - Steps 3–4: re-apply with two keys `{"a": "1", "b": "2"}`, then drop key `"b"`. Post-apply plan must be empty. CM-side verification confirms `"b"` is absent and `"a"` is preserved.
  - Steps 5–6: apply with one key `{"c": "3"}`, then inject key `"d"` out-of-band via a `PreConfig` `UpdateData` call. `RefreshState: true` + `ExpectNonEmptyPlan: true` confirms the drift is surfaced.
  - `PreConfig` closures use `t.Logf` + `return` (never `t.Fatal`/`t.Skip`) to avoid goroutine-exit panics.

## Read() status

`Read()` was already implemented in this resource before this change and was not modified in this PR. The ticket concern was Update() convergence only.

## How to test

  ```
  go test ./internal/provider/ -run TestAccCMUser_UserMetadata -v -timeout 15m
  ```
  Scenarios covered:
  - **Full clear** — apply with `user_metadata` set, then remove it from config; plan after apply must show no changes; CM API confirms all keys deleted.
  - **Partial key removal** — apply with two keys, drop one; plan after apply must show no changes; CM API confirms dropped key is absent and surviving key is preserved.
  - **OOB drift detection** — add a key directly on CM between steps; `RefreshState` must produce a non-empty plan.

## Checklist

- [ ] `tflog.Trace` at entry/exit of `Update()`: `[resource_cm_user.go -> Update][uuid]` pattern.
- [ ] Fresh `uuid.New().String()` at the top of `Update()` — not reused across calls.
- [ ] URL constant `URL_USER_MANAGEMENT` used — no inlined string literals.
- [ ] CM API endpoint `PATCH /v1/usermgmt/users/{user_id}` confirmed in swagger `auth-users` operations (merge-patch semantics verified against observed CM behavior; swagger area JSON unavailable in this workspace).

---

### DIFF
--- a/internal/provider/cm/resource_cm_user.go
+++ b/internal/provider/cm/resource_cm_user.go
@@ -329,6 +329,10 @@ func (r *resourceCMUser) Read(ctx context.Context, req resource.ReadRequest, res
 
 // Update updates the resource and sets the updated Terraform state on success.
 func (r *resourceCMUser) Update(ctx context.Context, req resource.UpdateRequest, resp *resource.UpdateResponse) {
+	id := uuid.New().String()
+	tflog.Trace(ctx, common.MSG_METHOD_START+"[resource_cm_user.go -> Update]["+id+"]")
+	defer tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_cm_user.go -> Update]["+id+"]")
+
 	var plan CMUserTFSDK
 	diags := req.Plan.Get(ctx, &plan)
 	resp.Diagnostics.Append(diags...)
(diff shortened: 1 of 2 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-408

PR #350, opened 2026-07-17
TFIN-414: ciphertrust_cm_reg_token: labels/lifetime never converge — plan perpetually re-proposes both fields

ashwani573:
## Summary

- Fixes `ciphertrust_cm_reg_token` `Read()`: removes the `else`-null branch on `lifetime` so the prior state value is preserved across refreshes (CM never returns `lifetime` in `GET /v1/client-management/regtokens/{id}` responses — confirmed by live reproduction).
- Fixes `ciphertrust_cm_reg_token` `Read()`: wraps labels hydration with `if !state.Labels.IsNull()` so that configs that omit `labels` are not polluted by the empty-map `{}` CM returns after every PATCH.
- Adds two acceptance tests to `internal/provider/resource_cm_reg_token_test.go` that act as regression gates for both drift bugs.

## Motivation

Implements TFIN-414: `ciphertrust_cm_reg_token` — `labels`/`lifetime` never converge — plan perpetually re-proposes both fields.

Users who set `lifetime` (e.g. `"10h"`) or omit `labels` in their config saw Terraform propose a change on every subsequent `terraform plan`, even when nothing had actually changed in CipherTrust Manager. The perpetual drift forced users to either accept a no-op `terraform apply` on every run or to remove the field from their config entirely.

**Fix (a) — lifetime (write-only field, CM never echoes it in GET)**

Before this change the `lifetime` guard looked like:

```go
if !state.Lifetime.IsNull() {
    if r := gjson.Get(response, "lifetime"); r.Exists() {
        state.Lifetime = types.StringValue(r.String())
    } else {
        state.Lifetime = types.StringNull()  // <- erased prior state on every refresh
    }
}
```

CM does not include `lifetime` in the `GET /v1/client-management/regtokens/{id}` response — it is accepted on POST/PATCH but is not echoed back. The `else` branch therefore fired on every `Read()` call, setting `state.Lifetime` to null. On the next plan, Terraform saw the config still had `lifetime = "10h"` but state had null, and proposed adding the field — producing perpetual drift.

The fix removes the `else` branch. When `lifetime` is absent from the GET response (always), the prior state value is left unchanged:

```go
if !state.Lifetime.IsNull() {
    if r := gjson.Get(response, "lifetime"); r.Exists() {
        state.Lifetime = types.StringValue(r.String())
    }
    // No else branch: CM never returns lifetime in GET responses (write-only).
    // Prior state value is retained as-is, preventing perpetual drift.
}
```

**Fix (b) — labels (empty-object echo after PATCH)**

Before this change, labels hydration ran unconditionally:

```go
labelsResult := gjson.Get(response, "labels")
if !labelsResult.Exists() || labelsResult.Type == gjson.Null {
    state.Labels = types.MapNull(types.StringType)
} else if len(labelsResult.Map()) == 0 {
    state.Labels = types.MapValueMust(types.StringType, map[string]attr.Value{})
} else {
    // ... populate from CM response
}
```

CM echoes `"labels": {}` in the PATCH response whenever labels were not set by the user. After an update, `Read()` would call the GET endpoint, receive `labels: {}`, hit the `len == 0` branch, and write an empty map into state. The next `terraform plan` saw `state.labels = {}` but `config.labels = null` and proposed removing the field — perpetual drift.

The fix wraps the entire block with a null-state guard:

```go
if !state.Labels.IsNull() {
    labelsResult := gjson.Get(response, "labels")
    // ... three-branch hydration unchanged ...
}
```

When `labels` was never configured (state is null), `Read()` skips the hydration entirely and leaves state null, eliminating the drift.

Adds two acceptance tests as permanent regression gates:

- `Test_CM_CMRegToken_LifetimeNoDrift` — 4-step test: Create with `lifetime = "10h"`, no-op plan (step 2 fails if Fix (a) is reverted), Update to change `max_clients`, no-op plan after update (step 4 fails on regression).
- `Test_CM_CMRegToken_LabelsNoDrift` — 3-step test: Create omitting `labels`, Update to trigger a PATCH (CM returns `labels: {}`), no-op plan after update (step 3 fails if Fix (b) is reverted).

Both tests call `RequireCM(t)` as their first statement and follow the `Test_CM_` naming convention.

## Read() fixes — summary

`Read()` was already implemented for this resource. These changes correct two specific hydration bugs within it; they do not add or remove overall Read() coverage.

**lifetime** — field absent from `GET /v1/client-management/regtokens/{id}` response (write-only, confirmed by live reproduction). Prior state is preserved unchanged when CM does not return the field, which is the correct behaviour for write-only fields in this provider.

**labels** — CM returns `labels: {}` in GET/PATCH responses even when the user never configured labels. The `!state.Labels.IsNull()` outer guard prevents unconditional hydration from overwriting a null state with an empty map, which would produce a perpetual null-to-empty-map plan diff.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_CMRegToken_LifetimeNoDrift|Test_CM_CMRegToken_LabelsNoDrift' -v -timeout 15m
  ```
  Scenarios covered:
  - **LifetimeNoDrift — Create**: apply config with `lifetime = "10h"`; assert state attribute is `"10h"`.
  - **LifetimeNoDrift — no-op after Create**: re-apply same config; assert no changes proposed (regression gate for Fix a).
  - **LifetimeNoDrift — Update**: change `max_clients`; assert `lifetime` is still `"10h"` in state.
  - **LifetimeNoDrift — no-op after Update**: re-apply same config; assert no changes proposed (second regression gate for Fix a).
  - **LabelsNoDrift — Create**: apply config omitting `labels`; assert `labels` attribute is not present in state.
  - **LabelsNoDrift — Update**: change `max_clients` (triggers PATCH; CM returns `labels: {}`); assert `labels` is still absent from state.
  - **LabelsNoDrift — no-op after Update**: re-apply same config; assert no changes proposed (regression gate for Fix b).
- [ ] Manual smoke-test: apply a config with `lifetime` set, run `terraform plan` a second time, confirm the plan shows no changes.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_cm_reg_token.go -> <Func>][uuid]` pattern — no new CRUD methods introduced; existing logging unchanged.
- [ ] URL constant `URL_REG_TOKEN` already defined in `urls.go` — no new inlined string literals.
- [ ] CM API endpoint `GET /v1/client-management/regtokens/{id}` verified in `operations.md` — endpoint exists; `lifetime` confirmed absent from GET response by live reproduction.
- [ ] All new test functions use `Test_CM_` prefix (`Test_CM_CMRegToken_LifetimeNoDrift`, `Test_CM_CMRegToken_LabelsNoDrift`).
- [ ] No hand-edited files under `docs/` — schema is unchanged, regeneration not required.

---

### DIFF
--- a/docs/data-sources/cm_users_list.md
+++ b/docs/data-sources/cm_users_list.md
@@ -33,7 +33,7 @@ Read-Only:
 - `is_domain_user` (Boolean)
 - `name` (String)
 - `nickname` (String)
-- `password` (String)
+- `password` (String, Sensitive)
 - `password_change_required` (Boolean)
 - `prevent_ui_login` (Boolean)
 - `user_id` (String)
--- a/internal/provider/cm/resource_cm_reg_token.go
+++ b/internal/provider/cm/resource_cm_reg_token.go
@@ -239,9 +239,9 @@ func (r *resourceCMRegToken) Read(ctx context.Context, req resource.ReadRequest,
 	if !state.Lifetime.IsNull() {
 		if r := gjson.Get(response, "lifetime"); r.Exists() {
 			state.Lifetime = types.StringValue(r.String())
-		} else {
-			state.Lifetime = types.StringNull()
 		}
+		// No else branch: CM never returns lifetime in GET responses (write-only).
+		// Prior state value is retained as-is, preventing perpetual drift.
 	}
 
 	if !state.NamePrefix.IsNull() {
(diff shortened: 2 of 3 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-414

PR #368, opened 2026-07-21
TFIN-422: ciphertrust_scheduler: start_date/end_date cannot be cleared via Terraform config even though CM API fully supports it (schema regex rejects empty string)

ashwani573:
## Summary

- Fixes `ciphertrust_scheduler` resource: schema validators on `start_date` and `end_date` rejected `""`, making it impossible to clear a previously-set date via Terraform config even though `PATCH /v1/scheduler/job-configs/{id}` supports clearing these fields.
- Fixes `ciphertrust_scheduler_list` data source: absent dates from the CM list response were rendered as `"0001-01-01T00:00:00Z"` (Go `time.Time` zero value) instead of null (TFIN-423).
- Fixes `ciphertrust_scheduler_list` data source: `cckm_key_rotation_params.aws_params` was a nested schema block that does not exist as a nested object in the CM API response; replaced with flat `aws_retain_alias` and `rotate_material` attributes (TFIN-424).
- Fixes `ciphertrust_ntp` resource: `key` and `key_type` used the provider-forbidden `stringplanmodifier.RequiresReplaceIfConfigured()` modifier; replaced with `modifiers.ImmutableString()` (TFIN-425).

## Motivation

Implements TFIN-422: `ciphertrust_scheduler` `start_date`/`end_date` cannot be cleared via Terraform config even though CM API fully supports it (schema regex rejects empty string). Three related regressions in the same files were identified and resolved in the same PR (TFIN-423, TFIN-424, TFIN-425).

- `start_date` and `end_date` validators changed from a bare `stringvalidator.RegexMatches()` to `stringvalidator.Any(regexValidator, stringvalidator.OneOf(""))`. This allows `""` as a valid value, which signals "send an explicit clear to CM." The regex still enforces the `YYYY-MM-DDTHH:MM:SSZ` format for non-empty values.
- Null-is-no-op semantics introduced across Create, Read, and Update:
  - **Create**: `start_date`/`end_date` are only included in the POST body when the plan value is non-null. Setting `""` sends `&""` to CM (clearing). Omitting the field from HCL (null plan value) skips it entirely.
  - **Read**: fields are only refreshed from the CM response when the prior state value is non-null, so users who never configure these fields see a stable null in state with no spurious diffs.
  - **Update**: fields are only sent in the PATCH body when the plan value is non-null and has changed from state. A `""` plan value sends the clear signal; null means "don't touch."
- Removed `time.Parse(time.RFC3339, ...)` error paths from Create and Update — values are now passed through as raw strings; format validation is handled entirely by the schema validators.
- `getParamsFromResponse` no longer sets `StartDate`/`EndDate`; each CRUD caller handles its own hydration after the API call returns.

- `StartDate`/`EndDate` in `CreateJobConfigParamsJSON` and `UpdateJobConfigParamsJSON`: changed from `time.Time` to `*string` with `json:"...,omitempty"`. A nil pointer is omitted from the JSON payload, enabling the null-is-no-op pattern. A non-nil pointer containing `""` marshals as an explicit empty string, which CM interprets as a clear.
- `StartDate`/`EndDate` in `CreateJobConfigParamsListJSON` (list response deserialisation struct): changed from `time.Time` to `string` so that absent dates deserialise as `""` rather than the zero `time.Time`.
- Removed `CCKMAwsKeyRotationParamsDatasourceTFSDK` nested struct. Promoted `RetainAlias` (`aws_retain_alias`) and `RotateMaterial` (`rotate_material`) as direct fields on `CCKMKeyRotationParamsDatasourceTFSDK`, aligning the TFSDK struct with the flat attribute schema.
- Removed the `"time"` import (no longer referenced).

### Modified file — `internal/provider/cm/data_source_scheduler.go`

- Removed the `aws_params` nested schema block from `cckm_key_rotation_params`; added flat `aws_retain_alias` (bool, Computed) and `rotate_material` (bool, Computed) attributes at the `cckm_key_rotation_params` level, matching how the CM API embeds these fields.
- Date hydration in `Read()`: added a null-safe branch before building each `JobConfigParamsTFSDK` — when `jobs.StartDate` or `jobs.EndDate` is an empty string (CM did not set the field), the corresponding attribute is set to `types.StringNull()` instead of formatting a zero `time.Time` as `"0001-01-01T00:00:00Z"`.
- `getCCKMKeyRotationParams`: removed `AwsParams` nested struct assignment; assigns `AWSRetainAlias` and `RotateMaterial` directly from the embedded `CCKMRotationAwsParamsJSON` fields. Added `else` branches for `Expiration`, `ExpireIn`, and `RotationAfter` so that absent pointer fields resolve to `types.StringNull()` rather than the zero `types.String{}`.
- Removed the `"time"` import.

- `key` and `key_type` schema attributes: `stringplanmodifier.RequiresReplaceIfConfigured()` replaced with `modifiers.ImmutableString()`. This emits a plan-time error when the user attempts to change either field after initial creation, consistent with the provider-wide convention. No destroy+recreate is triggered.
- Updated `Description` for both attributes to include the `(Immutable)` prefix, which surfaces the constraint in generated docs.

- Added helper `schedulerDatesConfig(name, startDate, endDate string)`: generates a `database_backup` scheduler HCL config. Pass `"omit"` to exclude the field (null), `""` to explicitly clear it, or a date string to set it.
- Added `Test_CM_Scheduler_StartDateEndDateClear`: multi-step test verifying set → clear (`""`) → omit (null, stable plan) → re-set lifecycle for both `start_date` and `end_date`.
- Added `Test_CM_Scheduler_StartDateEndDateDrift`: creates a scheduler with `start_date` set, then modifies `start_date` out-of-band in CM via the client API, and asserts that the next `terraform plan` detects a non-empty diff.
- Added `Test_CM_SchedulerList_DateAndAWSParams`: creates a `cckm_key_rotation` scheduler, reads it via the `ciphertrust_scheduler_list` data source, and asserts that absent dates are null (not zero-time) and that `aws_retain_alias` is accessible as a flat attribute.

- Updated `Test_CM_AccCipherTrust_NTP_ImmutableFields` steps 3 and 4: changed from `ExpectNonEmptyPlan: true` to `ExpectError: regexp.MustCompile("(?i)immutable")` to match the new `modifiers.ImmutableString()` behaviour, which fires a plan-time error rather than a replace plan.
- Updated `Test_CM_CipherTrust_NTP_OptionalFieldAdded_ForcesReplacement` final step: same change — expects a plan-time error, not a replace plan.
- Fixed two `PreConfig` closures where `t.Fatal("could not create CM client")` was called inside the closure goroutine (would cause a runtime panic). Replaced with `t.Logf(...); return` per the provider test convention.

### `ciphertrust_scheduler` resource — `start_date` and `end_date`

| Attribute | Before | After | Notes |
|---|---|---|---|
| `start_date` | `Optional+Computed`, regex-only validator | `Optional+Computed`, `Any(regex, OneOf(""))` validator | `""` now valid; clears the date in CM |
| `end_date` | `Optional+Computed`, regex-only validator | `Optional+Computed`, `Any(regex, OneOf(""))` validator | `""` now valid; clears the date in CM |

### `ciphertrust_scheduler_list` data source — `cckm_key_rotation_params`

| Before | After | Notes |
|---|---|---|
| `aws_params` (nested block, Computed) | Removed | Block did not exist in CM API response |
| `aws_params.retain_alias` (bool, Computed) | Removed | |
| `aws_params.rotate_material` (bool, Computed) | Removed | |
| — | `aws_retain_alias` (bool, Computed, flat) | Promoted from nested block |
| — | `rotate_material` (bool, Computed, flat) | Promoted from nested block |

### `ciphertrust_ntp` resource

| Attribute | Before | After |
|---|---|---|
| `key` | `Optional`, `RequiresReplaceIfConfigured()` | `Optional`, `modifiers.ImmutableString()` |
| `key_type` | `Optional+validator`, `RequiresReplaceIfConfigured()` | `Optional+validator`, `modifiers.ImmutableString()` |

## Read() status

`Read()` was already implemented in `ciphertrust_scheduler` before this PR. This change refines how `start_date` and `end_date` are hydrated:

Previously both fields were unconditionally written from the API response in `getParamsFromResponse`, which caused problems when the user had never configured these fields (null state got overwritten with CM's value on every refresh).

Now the hydration uses a null-is-no-op guard: if the prior state value is null (user never configured the field), Read does not overwrite it with the API value. If the prior state value is non-null (user previously set the field), Read updates state from the API response, enabling drift detection for explicitly configured dates.

**Limitation:** Drift for `start_date`/`end_date` is only detected when the user has explicitly set those fields in their Terraform config. An out-of-band CM change to a date that was never in the user's config will not surface as a plan diff. This is an intentional trade-off for null-is-no-op semantics.

## Breaking changes

**`ciphertrust_scheduler_list` data source** — the `aws_params` nested attribute block inside `cckm_key_rotation_params` is removed. Any existing configuration or `terraform state` that references `aws_params` within this block will break after upgrading.

Migration: replace references to the nested path with the new flat attributes:

```hcl
# Before
data.ciphertrust_scheduler_list.X.scheduler_jobs[i].cckm_key_rotation_params.aws_params.retain_alias
data.ciphertrust_scheduler_list.X.scheduler_jobs[i].cckm_key_rotation_params.aws_params.rotate_material

# After
data.ciphertrust_scheduler_list.X.scheduler_jobs[i].cckm_key_rotation_params.aws_retain_alias
data.ciphertrust_scheduler_list.X.scheduler_jobs[i].cckm_key_rotation_params.rotate_material
```

**`ciphertrust_ntp` resource** — `key` and `key_type` now emit a plan-time error when a user attempts to change them after the resource is created, instead of triggering a destroy+recreate. Users who never change these fields after initial creation are unaffected. Automation that previously relied on the replace behaviour will see a diagnostic error and must instead destroy and re-create the NTP resource manually.

## How to test

Acceptance tests (require live CM — set `TF_ACC=1`, `CIPHERTRUST_*` env vars, and `CIPHERTRUST_SCHEDULER_ENABLED=1`):

```
go test ./internal/provider/... -run Test_CM_Scheduler_StartDateEndDateClear -v -timeout 15m
go test ./internal/provider/... -run Test_CM_Scheduler_StartDateEndDateDrift -v -timeout 15m
go test ./internal/provider/... -run Test_CM_SchedulerList_DateAndAWSParams -v -timeout 15m
go test ./internal/provider/... -run Test_CM_AccCipherTrust_NTP_ImmutableFields -v -timeout 15m
go test ./internal/provider/... -run Test_CM_CipherTrust_NTP_OptionalFieldAdded_ForcesReplacement -v -timeout 15m
```

Scenarios covered by new and updated tests:

**Scheduler date set/clear lifecycle (`Test_CM_Scheduler_StartDateEndDateClear`):**
- **Set**: apply config with valid RFC3339 date strings; assert state reflects the configured dates.
- **Clear**: update config with `start_date = ""`; assert state stores `""` and CM has cleared the date.
- **Null**: remove `start_date`/`end_date` from HCL entirely; assert state becomes null and subsequent plan is stable (no diff).
- **Re-set**: re-add a date after clearing; assert the date is stored correctly.

**Scheduler date drift detection (`Test_CM_Scheduler_StartDateEndDateDrift`):**
- Create a scheduler with `start_date` set; modify `start_date` out-of-band in CM via the API; assert the next `terraform plan` produces a non-empty diff.

**Scheduler list data source (`Test_CM_SchedulerList_DateAndAWSParams`):**
- Verify absent dates are null in state (not `"0001-01-01T00:00:00Z"`).
- Verify `cckm_key_rotation_params.aws_retain_alias` is accessible as a flat attribute (not nested under `aws_params`).

**NTP immutability (`Test_CM_AccCipherTrust_NTP_ImmutableFields`, `Test_CM_CipherTrust_NTP_OptionalFieldAdded_ForcesReplacement`):**
- Changing `key` or `key_type` after creation must produce a plan-time error matching `(?i)immutable`.
- Adding `key`/`key_type` to an existing NTP resource (null → value) must produce a plan-time error matching `(?i)immutable`.

- [ ] Manual smoke-test: apply a scheduler with `start_date` set; verify the date appears in CM UI; update config with `start_date = ""`; apply; verify the date field is absent in CM UI; destroy.

## Checklist

- [ ] `tflog` trace logging present at entry/exit of CRUD methods: `[<file>.go -> <Func>][uuid]` pattern.
- [ ] CM API endpoint verified against swagger spec (`PATCH /v1/scheduler/job-configs/{id}` confirmed in `cm-ops` area).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+## Unreleased
+
+### BREAKING CHANGES
+
+- **TFIN-423**: `ciphertrust_scheduler_list` — the `cckm_key_rotation_params.aws_params` nested block has been removed. Users must update HCL references to the new flat attributes `cckm_key_rotation_params.aws_retain_alias` and `cckm_key_rotation_params.rotate_material`.
+
+### Bug Fixes
+
+- **TFIN-422**: `ciphertrust_scheduler` — `start_date` and `end_date` now accept an empty string (`""`) to clear a previously set value. Previously, the schema regex validator rejected empty strings, making it impossible to clear these fields via Terraform config even though the CM API supports it.
+- **TFIN-424**: `ciphertrust_scheduler_list` — `start_date` and `end_date` were previously stored as `time.Time` (producing the zero value `"0001-01-01T00:00:00Z"` when the CM API omitted the field). They are now stored as raw strings from the CM API response or null when absent. No state migration is required — the prior zero-time value was always incorrect.
--- a/docs/data-sources/scheduler_list.md
+++ b/docs/data-sources/scheduler_list.md
@@ -53,20 +53,12 @@ Read-Only:
 
 Read-Only:
 
-- `aws_params` (Attributes) (see [below for nested schema](#nestedatt--scheduler--cckm_key_rotation_params--aws_params))
+- `aws_retain_alias` (Boolean)
 - `cloud_name` (String)
 - `expiration` (String)
 - `expire_in` (String)
-- `rotation_after` (String)
-
-<a id="nestedatt--scheduler--cckm_key_rotation_params--aws_params"></a>
-### Nested Schema for `scheduler.cckm_key_rotation_params.aws_params`
-
-Read-Only:
-
-- `retain_alias` (Boolean)
 - `rotate_material` (Boolean)
-
+- `rotation_after` (String)
 
 
 <a id="nestedatt--scheduler--cckm_synchronization_params"></a>
--- a/docs/resources/ntp.md
+++ b/docs/resources/ntp.md
@@ -64,8 +64,8 @@ output "ntp_server_host" {
 
 ### Optional
 
-- `key` (String) Symmetric key value to be used for authenticated NTP servers. Changing this value forces replacement of the NTP resource.
-- `key_type` (String) Digest algorithm to be used for authenticated NTP servers; MD5, SHA-1, SHA-256, SHA-384 or SHA-512 (defaults to SHA-256). Changing this value forces replacement of the NTP resource.
+- `key` (String) (Immutable) NTP authentication key value. Cannot be changed after creation.
+- `key_type` (String) (Immutable) NTP authentication key type (e.g. SHA-256). Cannot be changed after creation.
 
 ### Read-Only
 
--- a/docs/resources/scheduler.md
+++ b/docs/resources/scheduler.md
@@ -121,9 +121,9 @@ For example:
 - `database_backup_params` (Attributes) Database backup operation specific arguments. Should be JSON-serializable. Required only for "database_backup" operations. Not allowed for other operations. (see [below for nested schema](#nestedatt--database_backup_params))
 - `description` (String) Description for the job configuration.
 - `disabled` (Boolean) By default, the job configuration starts in an active state. True disables the job configuration.
-- `end_date` (String) Date the job configuration becomes inactive. RFC3339 format. For example, 2018-10-02T14:24:37.436073Z
+- `end_date` (String) Date/time when the job ends. Format: YYYY-MM-DDTHH:MM:SSZ. Set to "" to clear a previously set value. Omitting or setting to null leaves the field unchanged in CM (no PATCH is sent for this field).
 - `run_on` (String) Default is 'any'. For database_backup, the default will be the current node if in a cluster. This attribute is not supported in CDSPaaS.
-- `start_date` (String) Date the job configuration becomes active. RFC3339 format. For example, 2018-10-02T14:24:37.436073Z
+- `start_date` (String) Date/time when the job starts. Format: YYYY-MM-DDTHH:MM:SSZ. Set to "" to clear a previously set value. Omitting or setting to null leaves the field unchanged in CM (no PATCH is sent for this field).
 
 ### Read-Only
 
--- a/internal/provider/cm/data_source_scheduler.go
+++ b/internal/provider/cm/data_source_scheduler.go
@@ -199,7 +199,7 @@ func (d *dataSourceScheduler) Read(ctx context.Context, req datasource.ReadReque
 		return
 	}
 
-	gjson.Parse(jsonStr).ForEach(func(_, item gjson.Result) bool {
+	gjson.Get(jsonStr, "resources").ForEach(func(_, item gjson.Result) bool {
 		jobJSON := item.Raw
 
 		var startDate, endDate types.String
--- a/internal/provider/cm/data_source_scheduler.go
+++ b/internal/provider/cm/data_source_scheduler.go
@@ -199,7 +199,7 @@ func (d *dataSourceScheduler) Read(ctx context.Context, req datasource.ReadReque
 		return
 	}
 
-	gjson.Get(jsonStr, "resources").ForEach(func(_, item gjson.Result) bool {
+	gjson.Parse(jsonStr).ForEach(func(_, item gjson.Result) bool {
 		jobJSON := item.Raw
 
 		var startDate, endDate types.String
--- a/internal/provider/cm/data_source_scheduler.go
+++ b/internal/provider/cm/data_source_scheduler.go
@@ -7,7 +7,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/attr"
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"strings"
-	"time"
 
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
 	"github.com/google/uuid"
@@ -15,6 +14,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
+	"github.com/tidwall/gjson"
 )
 
 var (
@@ -119,17 +119,6 @@ func (d *dataSourceScheduler) Schema(_ context.Context, _ datasource.SchemaReque
 						"cckm_key_rotation_params": schema.SingleNestedAttribute{
 							Computed: true,
 							Attributes: map[string]schema.Attribute{
-								"aws_params": schema.SingleNestedAttribute{
-									Computed: true,
-									Attributes: map[string]schema.Attribute{
-										"retain_alias": schema.BoolAttribute{
-											Computed: true,
-										},
-										"rotate_material": schema.BoolAttribute{
-											Computed: true,
-										},
-									},
-								},
 								"cloud_name": schema.StringAttribute{
 									Computed: true,
 								},
@@ -140,6 +129,12 @@ func (d *dataSourceScheduler) Schema(_ context.Context, _ datasource.SchemaReque
 									Computed: true,
 								},
 								"rotation_after": schema.StringAttribute{
+									Computed: true,
+								},
+								"aws_retain_alias": schema.BoolAttribute{
+									Computed: true,
+								},
+								"rotate_material": schema.BoolAttribute{
 									Computed: true,
 								},
 							},
@@ -204,51 +199,60 @@ func (d *dataSourceScheduler) Read(ctx context.Context, req datasource.ReadReque
 		return
 	}
 
-	schedulerJobConfigs := []CreateJobConfigParamsListJSON{}
+	gjson.Parse(jsonStr).ForEach(func(_, item gjson.Result) bool {
+		jobJSON := item.Raw
 
-	err = json.Unmarshal([]byte(jsonStr), &schedulerJobConfigs)
-	if err != nil {
-		tflog.Debug(ctx, common.ERR_METHOD_END+err.Error()+" [resource_scheduler.go -> Read]["+id+"]")
-		resp.Diagnostics.AddError(
-			"Unable to read scheduler job configs from CM",
-			err.Error(),
-		)
-		return
-	}
+		var startDate, endDate types.String
+		if r := gjson.Get(jobJSON, "start_date"); r.Exists() {
+			startDate = types.StringValue(r.String())
+		} else {
+			startDate = types.StringNull()
+		}
+		if r := gjson.Get(jobJSON, "end_date"); r.Exists() {
+			endDate = types.StringValue(r.String())
+		} else {
+			endDate = types.StringNull()
+		}
 
-	for _, jobs := range schedulerJobConfigs {
 		schedulerJobs := JobConfigParamsTFSDK{
 			CreateJobConfigParamsTFSDKCommon: CreateJobConfigParamsTFSDKCommon{
-				ID:          types.StringValue(jobs.ID),
-				URI:         types.StringValue(jobs.URI),
-				Account:     types.StringValue(jobs.Account),
-				Application: types.StringValue(jobs.Application),
-				DevAccount:  types.StringValue(jobs.DevAccount),
-				CreatedAt:   types.StringValue(jobs.CreatedAt),
-				UpdatedAt:   types.StringValue(jobs.UpdatedAt),
-				Name:        types.StringValue(jobs.Name),
-				Description: types.StringValue(jobs.Description),
-				Operation:   types.StringValue(jobs.Operation),
-				RunAt:       types.StringValue(jobs.RunAt),
-				RunOn:       types.StringValue(jobs.RunOn),
-				Disabled:    types.BoolValue(jobs.Disabled),
-				StartDate:   types.StringValue(jobs.StartDate.Format(time.RFC3339)),
-				EndDate:     types.StringValue(jobs.EndDate.Format(time.RFC3339)),
+				ID:          types.StringValue(gjson.Get(jobJSON, "id").String()),
+				URI:         types.StringValue(gjson.Get(jobJSON, "uri").String()),
+				Account:     types.StringValue(gjson.Get(jobJSON, "account").String()),
+				Application: types.StringValue(gjson.Get(jobJSON, "application").String()),
+				DevAccount:  types.StringValue(gjson.Get(jobJSON, "devAccount").String()),
+				CreatedAt:   types.StringValue(gjson.Get(jobJSON, "createdAt").String()),
+				UpdatedAt:   types.StringValue(gjson.Get(jobJSON, "updatedAt").String()),
+				Name:        types.StringValue(gjson.Get(jobJSON, "name").String()),
+				Description: types.StringValue(gjson.Get(jobJSON, "description").String()),
+				Operation:   types.StringValue(gjson.Get(jobJSON, "operation").String()),
+				RunAt:       types.StringValue(gjson.Get(jobJSON, "run_at").String()),
+				RunOn:       types.StringValue(gjson.Get(jobJSON, "run_on").String()),
+				Disabled:    types.BoolValue(gjson.Get(jobJSON, "disabled").Bool()),
+				StartDate:   startDate,
+				EndDate:     endDate,
 			},
 		}
 
-		switch jobs.Operation {
+		operation := gjson.Get(jobJSON, "operation").String()
+		jobConfigParamsRaw := json.RawMessage("{}")
+		if r := gjson.Get(jobJSON, "job_config_params"); r.Exists() {
+			jobConfigParamsRaw = json.RawMessage(r.Raw)
+		}
+
+		switch operation {
 		case "database_backup":
-			getDataBaseBackupParams(ctx, id, &schedulerJobs, jobs.JobConfigParams, &resp.Diagnostics)
+			getDataBaseBackupParams(ctx, id, &schedulerJobs, jobConfigParamsRaw, &resp.Diagnostics)
 		case "cckm_key_rotation":
-			getCCKMKeyRotationParams(ctx, id, &schedulerJobs, jobs.JobConfigParams, &resp.Diagnostics)
+			getCCKMKeyRotationParams(ctx, id, &schedulerJobs, jobConfigParamsRaw, &resp.Diagnostics)
 		case "cckm_synchronization":
-			getCCKMSynchronizationParams(ctx, id, &schedulerJobs, jobs.JobConfigParams, &resp.Diagnostics)
+			getCCKMSynchronizationParams(ctx, id, &schedulerJobs, jobConfigParamsRaw, &resp.Diagnostics)
 		case "cckm_xks_credential_rotation":
-			getCCKMCredentialRotationParams(ctx, id, &schedulerJobs, jobs.JobConfigParams, &resp.Diagnostics)
+			getCCKMCredentialRotationParams(ctx, id, &schedulerJobs, jobConfigParamsRaw, &resp.Diagnostics)
 		}
 		state.Scheduler = append(state.Scheduler, schedulerJobs)
-	}
+		return true
+	})
 
 	tflog.Trace(ctx, common.MSG_METHOD_END+"[resource_scheduler.go -> Read]["+id+"]")
 	diags := resp.State.Set(ctx, &state)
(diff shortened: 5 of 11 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-422

ashwani573:
## 🔍 Automated Code Review — TFIN-422

- Plan lists `internal/provider/resource_scheduler_test.go` as the file to modify for new scheduler test functions (Test_CM_Scheduler_StartDateEndDateClear, Test_CM_Scheduler_StartDateEndDateDrift, Test_CM_SchedulerList_DateAndAWSParams), and the diff correctly modifies that file. However, the plan also lists `internal/provider/resource_ntp_test.go` as a **Create** (new file), but the diff shows it as a **Modify** (it already exists at `internal/provider/resource_ntp_test.go` with existing content being modified). Per the TEST FILE PATH MISMATCH EXCEPTION, both files are in `internal/provider/` so the paths match — this is a plan description error (said Create but file already existed). Downgrading: the implementer correctly modified the existing file rather than creating a duplicate; the plan's 'Create' designation was wrong. This is a MEDIUM plan-authoring issue, not a HIGH implementer error.
- The plan specifies that Read() in `resource_scheduler.go` should unconditionally set `start_date`/`end_date` to `StringNull()` when absent (plan: `if r := gjson.Get(response, "start_date"); r.Exists() { ... } else { plan.StartDate = types.StringNull() }`). The diff instead implements a conditional guard in Read(): `if !state.StartDate.IsNull() { ... }` — meaning if state already has null, Read() leaves `start_date` completely untouched (no assignment at all). This deviates from the plan's specified Read() pattern and means if state is null (e.g. after import or first read), start_date/end_date are never updated from the CM response, causing permanent null even if CM returns a value.
- The plan specifies post-POST hydration in Create() should use an unconditional `r.Exists()` absent-branch: when user configured the field and CM omits it in the response, set `types.StringNull()`. The diff instead omits the else-null branch when `r.Exists()` is false and `plan.StartDate` is non-null (it keeps the plan value `""` rather than setting null). This means after a successful Create with `start_date = ""`, state holds `""` instead of null — subsequent Read() with `!state.StartDate.IsNull()` guard will then try to update from CM response, potentially causing inconsistency. The plan explicitly stated the absent-branch should produce `StringNull()` in Create().
- The `data_source_scheduler.go` Read() loop iterates over `gjson.Parse(jsonStr).ForEach(...)` but the plan specifies the response format as `gjson.Get(response, "resources").ForEach(...)`. The diff uses `gjson.Parse(jsonStr)` directly — if `jsonStr` is the full paginated CM response body (containing a `{"resources":[...], "total":N}` wrapper), this would iterate over the top-level keys (`resources` and `total`) rather than the array items, producing incorrect results. This is a potential runtime behavior deviation if CM returns a wrapped response.

### 🟢 LOW — informational
- In `resource_ntp.go`, the plan says to replace `stringplanmodifier.RequiresReplace()` but the diff shows the existing code used `stringplanmodifier.RequiresReplaceIfConfigured()` (not `RequiresReplace()`). The diff correctly replaces both with `modifiers.ImmutableString()`, which is the right fix. However, the plan description inaccurately describes the pre-existing modifier — this is a plan documentation inaccuracy, not an implementer error. No action needed from the implementer; noting for plan correction only.
- The diff adds `import "net/url"` and uses `client.ListWithFilters` in `schedulerSweepAll()` in `resource_scheduler_test.go`. These additions are not described in the plan's test case specifications for `resource_scheduler_test.go`. The plan only described three new test functions. While `schedulerSweepAll()` is a reasonable test helper, `client.ListWithFilters` and `net/url` import are unplanned additions — however, since build passed and these are test-only helpers that don't affect the provider API surface, this is LOW.
- The plan for TFIN-423 states `for non-CCKM jobs assign schedulerJob.CCKMKeyRotationParams = nil`. The diff's Read() in `data_source_scheduler.go` uses a switch statement that only calls `getCCKMKeyRotationParams` for `cckm_key_rotation` operation. However, the diff does NOT explicitly set `schedulerJobs.CCKMKeyRotationParams = nil` for non-CCKM jobs — the struct is zero-initialized so the pointer field starts as nil, which is correct by default. This is acceptable since Go zero-initialization of a pointer field is nil, matching the plan's intent. No issue.

ashwani573:
## 🔍 Automated Code Review — TFIN-422

ashwani573:
## 🔍 Automated Code Review — TFIN-422

- data_source_scheduler.go Read(): the diff iterates using `gjson.Parse(jsonStr).ForEach(...)` treating `jsonStr` as a JSON array. However, the existing code used `json.Unmarshal([]byte(jsonStr), &schedulerJobConfigs)` where `schedulerJobConfigs` was `[]CreateJobConfigParamsListJSON{}`. This implies `jsonStr` is a JSON array. But the plan's CRUD wiring section shows iterating `gjson.Get(response, "resources").ForEach(...)` — extracting the `resources` key from the response. The diff uses `gjson.Parse(jsonStr).ForEach(...)` directly on `jsonStr` without extracting a `resources` key. If `jsonStr` from the CM list API is the full response body `{"resources":[...], "total":N}` rather than a bare array, this would silently produce zero items. Since `make build` PASSED and the existing code unmarshal'd `jsonStr` directly as an array, `jsonStr` is likely already extracted from the `resources` field upstream, making this correct. However, the inconsistency with the plan's stated `gjson.Get(response, "resources").ForEach(...)` pattern is a MEDIUM risk if the upstream extraction changes.
- data_source_cckm_scheduler_test.go is modified in the diff but is NOT listed in the plan's 'Files to Modify' table. This is an unplanned file modification. However, the changes update test assertions from the old `aws_params.rotate_material`/`aws_params.retain_alias` paths to the new flat `rotate_material`/`aws_retain_alias` paths — these changes are a direct and necessary consequence of the TFIN-423 schema change (removing `aws_params` nested block). Without this update, the existing test would fail with the new schema. Downgraded to MEDIUM as the file should be listed in the plan's Files to Modify table for traceability.

### 🟢 LOW — informational
- resource_scheduler_test.go: Test_CM_Scheduler_StartDateEndDateDrift uses `client.UpdateData` with `schedulerURL = common.URL_SCHEDULER_JOB_CONFIGS + "/" + capturedID` (a relative path), but the existing UpdateData signature in this provider requires a relative endpoint path — this is consistent. However, verify the plan's stated check that `UpdateData` is the correct method to use here vs `UpdateDataV2`. The plan's Update() wiring section says post-PATCH hydration uses `UpdateDataV2` for the resource Update() method, but the test's OOB PreConfig correctly uses `UpdateData` (not the resource method). This is fine — no issue here. LOW: resource_scheduler.go Update() post-PATCH hydration uses a conditional `r.Exists()` guard without an else-null when plan is non-null and CM omits the field (the `else` branch only fires when plan is null/unknown). When the user sends `""` to clear the field and CM's response omits `start_date`, the code keeps `plan.StartDate` as `""` rather than setting it to `StringNull()`. The plan states the correct behavior is to keep the plan value as-is (`""`) in this case, so this is intentional and matches the plan. No issue.
- resource_scheduler.go Update(): the diff does NOT show `req.State.Get(ctx, &state)` being added — the Update() method references `state.StartDate` and `state.EndDate` in `!plan.StartDate.Equal(state.StartDate)` but the plan explicitly requires adding `req.State.Get` to load prior state. Searching the diff, the Update() function body does NOT include a new `req.State.Get` call — only the payload logic for start_date/end_date is shown. If `state` was already loaded via `req.State.Get` in the pre-existing Update() code (unchanged lines not in diff), this is fine. However, the diff removes the old `plan.StartDate.ValueString() != ""` guard and replaces it with `!plan.StartDate.Equal(state.StartDate)` which requires `state` to be populated. If `state` was not previously loaded in Update(), this would be a compile error — but since `make build` PASSED, `state` must already be declared and populated in the unchanged portion of Update(). Build passing confirms correctness — not flagged.
- resource_ntp.go: the diff replaces `stringplanmodifier.RequiresReplaceIfConfigured()` with `modifiers.ImmutableString()`, but the plan specifies replacing `stringplanmodifier.RequiresReplace()` (not `RequiresReplaceIfConfigured()`). The actual code uses `RequiresReplaceIfConfigured()` which is semantically different from `RequiresReplace()` — the former only fires when the field is configured (not null), while the latter fires on any change. This is a minor plan description inaccuracy but the implementation is correct (replacing whichever RequiresReplace variant existed with ImmutableString). No functional issue.
- resource_scheduler_test.go: the plan requires `Test_CM_Scheduler_StartDateEndDateClear` to use resource name `"ciphertrust_scheduler.test"` but the diff uses `"ciphertrust_scheduler.dates_test"`. This is a minor deviation from the plan's stated resource label but is internally consistent throughout the test function and does not affect correctness.
- resource_scheduler_test.go: the diff adds `schedulerSweepAll()`, `schedulerTestPrefixes`, `schedulerDatesConfig()` as new exported/package-level symbols, imports `"net/url"`, `"strings"`, `"context"`, `"github.com/tidwall/gjson"`, and `"github.com/hashicorp/terraform-plugin-testing/terraform"`. The plan does not explicitly describe these sweep helpers, but they are reasonable support infrastructure for the planned test functions and do not affect provider API surface. Accepted as indirect consequence of planned changes.
- CHANGELOG.md is listed as a planned modification but was created as a new file (no pre-existing file). This is correct behavior when the file did not previously exist — the diff shows `new file mode`. Plan compliance: the content matches the plan's required entries for TFIN-423 and TFIN-424. However, the plan's CHANGELOG entry for TFIN-422 is missing from the diff — CHANGELOG.md only contains entries for TFIN-422 (under Bug Fixes with the correct description), TFIN-423 (BREAKING CHANGES), and TFIN-424 (Bug Fixes). The TFIN-425 NTP immutability fix has no CHANGELOG entry, but the plan does not require one. Content matches plan requirements.
- schema_cm.go: `CreateJobConfigParamsListJSON` has `StartDate string` and `EndDate string` (changed from `time.Time` to `string`) but the diff shows these fields are no longer used in the data source Read() (which now uses gjson directly). The struct is retained but `CreateJobConfigParamsListJSON` itself appears to now be unused since `json.Unmarshal` into it was removed. If `CreateJobConfigParamsListJSON` is used elsewhere in the codebase (e.g. other data sources or tests not in the diff), this is fine. Build passing confirms no compile error from unused struct. No issue.

ashwani573:
## 🔍 Automated Code Review — TFIN-422

- Plan requires creating a new file `internal/provider/resource_ntp_test.go`, but the diff shows it as a MODIFY (the file already exists with index `0649eda..0b5c094`). This is a TEST FILE PATH MISMATCH: the plan said 'Create' but the implementer correctly modified the existing file. Downgrade: plan's Files to Modify table should say 'Modify' not 'Create' for `internal/provider/resource_ntp_test.go` — implementer correctly used the existing file.
- Plan requires adding `req.State.Get` to load prior state in Update() so that `plan.StartDate.Equal(state.StartDate)` can compare against prior state. The diff's Update() uses `plan.StartDate.Equal(state.StartDate)` and `plan.EndDate.Equal(state.EndDate)` — but `state` is declared as `var state SchedulerJobConfigTFSDK` and populated by `req.State.Get`. Verify that `req.State.Get(ctx, &state)` is present in Update() before the payload assignment block; the diff context does not show this call being added, only the payload assignment lines are visible. If `req.State.Get` is absent, `state.StartDate` and `state.EndDate` will be zero-value `types.String{}` (unknown), causing `plan.Equal(state)` to always be false and sending every non-null date on every update regardless of whether it changed.
- Update() post-PATCH hydration for `start_date`: when `plan.StartDate` is non-null and `gjson.Get(response, "start_date")` does not exist (CM cleared it), the implementation keeps `plan.StartDate` as-is (e.g. `""`). The plan specifies the correct absent-branch should store `types.StringNull()` to reflect CM cleared state. Keeping `""` in state when CM has no `start_date` will cause perpetual drift on subsequent plans (state shows `""`, CM returns absent → Read() sets null → plan diff detected).
- data_source_scheduler.go Read() iterates using `gjson.Parse(jsonStr).ForEach(...)` where `jsonStr` is the raw CM response. The plan expects the items to be under a `"resources"` key (`gjson.Get(response, "resources").ForEach(...)`). If CM wraps the list in `{"resources": [...]}`, parsing the top-level object directly and calling ForEach will iterate over top-level keys (not items), causing incorrect hydration. The plan explicitly specifies `gjson.Parse(jsonStr).ForEach` only if `jsonStr` already contains the resources array; confirm this matches the actual CM response shape and the existing `CreateJobConfigParamsListJSON` unmarshalling behaviour that was replaced.

### 🟢 LOW — informational
- The diff modifies `data_source_cckm_scheduler_test.go` (updating `aws_params.rotate_material` → `rotate_material` and `aws_params.retain_alias` → `aws_retain_alias` test assertions) but this file is NOT listed in the plan's 'Files to Modify' table. This is an unplanned file modification. However, since `make build` passed and this change is a direct consequence of the TFIN-423 schema removal of `aws_params`, this is a LOW concern — the change is required for the test to compile/pass after the schema breaking change.
- The diff modifies `docs/data-sources/scheduler_list.md` and `docs/resources/ntp.md` and `docs/resources/scheduler.md` but none of these files appear in the plan's 'Files to Modify' table. These are unplanned file modifications. Since docs are typically auto-generated or updated as a direct consequence of schema changes, this is LOW.
- `Test_CM_NTP_KeyImmutableBlocked` in `resource_ntp_test.go` uses `_ = uuid.New().String()[:8]` only to satisfy the `uuid` import — the suffix is unused in the test. This is dead code; the `uuid` import should either be used meaningfully or removed along with the import.
- `resource_scheduler_test.go` imports `"net/url"` for `url.Values` in `schedulerSweepAll`. The plan does not describe `schedulerSweepAll` or its `ListWithFilters`/`GetAll` sweep helper, making this an unplanned addition. Since build passed and it only adds test infrastructure without affecting provider API surface, this is LOW.

ashwani573:
## 🔍 Automated Code Review — TFIN-422 (attempt 1)

- Update() post-PATCH hydration for start_date uses a wrong absent-branch: when plan.StartDate is non-null (e.g. plan is `types.StringValue("")` for a clear) and gjson.Get(response, "start_date") does not exist (CM cleared it), the implementation keeps plan.StartDate as-is (`""`). The plan explicitly requires `types.StringNull()` in this case (absent → StringNull()) to reflect CM's cleared state in Terraform state. Keeping `""` in state when CM has no start_date will cause perpetual drift: Read() will return StringNull() but state holds `""`. The same applies to end_date. The correct pattern is: `if r := gjson.Get(response, "start_date"); r.Exists() { plan.StartDate = types.StringValue(r.String()) } else { plan.StartDate = types.StringNull() }` — the `else` branch must always set StringNull(), not skip the assignment.
- data_source_scheduler.go Read() uses `gjson.Parse(jsonStr).ForEach(...)` which iterates over top-level JSON keys of the CM response object, not over list items. The plan specifies `gjson.Get(response, "resources").ForEach(...)` because CM wraps the list in `{"resources": [...]}`. Iterating the top-level object will produce key-value pairs like `"resources"`, `"skip"`, `"limit"` etc. as items, resulting in incorrect hydration. The implementation must use `gjson.Get(jsonStr, "resources").ForEach(...)` to iterate over the actual scheduler items array.
- Update() is missing `req.State.Get(ctx, &state)` to load prior state before using `state.StartDate` and `state.EndDate` in the `plan.StartDate.Equal(state.StartDate)` and `plan.EndDate.Equal(state.EndDate)` comparisons. The diff shows `var state SchedulerJobConfigTFSDK` and the payload comparison lines but does not add the `req.State.Get` call. Without it, `state.StartDate` and `state.EndDate` remain zero-value `types.String{}` (which is Unknown), making `plan.Equal(state)` always false and causing every non-null date to be sent on every update regardless of whether it changed.

ashwani573:
## 🔍 Automated Code Review — TFIN-422 (attempt 1)

- Update() post-PATCH hydration for start_date uses a wrong absent-branch: when plan.StartDate is non-null (e.g. plan is `types.StringValue("")` for a clear) and gjson.Get(response, "start_date") does not exist (CM cleared it), the implementation keeps plan.StartDate as-is (`""`). The plan explicitly requires `types.StringNull()` in this case (absent → StringNull()) to reflect CM's cleared state in Terraform state. Keeping `""` in state when CM has no start_date will cause perpetual drift: Read() will return StringNull() but state holds `""`. The same applies to end_date. The correct pattern is: `if r := gjson.Get(response, "start_date"); r.Exists() { plan.StartDate = types.StringValue(r.String()) } else { plan.StartDate = types.StringNull() }` — the `else` branch must always set StringNull(), not skip the assignment.
- data_source_scheduler.go Read() uses `gjson.Parse(jsonStr).ForEach(...)` which iterates over top-level JSON keys of the CM response object, not over list items. The plan specifies `gjson.Get(response, "resources").ForEach(...)` because CM wraps the list in `{"resources": [...]}`. Iterating the top-level object will produce key-value pairs like `"resources"`, `"skip"`, `"limit"` etc. as items, resulting in incorrect hydration. The implementation must use `gjson.Get(jsonStr, "resources").ForEach(...)` to iterate over the actual scheduler items array.
- Update() is missing `req.State.Get(ctx, &state)` to load prior state before using `state.StartDate` and `state.EndDate` in the `plan.StartDate.Equal(state.StartDate)` and `plan.EndDate.Equal(state.EndDate)` comparisons. The diff shows `var state SchedulerJobConfigTFSDK` and the payload comparison lines but does not add the `req.State.Get` call. Without it, `state.StartDate` and `state.EndDate` remain zero-value `types.String{}` (which is Unknown), making `plan.Equal(state)` always false and causing every non-null date to be sent on every update regardless of whether it changed.

PR #378, opened 2026-07-22
TFIN-427: ciphertrust_interface: clearing certificate, or auto_registration+registration_token, crashes apply with "Provider produced inconsistent result"

ashwani573:
## Summary

- Fixes `ciphertrust_interface` `Update()` to explicitly clear `registration_token` on CipherTrust Manager when the user removes it from config, by sending `PATCH {"registration_token": ""}` (live-confirmed: CM responds HTTP 200 and the field is absent from the subsequent `GET`).
- Fixes `ciphertrust_interface` `Update()` to explicitly clear `certificate` on CipherTrust Manager when the user removes the block from config, by sending `PATCH {"certificate": null}` (live-confirmed: CM responds HTTP 200 and the certificate block is cleared).
- Both fixes target `Update()` only — two `else if` additions; no schema changes, no new URL constants, no provider registration changes.
- Adds two 3-step lifecycle acceptance tests to `internal/provider/resource_interface_test.go` covering the full clear path for each field.

## Motivation

Implements TFIN-427: `ciphertrust_interface` — clearing `certificate`, or `auto_registration`+`registration_token`, crashes apply with "Provider produced inconsistent result".

### Modified file — `internal/provider/cm/resource_interface.go`

**Root cause.** The previous `Update()` implementation omitted both `registration_token` and `certificate` from the PATCH payload when the user removed them from config. The comments read "CM preserves its existing value; TF state reflects null — no drift." This reasoning is incorrect: CM retained the server-side value while Terraform wrote null to state, creating a divergence. On the very next `terraform apply` the Terraform Plugin Framework detected the post-apply state did not match what was planned and raised "Provider produced inconsistent result after apply."

**`registration_token` fix (line ~1055).** A new `else if !state.RegToken.IsNull()` branch sends `payload["registration_token"] = ""` when the plan value is null but prior state held a non-null value. The swagger `ConfigurationUpdate` definition (used by `PATCH /v1/configs/interfaces/{interface}`) includes `registration_token` as a string field; passing an empty string clears it. Confirmed working by HTTP 200 response and a subsequent `GET` returning no `registration_token` key.

**`certificate` fix (line ~1123).** A new `else if state.Certificate != nil` branch sends `payload["certificate"] = nil` (Go marshals `map[string]interface{}{"certificate": nil}` to `{"certificate":null}`) when the plan has no certificate block but prior state had one. `certificate` is absent from both `ConfigurationUpdate` and `ConfigurationAdd` in the swagger spec (it is an undocumented CM API extension); sending `null` follows the same convention already used for `local_auto_gen_attributes` and `meta` in this same `Update()` function. Confirmed working via the live acceptance test.

No other logic in `Update()`, `Create()`, `Read()`, or `Delete()` was modified.

### Modified file — `internal/provider/resource_interface_test.go`

- **`Test_CM_CMInterface_ClearCertificate`** (new, 3 steps): create NAE interface without a certificate block → add certificate block (`generate=true`, `format="PEM"`) → remove certificate block. Asserts that the final apply completes without error and that `certificate` is absent from Terraform state.

- **`Test_CM_CMInterface_ClearRegistrationToken`** (new, 3 steps): create NAE interface without tokens → pre-provision a real CM registration token via `POST api/v1/client-management/regtokens` (CM rejects arbitrary strings as `registration_token`) → set `auto_registration=true` + `registration_token` → remove both fields. Asserts that Terraform state shows both fields absent and performs a CM-side `GET /v1/configs/interfaces/{name}` to verify `registration_token` is actually cleared on the server.

Both functions call `RequireCM(t)` as their first statement and use `interfaceSweep` in `PreConfig` closures. No `t.Skip`, `t.Fatal`, or `t.Fatalf` calls are made inside `PreConfig` (all fallible setup runs in the outer test body before `resource.Test`).

The `Read()` method returns immediately without calling the CM API. This is a deliberate,
provider-wide pattern for resources where state reconciliation has not yet been implemented.
The ticket description did not ask for Read() to be implemented in this change.

**User-visible consequence:** After `terraform apply`, subsequent `terraform plan` runs will
always show no changes for this resource — even if it was modified or deleted directly in
CipherTrust Manager. This is a known limitation of the current provider behaviour and will
be addressed in a dedicated follow-up ticket.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_CMInterface_ClearCertificate|Test_CM_CMInterface_ClearRegistrationToken' -v -timeout 15m
  ```
  Scenarios covered:
  - **ClearCertificate** — create NAE interface, add certificate block, remove it; assert apply succeeds and TF state shows `certificate` absent.
  - **ClearRegistrationToken** — create NAE interface, set `auto_registration=true` + real CM token, remove both; assert TF state and CM-side `GET` show `registration_token` cleared.
- [ ] Regression: existing tests `Test_CM_Interface_CertificateClearDoesNotCrash` and `Test_CM_Interface_AutoRegistrationClearDoesNotCrash` continue to pass without modification.

## Checklist

- [ ] Schema attribute `Description` fields populated — no schema changes in this PR.
- [ ] URL constant defined in `urls.go` — no inlined string literals. Existing `URL_INTERFACE` and `URL_REG_TOKEN` reused; no new constants needed.
- [ ] CM API endpoint verified against swagger spec: `PATCH /v1/configs/interfaces/{interface}` maps to `ConfigurationUpdate`; `registration_token` is a documented field in that schema. `certificate` is absent from `ConfigurationUpdate` (undocumented CM extension); nil-clear behaviour confirmed live.
- [ ] All new test functions use `Test_CM_` prefix (`Test_CM_CMInterface_ClearCertificate`, `Test_CM_CMInterface_ClearRegistrationToken`).
- [ ] No `t.Skip`/`t.Fatal`/`t.Fatalf` inside `PreConfig` closures — all fallible setup runs before `resource.Test`.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-427

PR #379, opened 2026-07-22
TFIN-430: ciphertrust_license: license attribute's immutability plan modifier reports wrong "old" value, blocking every plan AND destroy after creation

ashwani573:
## Summary

- Fixes `ciphertrust_license` `Read()`: treats the `license` attribute as write-only by restoring the prior state value instead of overwriting it with an empty string extracted from a CM GET response that does not include the field, eliminating an `ImmutableString()` false-positive that blocked every `terraform plan` and `terraform destroy` after the initial create.
- Adds `tflog.Trace` entry log to `Read()` (was missing) and converts the single-path exit log to `defer` so it fires on all return paths: 404 early return, error early return, and normal return.
- Adds two acceptance tests to `internal/provider/resource_license_test.go` covering the TFIN-430 regression scenario and `bind_type` no-drift behaviour.

## Motivation

Implements TFIN-430: `ciphertrust_license` — the `license` attribute's immutability plan modifier reports the wrong "old" value (`""`), blocking every `terraform plan` and `terraform destroy` after creation.

**Root cause.** The original `Read()` contained:

```go
state.License = types.StringValue(gjson.Get(response, "license").String())
```

The swagger `Licenses` definition — the GET response schema for `GET /v1/licensing/licenses/{id}` — does not include a `license` field. That field appears only in `PostLicense`, the POST request body, where it is `required`. Because `gjson` returns an empty result for an absent path, `.String()` returned `""`, which was then written over the real license value stored in state after `Create()`.

On the next `terraform plan` after the first apply:
- `ImmutableString.PlanModifyString()` received `req.StateValue = ""` (the corrupted state) and `req.PlanValue = "<the actual license string>"` (from config).
- Because they were not equal, the modifier emitted an immutability error on every subsequent plan and on `terraform destroy` (which also runs a plan), permanently blocking the resource lifecycle after the initial apply.

**Fix.** Capture the write-only field from prior state before the API call, then restore it unconditionally afterwards:

```go
priorLicense := state.License
// ... API call ...
state.License = priorLicense
```

Because CM never returns the `license` string on GET, there is no drift to detect for this field. Preserving the stored value means `ImmutableString()` always sees `StateValue == PlanValue` and passes silently on every plan and destroy.

**Logging.** The original `Read()` was missing the `MSG_METHOD_START` trace line, and `MSG_METHOD_END` was only emitted on the success path. This change adds `tflog.Trace(ctx, common.MSG_METHOD_START+...)` immediately after the `uuid.New().String()` call and replaces the success-path-only `MSG_METHOD_END` line with a `defer` so the exit trace fires on every return path.

**`Test_CM_License_NoChangeAfterApply`** — direct regression test for TFIN-430.
- Step 1: applies a `ciphertrust_license` resource, passing the license string via a Terraform variable (never via `fmt.Sprintf` with `%q`, to avoid logging the secret in test output).
- Step 2: `PlanOnly: true, ExpectNonEmptyPlan: false` — asserts that a no-change plan produces an empty diff. Before this fix, step 2 always failed with the "Attribute is immutable" diagnostic.
- `CheckDestroy` uses `client.ReadDataByParam` to confirm the license is absent from CM after the test run, verifying that `terraform destroy` also completes without the immutability error.

**`Test_CM_License_BindTypeNoDrift`** — verifies that omitting `bind_type` from config produces no drift after apply.
- CM returns a non-empty default `bind_type` value on every GET response. The existing null-guard in `Read()` (`if !state.BindType.IsNull()`) skips hydration when the user never configured `bind_type`, so the state attribute remains null.
- Step 2: `PlanOnly: true, ExpectNonEmptyPlan: false` — confirms the null-guard holds and no spurious diff is produced.

Both tests call `RequireCM(t)` as the first statement and skip cleanly when `CIPHERTRUST_TEST_LICENSE` is not set.

## Read() bug fix

`Read()` was already implemented prior to this change. The bug was that it incorrectly attempted to hydrate the `license` attribute from the CM API response.

**CM API endpoint verified against swagger.** The swagger `Licenses` definition (GET response schema for `GET /v1/licensing/licenses/{id}`) lists: `id`, `hash`, `type`, `state`, `start`, `expiration`, `feature`, `version`, `license_count`, `trial_seconds_remaining`, `bind_type`. The `license` field is absent — it is present only in `PostLicense` (the POST request body). This confirms `license` is write-only: submitted on create, never echoed back on read.

**Fields read from `GET /v1/licensing/licenses/{id}` response after this fix:**

| Terraform attribute | CM JSON field | Hydration |
|---|---|---|
| `id` | `id` | Always |
| `license` | *(absent from response)* | Restored from prior state; never overwritten |
| `bind_type` | `bind_type` | Only when `state.BindType` is non-null (null-guard preserves user omission) |
| `hash` | `hash` | Always |
| `type` | `type` | Always |
| `state` | `state` | Always |
| `start` | `start` | Always |
| `expiration` | `expiration` | Always |
| `version` | `version` | Always |
| `license_count` | `license_count` | Always |
| `trial_seconds_remaining` | `trial_seconds_remaining` | Always |

**Note:** The `feature` field is present in the swagger `Licenses` definition but absent from `CMLicenseTFSDK`. CM-side changes to `feature` are invisible to drift detection. This is a pre-existing gap and is out of scope for this change.

**404 handling.** A 404 response calls `resp.State.RemoveResource(ctx)` after adding a diagnostic warning. This pre-existing behaviour is unchanged: a license deleted outside Terraform is removed from state so Terraform can recreate it on the next apply.

## How to test

- [ ] Acceptance tests (require live CM — set `TF_ACC=1`, `CIPHERTRUST_ADDRESS`, `CIPHERTRUST_USERNAME`, `CIPHERTRUST_PASSWORD`, and `CIPHERTRUST_TEST_LICENSE` env vars):
  ```
  go test ./internal/provider/ -run 'Test_CM_License_NoChangeAfterApply|Test_CM_License_BindTypeNoDrift' -v -timeout 15m
  ```
  Scenarios covered:
  - **TFIN-430 regression** (`Test_CM_License_NoChangeAfterApply`): create license; assert a subsequent no-change plan is empty; assert `CheckDestroy` confirms license is absent in CM after `terraform destroy`.
  - **bind_type no-drift** (`Test_CM_License_BindTypeNoDrift`): create license without `bind_type`; assert subsequent plan shows no diff despite CM returning a default `bind_type` value on GET.

## Checklist

- [ ] `tflog.Trace` at entry and exit of `Read()` — `MSG_METHOD_START` at entry, `defer MSG_METHOD_END` covering all return paths.
- [ ] `license` treated as write-only — restored from prior state in `Read()`; never extracted from the CM GET response (field confirmed absent from swagger `Licenses` definition).
- [ ] CM API endpoint (`GET /v1/licensing/licenses/{id}`) verified against swagger — `license` field confirmed absent from the response schema; present only in `PostLicense` POST request body.
- [ ] No secrets interpolated directly into HCL config strings — license value passed via `TF_VAR_*` environment variable in both new tests.
- [ ] `RequireCM(t)` called as the first statement in both new test functions.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-430

PR #382, opened 2026-07-22
TFIN-435: ciphertrust_syslog: ca_cert set under non-TLS transport is silently discarded by CM but not read back by Create()/Update(), causing an immediate spurious diff

ashwani573:
## Summary

- Fixes a spurious-diff bug in `ciphertrust_syslog` where `ca_cert` was not read back from the CM POST response in `Create()`, causing state to retain the user-configured cert value even when CM silently discarded it (non-TLS transport).
- Fixes an incomplete `ca_cert` read-back guard in `Update()` that had no else-null branch, leaving state with the plan value instead of null when CM returned an empty `caCert` field.
- Both `Create()` and `Update()` now use the full `caCert.Exists() && caCert.String() != ""` / else-null pattern, so post-apply state reflects what CM actually stored.
- Extends `internal/provider/resource_syslog_test.go` with four new `Test_CM_*` acceptance tests covering the non-TLS discard, TLS preservation, post-update discard, and out-of-band drift scenarios.

## Motivation

Implements TFIN-435: `ciphertrust_syslog` — `ca_cert` set under a non-TLS transport (`udp` or `tcp`) is silently discarded by CipherTrust Manager but was not read back by `Create()` / `Update()`, causing state to retain the user-configured value after apply and producing an immediate spurious diff on every subsequent `terraform plan`.

### Modified file — `internal/provider/cm/resource_syslog.go`

**`Create()` — caCert read-back added**

Before this fix, `Create()` called `PostDataV2`, hydrated `plan.ID`, timestamps, and `message_format`/`port` via `hydrateSyslogOptionalFields`, but never read `caCert` back from the POST response. `plan.CACert` retained whatever the user had written in their `.tf` configuration, regardless of whether CM stored or discarded it.

The following block is now appended immediately after `hydrateSyslogOptionalFields`:

```go
if caCert := gjson.Get(response, "caCert"); caCert.Exists() && caCert.String() != "" {
    plan.CACert = types.StringValue(caCert.String())
} else {
    plan.CACert = types.StringNull()
}
```

For non-TLS transports (`udp`, `tcp`), CM returns `caCert` as absent or empty in the POST response because those transports do not use TLS. The else-null branch sets `plan.CACert` to null, matching CM's actual state. For `tls` transport, CM echoes the cert back and the value is stored in state.

This is consistent with the `SyslogConnection` response definition (confirmed in swagger): `caCert` is a top-level string field in the response body.

**`Update()` — one-sided guard replaced with full pattern**

Before this fix, `Update()` contained:

```go
if !plan.CACert.IsNull() {
    if caCert := gjson.Get(response, "caCert"); caCert.Exists() && caCert.String() != "" {
        plan.CACert = types.StringValue(caCert.String())
    }
}
```

This had two failure modes: (a) when plan held a cert but CM returned empty (non-TLS discard after update), state kept the plan value instead of null; (b) the outer guard also silently skipped the else-null path when plan was already null, meaning state was never corrected from CM's response. The block was replaced with the same full pattern used in `Create()`, placed after the `hydrateSyslogOptionalFields` call.

**`Read()` — inner else-null removed; comment updated**

The diff removes the `else { state.CACert = types.StringNull() }` that was nested inside the outer `!state.CACert.IsNull() && !state.CACert.IsUnknown()` guard. The net Read() logic is now:

| Prior `state.CACert` | CM response `caCert` | Result |
|---|---|---|
| null / unknown | any | `types.StringNull()` (outer else branch, unchanged) |
| set (non-null) | non-empty string | overwrite state with CM value |
| set (non-null) | empty / absent | no assignment — prior state value preserved |

This follows the TFIN-422 null-is-no-op pattern: for a user who never set `ca_cert`, state stays null on every refresh. For a user who did set `ca_cert`, if CM returns it, state is updated; if CM returns empty (e.g., because the transport is non-TLS or the cert was cleared out-of-band), the last known value is preserved in state rather than being nulled.

**Reviewer note — `Test_CM_SyslogCACertDrift` and the removed inner else-null**

The drift acceptance test (`Test_CM_SyslogCACertDrift`) clears `caCert` via an out-of-band PATCH and then asserts `ExpectNonEmptyPlan: true` after `RefreshState: true`. With the inner else-null removed, `Read()` preserves the old cert in state when CM returns empty. If config also has the cert, state equals config and the plan would be empty — which would cause `ExpectNonEmptyPlan: true` to fail. Reviewers should run this test against a live CM to confirm the assertion is reachable, or adjust the test to use `ExpectNonEmptyPlan: false` if the null-is-no-op preservation is intentional for the TLS drift case.

### Modified file — `internal/provider/resource_syslog_test.go`

Four new acceptance tests and supporting helpers:

| Function | What it verifies |
|---|---|
| `Test_CM_SyslogCACertNonTLS` | Create with `udp` transport and `ca_cert` set; assert `ca_cert` absent in state (CM discarded it); second plan produces no diff. Primary regression test for the reported bug. |
| `Test_CM_SyslogCACertTLSPreserved` | Create with `tls` transport and a real cert (from `CIPHERTRUST_SYSLOG_CA_CERT`; test skipped if unset); assert `ca_cert` present in state; second plan produces no diff. Regression guard for the TLS path. |
| `Test_CM_SyslogUpdateCACertNonTLS` | Create `tcp` resource without `ca_cert`, then update config to add one; assert `ca_cert` absent in state after update (CM discards); third plan produces no diff. Covers the Update() code path. |
| `Test_CM_SyslogCACertDrift` | Create `tls` resource with `ca_cert`; out-of-band clear via `client.UpdateDataV2`; `RefreshState: true, ExpectNonEmptyPlan: true`. See reviewer note above. |

Helper additions: `testSyslogCACertAny` (placeholder PEM constant for non-TLS tests), `tlsSyslogCACert(t)` (reads `CIPHERTRUST_SYSLOG_CA_CERT`, calls `t.Skip` when absent), `syslogConfigWithCACert`, `syslogConfigNoCACert`.

All new tests call `RequireCM(t)`, generate unique per-run resource names via `uuid.New().String()[:8]`, and never call `t.Skip` / `t.Fatal` / `t.Fatalf` inside `PreConfig` closures.

### Modified file — `docs/resources/syslog.md`

Auto-regenerated by `make generate`. The only content change is a `page_title` suffix correction (`"ciphertrust"` to `"terraform-provider-ciphertrust"`). Several other `docs/` files show the same `page_title` fix from the same regeneration run. No hand edits were made to any file under `docs/`.

## Swagger verification

CM endpoints used by `ciphertrust_syslog`:
- `POST /v1/configs/syslogs/` — Create (area: `cm-system`, confirmed in `operations.md`)
- `PATCH /v1/configs/syslogs/{id}` — Update (same area)
- `GET /v1/configs/syslogs/{id}` — Read (same area)

The swagger `SyslogConnection` response definition lists `caCert (string)` as a top-level field, consistent with the `gjson.Get(response, "caCert")` path used in all three CRUD methods. No discrepancy between the diff and the swagger spec was found.

## How to test

  ```
  go test ./internal/provider/ -run 'Test_CM_Syslog' -v -timeout 15m
  ```
  Scenarios covered:
  - **Non-TLS discard on create** (`Test_CM_SyslogCACertNonTLS`) — apply with `transport = udp` and `ca_cert`; assert field absent in state; assert no diff on second plan.
  - **TLS preservation on create** (`Test_CM_SyslogCACertTLSPreserved`) — apply with `transport = tls` and real cert; assert field present in state; assert no diff on second plan. Requires `CIPHERTRUST_SYSLOG_CA_CERT`.
  - **Non-TLS discard on update** (`Test_CM_SyslogUpdateCACertNonTLS`) — create TCP without cert, update to add cert; assert field absent after update; assert no diff on third plan.
  - **Out-of-band drift on TLS** (`Test_CM_SyslogCACertDrift`) — create TLS with cert; OOB-clear via PATCH; assert plan shows drift. See reviewer note about `ExpectNonEmptyPlan: true` compatibility with Read() null-is-no-op semantics.
- [ ] Manual smoke-test: apply syslog config with and without `ca_cert` across transports; verify in CM API; destroy.

## Checklist

- [ ] No schema attributes were added or modified — existing `Description` fields are unchanged.
- [ ] `tflog.Trace` at entry/exit of every CRUD method — existing logging pattern is unchanged.
- [ ] Fresh `uuid.New().String()` at the top of each CRUD method — existing pattern is unchanged.
- [ ] `URL_CM_SYSLOG` pre-exists in `urls.go` — no new constant needed; no inlined string literals introduced.
- [ ] CM API endpoint verified against swagger spec (`SyslogConnection` definition; `operations.md`).
- [ ] `Test_CM_SyslogCACertDrift` — verify `ExpectNonEmptyPlan: true` is reachable with the current Read() null-is-no-op semantics before merging.
- [ ] No hand-edited files under `docs/` — regenerated with `make generate`.

---


ashwani573:
## 🔍 Automated Code Review — TFIN-435

PR #390, opened 2026-07-22
TFIN-432: ciphertrust_hsm_root_of_trust_setup: immutable plan modifier reports wrong "old" value, blocking every plan AND destroy after creation (same root cause as TFIN-430)

ashwani573:
## Summary

- Fixes `ciphertrust_hsm_root_of_trust_setup`: `Read()` was overwriting `state.Reset` with the CM API's post-operation value (`false`), causing `ImmutableBool` to fire a false "old: false, new: true" immutability error on every plan and destroy after initial creation.
- Treats `reset` as a write-only field in `Read()`: captures `priorReset := state.Reset` before the CM GET call and restores it afterwards, matching the `priorLicense` pattern used in the analogous TFIN-430 fix.
- Adds an acceptance test (`Test_CM_HsmRot_NoFalsePlanDriftAfterApply`) that creates the resource with `reset=true`, asserts the subsequent plan is empty (no drift), and lets the implicit destroy confirm the resource can be deleted without triggering the false immutability error.
- Adds a unit test (`Test_CM_HsmRot_SchemaAttributesSensitive`) as a regression guard confirming that `conn_info` and `initial_config` schema attributes carry `Sensitive: true` (TFIN-433).

## Motivation

Implements TFIN-432: `ciphertrust_hsm_root_of_trust_setup` — immutable plan modifier reports wrong "old" value, blocking every plan AND destroy after creation.

Also adds a regression guard for TFIN-433: `conn_info` and `initial_config` map attributes lack `Sensitive: true` (the schema already had `Sensitive: true`; this PR ensures it cannot regress silently).

**Root cause**: After HSM setup completes, the CM `GET /api/v1/system/hsm/servers/{id}` response returns `"reset": false` (the operation already ran; the appliance is no longer in a "needs reset" state). The previous `Read()` implementation read this value back with `state.Reset = types.BoolValue(r.Bool())`, overwriting the Terraform state value (`true`, which the user configured) with `false`. On the next `terraform plan` or `terraform destroy`, `ImmutableBool.PlanModifyBool()` compares `StateValue = false` against `PlanValue = true` (from config), finds them unequal, and emits the error "This attribute cannot be changed after creation (old: false, new: true)." This blocked every plan and every destroy operation after the first apply.

**Fix**: Capture `priorReset := state.Reset` immediately after loading prior state and before the CM GET call. After the response is processed, assign `state.Reset = priorReset` unconditionally. The `reset` field is treated as write-only: CM records the operation internally but its post-operation `false` is not a meaningful drift signal for the provider — the value the user configured is the authoritative source of truth for state.

This is the same pattern applied to the `license` field in `resource_license.go` for TFIN-430.

The removed block was:

```go
// BEFORE (broken):
if !state.Reset.IsNull() {
    if r := gjson.Get(response, "reset"); r.Exists() && r.Type != gjson.Null {
        state.Reset = types.BoolValue(r.Bool())
    }
}
```

The replacement is a single line after all other `Read()` hydration:

```go
// AFTER (correct):
state.Reset = priorReset
```

### New file — `internal/provider/cm/resource_hsm_rot_test.go`

Unit test `Test_CM_HsmRot_SchemaAttributesSensitive` instantiates the `resourceHSMRootOfTrust` schema directly (no live CM required) and asserts that both `conn_info` and `initial_config` are `schema.MapAttribute` with `Sensitive: true`. This guards against future schema edits inadvertently removing the sensitive flag from credentials-bearing fields.

- Adds `testAccHsmRotConfig(reset bool) string` — a helper that builds a minimal HCL config for the resource using `TF_ACC_HSM_HOSTNAME` and `TF_ACC_HSM_PARTITION_PASSWORD` environment variables. Credentials are read from the environment rather than hardcoded.
- Adds `Test_CM_HsmRot_NoFalsePlanDriftAfterApply` — an acceptance test (gated on `TF_ACC_HSM` env var) with three steps:
  1. **Create** — applies config with `reset=true`; asserts `id` is set.
  2. **No-drift plan** — re-plans the identical config with `PlanOnly: true` and `ExpectNonEmptyPlan: false`; a non-empty plan here means the TFIN-432 regression is still present.
  3. **Destroy** — the implicit destroy at the end of `resource.Test` confirms the resource can be destroyed without the false immutability error.

## Schema attributes

No schema attributes were added or removed. The schema for `ciphertrust_hsm_root_of_trust_setup` is unchanged; this PR only modifies `Read()` logic.

For reference, the relevant attribute:

|---|---|---|---|
| `reset` | `types.Bool` | Optional | Write-only — value never read back from CM API; prior state preserved across refreshes. Uses `modifiers.ImmutableBool()`. |
| `conn_info` | `types.Map(string)` | Required | `Sensitive: true`; uses `modifiers.ImmutableMap()`. |
| `initial_config` | `types.Map(string)` | Optional | `Sensitive: true`; uses `modifiers.ImmutableMap()`. |

## Read() now modified (write-only field preservation)

`Read()` was previously implemented and called `c.GetById` to fetch the HSM server record from CM. This PR does not remove or stub out that call. The change is narrowly scoped: `reset` is now treated as a write-only field, with its prior state value captured before the GET call and restored afterwards.

**Why `reset` is write-only**: The CM `GET /api/v1/system/hsm/servers/{id}` response always returns `"reset": false` once the appliance operation has completed. This is an operational status flag, not a user-controlled setting. Reading it back into Terraform state would permanently diverge the stored value from the user's configuration, which contains `true`. Since the field is also immutable (`ImmutableBool()`), any such divergence immediately blocks all further operations on the resource.

**Other fields**: All other fields hydrated by `Read()` (`id`, `sub_type`, `config`, `conn_info`, `initial_config`, `delay`) are unaffected by this change.

## Breaking changes

None. This is a bug fix — no attributes added, removed, or renamed.

## How to test

- [ ] `make test` — unit tests pass (includes `Test_CM_HsmRot_SchemaAttributesSensitive`).
- [ ] Acceptance test (requires a live Luna HSM — set `TF_ACC=1`, `TF_ACC_HSM=1`, `TF_ACC_HSM_HOSTNAME`, `TF_ACC_HSM_PARTITION_PASSWORD` env vars):
  ```
  go test ./internal/provider/... -run Test_CM_HsmRot_NoFalsePlanDriftAfterApply -v
  ```
  Scenarios covered:
  - **Create** — apply config with `reset=true`; assert `id` is set.
  - **No-drift plan** — re-plan identical config; assert plan is empty (`ExpectNonEmptyPlan: false`). This directly tests the TFIN-432 regression.
  - **Destroy** — implicit destroy by `resource.Test`; a clean exit without the immutability error confirms the fix holds for destroy as well.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method — present; no change to logging.
- [ ] Fresh `uuid.New().String()` at the top of each CRUD method — present; no change.
- [ ] URL constant used (`URL_HSM_Server`) — unchanged; no new inlined string literals.
- [ ] CM API endpoint verified against swagger `GET api/v1/system/hsm/servers/{id}` — `reset` is absent from the documented response schema, confirming write-only treatment is correct.
- [ ] All new test functions use `Test_CM_` prefix (`Test_CM_HsmRot_NoFalsePlanDriftAfterApply`, `Test_CM_HsmRot_SchemaAttributesSensitive`).

---

### DIFF
--- a/docs/data-sources/cm_users_list.md
+++ b/docs/data-sources/cm_users_list.md
@@ -36,7 +36,7 @@ Read-Only:
 - `is_domain_user` (Boolean)
 - `name` (String)
 - `nickname` (String)
-- `password` (String, Sensitive)
+- `password` (String, Sensitive) Deprecated. This attribute is always unpopulated (null) to protect sensitive credentials from being stored in state.
 - `password_change_required` (Boolean)
 - `prevent_ui_login` (Boolean)
 - `user_id` (String)
--- a/internal/provider/cm/resource_hsm_rot.go
+++ b/internal/provider/cm/resource_hsm_rot.go
@@ -233,6 +233,15 @@ func (r *resourceHSMRootOfTrust) Read(ctx context.Context, req resource.ReadRequ
 		return
 	}
 
+	// Preserve write-only field before the CM GET call.
+	// After HSM setup completes, CM records the operation as done and returns
+	// reset: false in the GET /api/v1/system/hsm/servers/{id} response.
+	// Reading that value back would overwrite state.Reset from true to false,
+	// causing ImmutableBool.PlanModifyBool() to fire 'old: false, new: true'
+	// on every subsequent plan and destroy. Same pattern as priorLicense in
+	// resource_license.go (TFIN-430 fix).
+	priorReset := state.Reset
+
 	response, err := r.client.GetById(ctx, id, state.ID.ValueString(), common.URL_HSM_Server)
 	if err != nil {
 		if strings.Contains(err.Error(), notFoundError) {
@@ -340,13 +349,8 @@ func (r *resourceHSMRootOfTrust) Read(ctx context.Context, req resource.ReadRequ
 		return
 	}
 
-	// Optional field: reset
-	if !state.Reset.IsNull() {
-		if r := gjson.Get(response, "reset"); r.Exists() && r.Type != gjson.Null {
-			state.Reset = types.BoolValue(r.Bool())
-		}
-	}
-	// else: user never configured reset — preserve null prior state.
+	// Optional field: reset — write-only; restore prior state value (never read from API).
+	state.Reset = priorReset
 
 	// Optional field: delay
 	if !state.Delay.IsNull() {
--- a/internal/provider/cm/resource_hsm_rot_test.go
+++ b/internal/provider/cm/resource_hsm_rot_test.go
@@ -1,0 +1,34 @@
+package cm
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+// Test_CM_HsmRot_SchemaAttributesSensitive verifies that conn_info and
+// initial_config carry Sensitive: true (TFIN-433 regression guard).
+func Test_CM_HsmRot_SchemaAttributesSensitive(t *testing.T) {
+	r := &resourceHSMRootOfTrust{}
+	schemaReq := resource.SchemaRequest{}
+	schemaResp := &resource.SchemaResponse{}
+	r.Schema(context.Background(), schemaReq, schemaResp)
+
+	connInfo, ok := schemaResp.Schema.Attributes["conn_info"]
+	if !ok {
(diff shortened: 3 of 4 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-432

PR #395, opened 2026-07-23
TFIN-439: ciphertrust_proxy: http_proxy/https_proxy drift permanently undetectable once masked - terraform plan -refresh-only shows no changes

ashwani573:
## Summary

- Fixes `ciphertrust_proxy` drift detection for `http_proxy` and `https_proxy` by replacing the ad-hoc `maskProxyURL` string-splitting helper with `proxyNonPasswordPart`, which uses stdlib `net/url.Parse` and compares only the password-free portion (`scheme://username@host:port`).
- Applies the same fix symmetrically in both `Read()` and `Update()` post-response hydration so state never ends up holding a CM-masked URL that silently defeats all future drift checks.
- Expands `http_proxy` and `https_proxy` schema `Description` fields to document the remaining known limitation: password-only out-of-band changes are structurally undetectable because the CM API always returns the password masked as `xxxxxx`.
- Adds 9 unit tests for `proxyNonPasswordPart` (`internal/provider/cm/resource_proxy_test.go`) and 3 acceptance tests covering no-spurious-diff and host/port drift detection for both proxy attributes.

## Motivation

Implements TFIN-439: `ciphertrust_proxy`: `http_proxy`/`https_proxy` drift permanently undetectable once masked — `terraform plan -refresh-only` shows no changes.

## Root cause

CipherTrust Manager always returns proxy URLs with the password replaced by `xxxxxx` in `GET /v1/configs/proxy` responses. The previous `maskProxyURL` helper mirrored this masking on the Terraform state side and then compared the two masked strings. This worked correctly on the first refresh cycle, but broke in a specific sequence:

1. An out-of-band structural change (different host or port) was detected, so state was overwritten with the CM-masked URL (e.g. `http://user:xxxxxx@newhost:9090`).
2. On the next refresh, `maskProxyURL` applied to the already-masked state produced `http://user:xxxxxx@newhost:9090`, which matched the API response exactly.
3. Because the strings matched, Read() preserved the masked state — meaning subsequent `terraform plan -refresh-only` runs reported no drift even when the underlying resource continued to diverge.

Once state held a masked URL, drift became permanently invisible. The new `proxyNonPasswordPart` helper avoids the problem by stripping the password entirely before comparison. Both `http://user:cleartext@host:8080` and `http://user:xxxxxx@host:8080` produce `http://user@host:8080`, so the comparison is invariant to password representation. Changes to scheme, host, port, or username are always detected; password-only changes remain undetectable (an inherent API limitation, now documented in the schema).

### Modified file — `internal/provider/cm/resource_proxy.go`

- Removes `maskProxyURL` and replaces it with `proxyNonPasswordPart(rawURL string) string`.
  - Uses `url.Parse` from `net/url` (added import) instead of manual `strings.SplitN` chains.
  - Returns `scheme://username@host:port` — identical for the cleartext config value and the CM-masked API value when only the password differs.
  - Returns `""` for empty, unparseable, or schemeless URLs (the schema validator already rejects schemeless URLs at plan time).
- `Read()`: For each of `http_proxy` and `https_proxy`, the comparison is now:
  - If state attribute is null: leave it null (never configured in TF — do not populate from CM).
  - If CM returns a non-empty value: compare `proxyNonPasswordPart(apiValue)` vs `proxyNonPasswordPart(stateValue)`. If they differ (structural drift — different host, port, or username), overwrite state with the CM-masked API value. If they are equal, preserve the cleartext state value from the prior config.
  - If CM returns empty: set state to null (attribute cleared out-of-band).
- `Update()`: Applies the same `proxyNonPasswordPart` comparison to the post-PUT response before saving the plan value back to state. When the non-password portions match (including a password update from Terraform config), the cleartext plan value is preserved in state rather than overwritten with the masked API response.
- Schema `Description` for `http_proxy` and `https_proxy`: expanded with format guidance (scheme required) and an explicit known-limitation note that password-only OOB changes are undetectable while scheme/host/port/username changes are fully detectable.

The CM API endpoints used — `PUT /v1/configs/proxy` (Create/Update) and `GET /v1/configs/proxy` (Read) — are confirmed present in the swagger spec under the `cm-system` / Proxy area. No new URL constants were added; `URL_CM_PROXY = "api/v1/configs/proxy"` is unchanged.

### New file — `internal/provider/cm/resource_proxy_test.go`

- Unit tests for `proxyNonPasswordPart` (9 cases): empty string, unparseable URL, schemeless URL, cleartext credentialed URL, CM-masked credentialed URL, password-change produces same result, URL with no credentials, HTTPS scheme, and host/port change produces a different result (drift detectable).

- Adds `github.com/hashicorp/terraform-plugin-testing/terraform` import.
- `Test_CM_CipherTrustProxy_NoSpuriousDiffAfterApply`: applies a credentialed `http_proxy`, then immediately runs `RefreshState` with `ExpectNonEmptyPlan: false` — confirms no perpetual cleartext-vs-masked spurious diff.
- `Test_CM_CipherTrustProxy_HTTPProxyHostPortDriftDetection`: applies `http_proxy`, then uses a `PreConfig` out-of-band `PutData` to change the host/port, then asserts `ExpectNonEmptyPlan: true` — confirms host/port drift surfaces correctly.
- `Test_CM_CipherTrustProxy_HTTPSProxyHostPortDriftDetection`: same pattern for `https_proxy` with a different host/port.

All three new acceptance tests call `RequireCM(t)` as their first statement and use `t.Logf` (not `t.Fatalf`) inside `PreConfig` closures to avoid goroutine panics.

## Read() drift-detection behaviour (after this fix)

| Scenario | Before fix | After fix |
|---|---|---|
| No OOB change; `plan -refresh-only` after apply | No diff (correct) | No diff (correct) |
| OOB host/port change detected; subsequent refreshes | Permanently invisible after first detection | Detectable on every subsequent refresh |
| OOB password-only change | Undetectable (CM masks password) | Undetectable — documented known limitation |
| OOB clear of `http_proxy` | Null in state (correct) | Null in state (correct) |
| Attribute never configured; state null | Populated from CM masked URL (spurious diff source) | Left null (no spurious diff) |

## How to test

- [ ] `make test` — unit tests pass (includes `Test_CM_ProxyNonPasswordPart`).
  ```
  go test ./internal/provider/ -run 'Test_CM_CipherTrustProxy_' -v -timeout 15m
  ```
  Scenarios covered:
  - **No spurious diff** — apply config with credentialed proxy URL; `RefreshState` must report no changes.
  - **HTTP proxy host/port drift detection** — apply, mutate host/port out-of-band via CM API, `RefreshState` must report non-empty plan.
  - **HTTPS proxy host/port drift detection** — same as above for `https_proxy`.
- [ ] Manual smoke-test: apply proxy config with credentials, verify `terraform plan -refresh-only` shows no changes, mutate host OOB in CM, re-run `plan -refresh-only` and confirm diff is surfaced.

## Checklist

- [ ] `tflog.Trace` at entry/exit of every CRUD method: `[resource_proxy.go -> Func][uuid]` pattern.
- [ ] URL constant defined in `urls.go` — no inlined string literals (`URL_CM_PROXY` unchanged).
- [ ] CM API endpoint (`GET /v1/configs/proxy`, `PUT /v1/configs/proxy`) verified against swagger `cm-system` area.
- [ ] `proxyNonPasswordPart` returns `""` for schemeless or unparseable URLs — consistent with validator behaviour.

---

### DIFF
--- a/docs/resources/proxy.md
+++ b/docs/resources/proxy.md
@@ -62,8 +62,8 @@ output "proxie_id" {
 ### Optional
 
 - `certificate` (String) CA certificate to trust for proxy.
-- `http_proxy` (String, Sensitive) HTTP proxy URL for proxy configurations. If the proxy server's password contains any special character replace it with encoded values.
-- `https_proxy` (String, Sensitive) HTTPS proxy URL for proxy configurations. If the proxy server's password contains any special character replace it with encoded values.
+- `http_proxy` (String, Sensitive) HTTP proxy URL for proxy configurations. Include the scheme (e.g. `http://username:password@proxy.example.com:8080`). If the proxy server's password contains any special character replace it with percent-encoded values. **Known limitation**: CipherTrust Manager always returns this value with the password masked (replaced with `xxxxxx`) in GET responses. After `terraform apply`, Terraform state holds the cleartext value from your configuration. A password-only out-of-band change (same scheme, host, port, and username; different password only) is undetectable by `terraform plan -refresh-only` because the masked URL is structurally identical before and after. Changes to scheme, host, port, or username are fully detectable and will surface as drift.
+- `https_proxy` (String, Sensitive) HTTPS proxy URL for proxy configurations. Include the scheme (e.g. `https://username:password@proxy.example.com:8080`). If the proxy server's password contains any special character replace it with percent-encoded values. **Known limitation**: CipherTrust Manager always returns this value with the password masked (replaced with `xxxxxx`) in GET responses. After `terraform apply`, Terraform state holds the cleartext value from your configuration. A password-only out-of-band change (same scheme, host, port, and username; different password only) is undetectable by `terraform plan -refresh-only` because the masked URL is structurally identical before and after. Changes to scheme, host, port, or username are fully detectable and will surface as drift.
 - `no_proxy` (List of String) List of hosts for a proxy exception.
 
 ### Read-Only
--- a/internal/provider/cm/resource_proxy.go
+++ b/internal/provider/cm/resource_proxy.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/google/uuid"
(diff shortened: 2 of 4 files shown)


ashwani573:
## 🔍 Automated Code Review — TFIN-439

PR #501, opened 2026-08-08
Check CM version fo aws-key-material and improve mrk support

SarahThompson:
Block aws_key_material for CM version < 223 - ModifyPlan
Error if multi-region key and and CM version < 224  - Create function
On 2.23 support rotate-material the "old" way (import-material then rotate-material)
Improve support for replication and primary region change.

### DIFF
--- a/internal/provider/common/client.go
+++ b/internal/provider/common/client.go
@@ -275,7 +275,7 @@ func fetchCMVersion(ctx context.Context, c *Client) error {
 	response, err := c.GetById(ctx, "", "", URL_SYSTEMINFO)
 	if err != nil {
 		tflog.Error(ctx, "fetchCMVersion -> failed to fetch CM system info: "+err.Error())
-		return fmt.Errorf("failed to fetch CM system info: %w", err)
+		return fmt.Errorf("failed to fetch CM system info: %s", err.Error())
 	}
 	version := gjson.Get(response, "version").String()
 	if version == "" {
@@ -296,8 +296,8 @@ func fetchCMVersion(ctx context.Context, c *Client) error {
 	}
 	v, err := strconv.Atoi(parts[0] + parts[1])
 	if err != nil {
-		tflog.Error(ctx, fmt.Sprintf("fetchCMVersion -> failed to parse CM version %q: %w", version, err))
-		return fmt.Errorf("failed to parse CM version %q: %w", version, err)
+		tflog.Error(ctx, fmt.Sprintf("fetchCMVersion -> failed to parse CM version %q: %s", version, err.Error()))
+		return fmt.Errorf("failed to parse CM version %q: %s", version, err.Error())
 	}
 	c.CMVersion = v
 	c.CMFullVersion = version
--- a/internal/provider/common/client.go
+++ b/internal/provider/common/client.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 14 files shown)


internal/provider/provider.go
```diff
@@ -596,6 +596,11 @@ func (p *ciphertrustProvider) Configure(ctx context.Context, req provider.Config
 		client.Log = providerLogger
 		resp.DataSourceData = client
 		resp.ResourceData = client
+		if !client.IsCDSPaaS {
+			client.Log.Info(fmt.Sprintf("CipherTrust Manager k170v version %s", client.CMFullVersion))
```
ashwani573: we should not hardcode k170v as there can be other platforms also :"CipherTrust Manager k170v version"

internal/provider/common/client.go
```diff
@@ -251,11 +259,51 @@ func NewClient(ctx context.Context, uuid string, address, auth_domain, domain, u
...
+	}
 	tflog.Trace(ctx, MSG_METHOD_END+" [client.go -> NewClient]["+uuid+"]")
 	return &c, nil
 }
 
+// fetchCMVersion calls GET /api/v1/system/info and sets c.CMVersion and c.CMFullVersion.
+// Returns an error if the version cannot be fetched or parsed.
+// Sets CMVersion=9999 and CMFullVersion="Development" for Development builds.
+func fetchCMVersion(ctx context.Context, c *Client) error {
+	response, err := c.GetById(ctx, "", "", URL_SYSTEMINFO)
```
ashwani573: GetById with empty id is not the correct way, can we check if we can use: 
response, err := c.ReadDataByParam(ctx, "version-check", "all", URL_SYSTEMINFO)

internal/provider/common/client.go
```diff
@@ -251,11 +259,51 @@ func NewClient(ctx context.Context, uuid string, address, auth_domain, domain, u
...
+	version := gjson.Get(response, "version").String()
+	if version == "" {
+		tflog.Error(ctx, "fetchCMVersion -> CM system info response missing version field")
+		return fmt.Errorf("CM system info response missing version field")
+	}
+	if version == "Development" {
+		c.Log.Info("CM version: Development (returning 9999)")
+		c.CMVersion = 9999
+		c.CMFullVersion = "Development"
+		tflog.Error(ctx, "fetchCMVersion -> Development")
```
ashwani573: instead of tflog.Error, i think we should use tflog.Info

internal/provider/common/client.go
```diff
@@ -251,11 +259,51 @@ func NewClient(ctx context.Context, uuid string, address, auth_domain, domain, u
...
+		}
+	}
 	tflog.Trace(ctx, MSG_METHOD_END+" [client.go -> NewClient]["+uuid+"]")
 	return &c, nil
 }
 
+// fetchCMVersion calls GET /api/v1/system/info and sets c.CMVersion and c.CMFullVersion.
+// Returns an error if the version cannot be fetched or parsed.
+// Sets CMVersion=9999 and CMFullVersion="Development" for Development builds.
+func fetchCMVersion(ctx context.Context, c *Client) error {
```
ashwani573: i think we can use the  method receiver convention  for getting the CM version.

ashwani573 (changes requested):

internal/provider/common/client.go
SarahThompson: Made the function a receiver

internal/provider/common/client.go
SarahThompson: Updated to info log

internal/provider/common/client.go
SarahThompson: used c.ReadDataByParams

internal/provider/provider.go
SarahThompson: Removed k170v

PR #502, opened 2026-08-10
fix(cm): log_forwarder Update() crashes on every real update

zubeensyed:
Summary

ciphertrust_log_forwarder was effectively unusable for updates: a previous fix (5cd1bb5) resolved the originally-reported symptoms (unconditional name resend, perpetual updated_at diff, plan-shaped state overwrite) but the fix itself had two schema-level bugs that made Terraform abort with Provider produced inconsistent result after apply on almost every update. Both are fixed here, live-verified against a CM instance (10.171.98.28).

Root causes:

1. updated_at had UseStateForUnknown(), which pins the planned value to prior state. CM bumps updated_at on every successful PATCH, so the real post-apply value always diverged from the pinned plan value → crash on every update. Fixed by removing the modifier; a server-driven timestamp can't be predicted at plan time and must resolve at apply.

2. The 12 nested leaf attributes (elasticsearch_params.indices.*, loki_params.labels.*, syslog_params.forward_logs.*) were plain Optional, not Computed. CM's PATCH is a genuine partial merge — an omitted field keeps its previous server-side value (verified via raw GET) — but Optional-only means Terraform demands the final state match the plan exactly. The moment a config omitted a previously-set field, the correctly-preserved value from CM collided with the plan's implied null and crashed. Fixed by making these Optional + Computed with UseStateForUnknown(). This also incidentally fixed a perpetual ""-null / false-null diff that showed up on the very first plan after create (CM defaults never-set fields to ""/false rather than omitting them).

3. Companion fix: Create() wrote the plan object straight to state instead of hydrating from the live response. Once the leaves became Computed, an omitted config field stayed Unknown through Create(), which Terraform rejects outright. Create() now reuses the existing hydrateLogForwarderState() helper Update() already relied on.

### DIFF
--- a/internal/provider/cm/resource_log_forwarder.go
+++ b/internal/provider/cm/resource_log_forwarder.go
@@ -81,7 +81,7 @@ func (r *resourceCMLogForwarders) Schema(_ context.Context, _ resource.SchemaReq
 							"activity_kmip": schema.StringAttribute{
 								Optional:    true,
 								Computed:    true,
-								Description: "Index to be used for entries coming from the KMIP activity log. Logs will not be forwarded if index is not provided. Consult Elasticsearch documentation for allowed characters.",
+								Description: "Index to be used for entries coming from the KMIP activity log. Logs will not be forwarded if index is not provided. Consult Elasticsearch documentation for allowed characters. Once set, this cannot currently be cleared by removing it from configuration or setting it to null - the previous value is preserved.",
 								PlanModifiers: []planmodifier.String{
 									stringplanmodifier.UseStateForUnknown(),
 								},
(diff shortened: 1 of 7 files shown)


internal/provider/cm/resource_log_forwarder.go
```diff
@@ -79,19 +80,35 @@ func (r *resourceCMLogForwarders) Schema(_ context.Context, _ resource.SchemaReq
 						Attributes: map[string]schema.Attribute{
 							"activity_kmip": schema.StringAttribute{
 								Optional:    true,
+								Computed:    true,
 								Description: "Index to be used for entries coming from the KMIP activity log. Logs will not be forwarded if index is not provided. Consult Elasticsearch documentation for allowed characters.",
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.UseStateForUnknown(),
```
ashwani573, line 86: The real compat break — silently ignoring removals from config:                                                                                                                                           
  
  With Optional+Computed+UseStateForUnknown(), if a user removes activity_kmip from their config to stop forwarding, Terraform will silently keep the prior state value in the plan instead of planning a   
  null/empty update. Users who manage this field declaratively will be surprised:
                                                                                                                                                                                                            
  # User removes this line expecting it to clear the index:
  # activity_kmip = "my-index"                                                                                                                                                                              
  
  # → Terraform plans NO change (UseStateForUnknown preserves "my-index" from state)                                                                                                                        
  # → CM is never told to clear it
                                                                                                                                                                                                            
  To explicitly clear a field, users now need activity_kmip = null. This should be documented in the resource docs and the CHANGELOG. It's a silent behavior change, not a crash.

PR #503, opened 2026-08-10
TFIN-599: Fix ciphertrust_cm_key revocation_reason/revocation_message never reaching CM

zubeensyed:
Setting revocation_reason and revocation_message on a ciphertrust_cm_key resource made terraform apply fail with a 400 error instead of revoking the key. CipherTrust Manager has a separate, dedicated endpoint just for revoking keys, but the provider was sending those two fields on the same request it uses for all other key updates, which does not accept them (and expects different field names).

The provider now calls CM's dedicated revoke endpoint with the correct field names instead of bundling revocation into the general update request. It only calls it when the revocation fields actually change, so it won't try to re-revoke an already-revoked key on every apply.

Added a test that creates a key, revokes it through Terraform, and then independently checks CipherTrust Manager's own API to confirm the key shows up as "Compromised" there — not just in Terraform's state. Ran it against a live CM instance, along with a few existing key tests, to confirm nothing else broke.

### DIFF
--- a/internal/provider/cm/resource_cm_key.go
+++ b/internal/provider/cm/resource_cm_key.go
@@ -859,6 +859,20 @@ func (r *resourceCMKey) Schema(_ context.Context, _ resource.SchemaRequest, resp
 	}
 }
 
+// revokeKey calls CM's dedicated key revoke endpoint. This is a separate operation from the
+// general key PATCH/POST endpoint, which does not accept revocation fields under any name.
+// Note the request body uses "reason"/"message", not the revocationReason/revocationMessage
+// names CM reports back on GET.
+func (r *resourceCMKey) revokeKey(ctx context.Context, id, reason, message string) error {
+	payload := CMKeyRevokeJSON{Reason: reason, Message: message}
+	payloadJSON, err := json.Marshal(payload)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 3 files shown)


internal/provider/cm/resource_cm_key.go
```diff
@@ -859,6 +859,20 @@ func (r *resourceCMKey) Schema(_ context.Context, _ resource.SchemaRequest, resp
 	}
 }
 
+// revokeKey calls CM's dedicated key revoke endpoint. This is a separate operation from the
+// general key PATCH/POST endpoint, which does not accept revocation fields under any name.
+// Note the request body uses "reason"/"message", not the revocationReason/revocationMessage
+// names CM reports back on GET.
+func (r *resourceCMKey) revokeKey(ctx context.Context, id, reason, message string) error {
```
ashwani573, line 866: do we need to document it as a new resource, create a story

zubeensyed: No docs update is required. 
Current docs (docs/resources/cm_key.md) already match the schema descriptions exactly. 
I only updated the internal payload/endpoint handling in Create()/Update() and did not modify any attribute descriptions. Also, the example under examples/resources/ciphertrust_cm_key/ does not involve revocation, so no changes are needed there either.

PR #513, opened 2026-08-11
TFIN-594: Fix ciphertrust_oci_connection crash when meta or description is set to an explicitly empty value

zubeensyed:
Setting `meta = {}` or `description = ""` on a new (or existing) OCI connection crashed the apply with "Provider produced inconsistent result after apply", even though CipherTrust Manager had already created the connection - leaving it orphaned in Terraform's eyes.

The cause: CM's API omits (or returns null for) a field that has no value, so after every create/update/read the provider was collapsing meta/description straight to null. That contradicted an explicitly-empty plan, which Terraform requires back unchanged.

The fix only nulls a field when there was nothing meaningful to keep: either it was already null, or CM shows it changed from a real value to nothing (i.e. cleared on CM's side). 
An explicitly-empty value the user configured is now preserved instead of being wiped out.


ashwani573:
approved

PR #514, opened 2026-08-11
Fix ciphertrust_interface tls_ciphers and auto_gen_ca_id non-convergence

zubeensyed:
Two independent perpetual-diff bugs on ciphertrust_interface, both in the response-to-state hydration direction (the payload builders were already correct, which is why the symptoms appeared on post-apply refresh rather than on apply itself).

1. tls_ciphers never converged (element ordering)

tls_ciphers was a schema.ListNestedAttribute, so element order was significant. CM re-orders the cipher-suite entries unpredictably on every write: repeated GETs are stable, but after a PATCH the order CM returns matches neither the previously-returned order nor the order just submitted -- only the membership is preserved (live-confirmed against 10.171.98.28). Read() therefore hydrated state in CM's post-write order, which never equalled the config's order, so every plan showed a positional diff:

    ~ cipher_suite = "TLS_RSA_WITH_AES_128_CBC_SHA256" -> "TLS_AES_128_GCM_SHA256"
    ~ enabled      = false -> true
    Plan: 0 to add, 1 to change, 0 to destroy.

Re-applying just re-shuffled it, so the resource never stabilised.

Changed the attribute to schema.SetNestedAttribute, which takes ordering out of both state and diffing. No struct changes were needed -- SetNestedAttribute binds to the same []TLSCiphersTFSDK slice, and this is the established idiom in the repo (acls, defined_tags, replica_keys).

2. auto_gen_ca_id = "" never converged

Setting auto_gen_ca_id to an empty string is the documented way to disable server-certificate auto-generation, and CM treats it as a real value: PATCH {"auto_gen_ca_id":""} returns 200 with "auto_gen_ca_id":"" and the subsequent GET agrees -- the key is echoed, not omitted (live-confirmed). Both hydration sites guarded on
`r.Exists() && r.String() != ""` and so collapsed that "" to null:

  - Read()   -- ran on every refresh, so this is the site that produced
               the perpetual `+ auto_gen_ca_id = ""` plan
  - Create() -- same bug on the create-with-"" path

Dropped the `&& r.String() != ""` half of the guard at both sites, keeping r.Exists(). Scoped deliberately to this field: the same idiom on other optional strings (name, cert_user_field, ...) has no documented empty-string sentinel and is left untouched.

Tests

Unit (internal/provider/cm/resource_interface_convergence_test.go, new):
  - Test_CMInterfaceSchema_TLSCiphersIsSetNotList
  - Test_CMInterfaceRead_TLSCiphersReorderedByCMConverges
  - Test_CMInterfaceRead_AutoGenCAIdEmptyStringPreserved
  - Test_CMInterfaceRead_AutoGenCAIdDriftFromCAToEmptyDetected
  - Test_CMInterfaceCreate_AutoGenCAIdEmptyStringPreserved

Acceptance (internal/provider/resource_interface_test.go):
  - Test_CM_Interface_TLSCiphersOrderInsensitiveConverges
  - Test_CM_Interface_AutoGenCAIdEmptyStringConverges

Each test was verified to FAIL with its fix reverted, not merely to pass with it applied. The drift test needed tightening for this: types.String ValueString() returns "" for a null value too, so asserting only the string passed vacuously against the pre-fix behaviour; it now asserts the value is known rather than null.

The acceptance tests build their tls_ciphers config from CM's live cipher set rather than hardcoding suite names, because CM rejects adding or removing members ("Adding or removing TLS cipher suites is not allowed", HTTP 400) -- a config must list every suite the interface has and only the enabled flags are mutable. Hardcoding would break on any CM release that changes the set. This constraint is now documented on the attribute.

Validated against CM 10.171.98.28: full unit suite green, both new acceptance tests pass, and the tls_ciphers acceptance test reproduces the exact reported diff when the schema type is reverted to List. The full interface acceptance suite passes with these changes (733s).

Unrelated observation while validating: the pre-existing Test_CM_CMInterface_ClearRegistrationToken is intermittently flaky. It failed once, then passed in isolation, passed on an unmodified 1.0.1 baseline, and passed again on a second full-suite run of this same code -- identical code, both outcomes, so it is not caused by this change. It picks a random port (9400 + rand(400)) and provisions a live registration token each run, either of which could explain the intermittency. Not touched here; worth its own look.

Notes / follow-ups (deliberately not in this change)

- State-format change: List -> Set alters the state representation of tls_ciphers, and this provider has no SchemaVersion/UpgradeState machinery. Anyone already managing tls_ciphers will likely need `terraform state rm` + re-import. Worth a release note.
- Create() does not put tls_ciphers in its POST payload but does hydrate the field from the POST response, so a create-time tls_ciphers config fails with "Provider produced inconsistent result after apply". This is a separate pre-existing gap, unrelated to ordering; the new acceptance test manages the attribute via PATCH, which is the working path. Needs its own ticket.
- tls_groups and allowed_identity_types (finding 3 of the ticket) remain unimplemented and are out of scope here. tls_groups is confirmed present in live GET responses as [{group_name, enabled}]; allowed_identity_types was not present on the NAE interface at all, so its shape needs live confirmation before being added.

### DIFF
--- a/docs/resources/interface.md
+++ b/docs/resources/interface.md
@@ -93,7 +93,7 @@ output "interface_id" {
 - `network_interface` (String) Defines what ethernet adapter the interface should listen to, use "all" for all. Defaults to all if not specified.
 - `registration_token` (String, Sensitive) Registration token in case auto registration is true. Write-only: never stored in Terraform state or plan artifacts (requires Terraform 1.11+). To resend a rotated token, change `registration_token` and bump `registration_token_version` in the same apply.
 - `registration_token_version` (Number) Arbitrary version number stored in state and used to trigger re-sending `registration_token` to CipherTrust Manager. Since `registration_token` is write-only, Terraform cannot detect a change in its value on its own; increment this on every apply where you want the current `registration_token` value re-sent.
-- `tls_ciphers` (Attributes List) The list of TLS cipher suites available for the interface's (KMIP, NAE, or Web) TLS handshake, and whether each is enabled. (see [below for nested schema](#nestedatt--tls_ciphers))
+- `tls_ciphers` (Attributes Set) The set of TLS cipher suites available for the interface's (KMIP, NAE, or Web) TLS handshake, and whether each is enabled. Ordering is not significant. CipherTrust Manager does not permit adding or removing cipher suites, so this must list every suite the interface already has; only the enabled flags can be changed. (see [below for nested schema](#nestedatt--tls_ciphers))
 - `trusted_cas` (Attributes) Collection of local and external CA IDs to trust for client authentication on this interface. (see [below for nested schema](#nestedatt--trusted_cas))
 
 ### Read-Only
--- a/internal/provider/cm/resource_interface.go
+++ b/internal/provider/cm/resource_interface.go
@@ -316,9 +316,21 @@ func (r *resourceCMInterface) Schema(_ context.Context, _ resource.SchemaRequest
 					},
 				},
 			},
-			"tls_ciphers": schema.ListNestedAttribute{
+			// Set, not List: CM re-orders the cipher-suite entries unpredictably on every
+			// write. Live-confirmed: repeated GETs are stable, but after a PATCH the returned
+			// order matches neither the previously-returned order nor the order just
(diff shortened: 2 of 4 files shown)


ashwani573:
approved

PR #519, opened 2026-08-11
TFIN-606: fix(cm_key): read meta.permissions regardless of PascalCase/snake_case

zubeensyed:
CM accepts and persists meta.permissions field names (ReadKey/read_key, ExportKey/export_key, etc.) in either casing, and echoes back whichever casing was last written - including via non-Terraform clients. Read() only checked the PascalCase gjson path, so any field CM stored in snake_case silently hydrated to null on every refresh, with no Terraform-native way to recover the value.

Add a per-field fallback so each meta.permissions field is read from its PascalCase path first, then its snake_case path if that's absent.

Verified against a live CM instance: PATCHing meta.permissions in snake_case and echoing it back on GET previously corrupted state to null on refresh; with this fix, terraform apply -refresh-only reports no drift and preserves the correct values.

### DIFF
--- a/internal/provider/cm/resource_cm_key.go
+++ b/internal/provider/cm/resource_cm_key.go
@@ -1406,6 +1406,22 @@ func (r *resourceCMKey) Create(ctx context.Context, req resource.CreateRequest, 
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ashwani573:
approved.

PR #525, opened 2026-08-12
Cluster test fixes

zubeensyed:
• fix(cm_cluster): stop CM's post-join publicAddress lag from failing the cluster test

## Problem
`Test_CM_ResourceCMCluster` fails on 2.21, passes on 2.22+/2.26:
`→ public_address = "" → ... → refresh plan was not empty, then the destroy aborts and everything after it fails with "cluster already exists".`

## Root cause
The provider races CM; only 2.21 loses. Measured via `GET /nodes/{id}` on the member — the call `Read()` makes — 2.21.3 reports a joined node as `status=""` for ~99-3 seconds before** publishing its `publicAddress`; 2.26 publishes it up front. `Create()` returned inside that window, the refresh read `""` and stored it over the configured value.

Two more bugs surfaced past that point: `cluster_node.Delete()` hard-erred when the node being removed answered with HTTP 500 (normal while it catches up), aborting the destroy with the node left in state *and* in the cluster; and `cluster.Delete()` verified for only 3 × 5 s, far short of the service restart `DELETE /cluster` triggers, failing a teardown that had succeeded.

## Changes (resource code only, 2 files)
- `memberPublicAddress()` — re-read `GET /nodes/{id}` while the member reports none (24 × 5 s).
- `cluster_node.Create()` — wait for publication before returning; warn-only on timeout.
- `cluster_node.Read()` / `cluster.Read()` — re-check an empty value instead of storing it; a value that outlasts the budget is still recorded, so genuine drift surfaces.
- `cluster_node.Delete()` — retry the removed node's client, then proceed without it.
- `cluster.Delete()` — verify on both paths; budgets sized to outlast the CM restart.

No schema, plan-modifier or version-detection changes. Costs one extra GET per join on versions that publish promptly.

## Verification (live CM, full suite)
2.21.3: FAIL before → **PASS** after (two runs). 2.26.0-beta1: PASS before and after. Nodes left unclustered every time. Acceptance test untouched; existing unit tests pass unchanged.

### DIFF
--- a/internal/provider/cm/resource_cm_cluster.go
+++ b/internal/provider/cm/resource_cm_cluster.go
@@ -28,11 +28,17 @@ var (
 	_ resource.ResourceWithConfigure      = &resourceCMCluster{}
 	_ resource.ResourceWithValidateConfig = &resourceCMCluster{}
 
-	// deleteVerifyRetries/deleteVerifyInterval bound how many times Delete() re-checks
-	// actual cluster status after a DELETE /cluster error before giving up (see Delete
-	// below). Vars (not consts) so tests can shorten deleteVerifyInterval.
-	deleteVerifyRetries  = 3
-	deleteVerifyInterval = 5 * time.Second
+	// deleteVerifyRetries/deleteVerifyInterval bound Delete()'s post-teardown polling.
+	// The budget has to outlast a CM service restart, not just the DELETE call: on 2.21.3
+	// the DELETE never returned and logins were still failing 20s later, so the previous
+	// 3 × 5s expired mid-restart and failed a destroy that had succeeded. Vars so tests
+	// can shorten them.
+	deleteVerifyRetries  = 90 // 15 minutes
+	deleteVerifyInterval = 10 * time.Second
+
+	// deleteVerifiedHealthRetries is the shorter budget for when the DELETE succeeded: CM
+	// already confirmed the deletion, so this only waits for the node to serve again.
+	deleteVerifiedHealthRetries = 30 // 5 minutes
 )
 
 func NewResourceCMCluster() resource.Resource {
@@ -240,11 +246,22 @@ func (r *resourceCMCluster) Read(ctx context.Context, req resource.ReadRequest, 
 	// A transient fetch failure here shouldn't fail the whole Read, so leave the prior
 	// state value in place rather than erroring.
 	if nodeInfo, nerr := r.client.GetById(ctx, id, nodeID, common.URL_NODES); nerr == nil {
-		// CM returns "" (not an omitted field) when no public_address was ever
-		// configured. public_address is Optional (not Computed), so an unconfigured
-		// attribute plans as null, not "" — mapping "" to StringValue("") here would
-		// permanently disagree with that null and show a spurious diff on every plan.
-		if publicAddress := gjson.Get(nodeInfo, "publicAddress").String(); publicAddress != "" {
+		// CM reports no publicAddress when none was ever configured. public_address is
+		// Optional (not Computed), so an unconfigured attribute plans as null, not "" —
+		// mapping that to StringValue("") here would permanently disagree with the null
+		// and show a spurious diff on every plan.
+		publicAddress := gjson.Get(nodeInfo, "publicAddress").String()
+		// But when state holds an address, an empty reply is ambiguous — CM also omits the
+		// field for a while after a create/join (see memberPublicAddress) — so re-check.
+		if publicAddress == "" && !state.PublicAddress.IsNull() {
+			if waited, waitErr := memberPublicAddress(ctx, r.client, id, nodeID); waitErr != nil {
+				r.client.Log.Debug("[resource_cm_cluster.go -> Read][" + id + "] could not re-check publicAddress, keeping prior state value: " + waitErr.Error())
+				publicAddress = state.PublicAddress.ValueString()
+			} else {
+				publicAddress = waited
+			}
+		}
+		if publicAddress != "" {
 			state.PublicAddress = types.StringValue(publicAddress)
 		} else {
 			state.PublicAddress = types.StringNull()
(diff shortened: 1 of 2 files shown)


internal/provider/cm/resource_cm_cluster.go
```diff
@@ -240,11 +246,22 @@ func (r *resourceCMCluster) Read(ctx context.Context, req resource.ReadRequest,
...
-		if publicAddress := gjson.Get(nodeInfo, "publicAddress").String(); publicAddress != "" {
+		// CM reports no publicAddress when none was ever configured. public_address is
+		// Optional (not Computed), so an unconfigured attribute plans as null, not "" —
+		// mapping that to StringValue("") here would permanently disagree with the null
+		// and show a spurious diff on every plan.
+		publicAddress := gjson.Get(nodeInfo, "publicAddress").String()
+		// But when state holds an address, an empty reply is ambiguous — CM also omits the
+		// field for a while after a create/join (see memberPublicAddress) — so re-check.
+		if publicAddress == "" && !state.PublicAddress.IsNull() {
+			if waited, waitErr := memberPublicAddress(ctx, r.client, id, nodeID); waitErr != nil {
```
ashwani573, line 257: Please check this issue               
   — cluster_node.go Read() unconditionally calls memberPublicAddress() for every node without a public address
                                                                                                                                                                                                            
  File: resource_cm_cluster_node.go, new Read block (~line 715)                                                                                                                                             
                                                                                                                                                                                                            
  publicAddress := gjson.Get(nodeInfo, "publicAddress").String()                                                                                                                                            
  if publicAddress == "" {                                                                                                                                                                                  
      waited, waitErr := memberPublicAddress(ctx, r.client, id, nodeID)                                                                                                                                     
      ...                                                                                                                                                                                                   
  }                                                                                                                                                                                                         
                  
  There is no guard here. cluster.go Read() correctly gates this on !state.PublicAddress.IsNull():                                                                                                          
  if publicAddress == "" && !state.PublicAddress.IsNull() {
                                                                                                                                                                                                            
  But cluster_node.go's original code set state.PublicAddress = types.StringValue("") (never null) so IsNull() would always be false anyway — the correct guard here is state.PublicAddress.ValueString() !=
   "":                                                                                                                                                                                                      
                                                                                                                                                                                                            
  if publicAddress == "" && state.PublicAddress.ValueString() != "" {                                                                                                                                       
                  
  Impact: Any cluster node with no public_address configured will block in Read() for up to 2 minutes on every plan/refresh. This is a significant regression for normal operations.

internal/provider/cm/resource_cm_cluster.go
```diff
@@ -321,34 +338,76 @@ func (r *resourceCMCluster) Delete(ctx context.Context, req resource.DeleteReque
...
+	}
+
+	// The teardown restarts CM whether or not the DELETE returned, so wait for the node to
+	// answer before returning — otherwise the next plan fails building its API client
+	// (HTTP 500 from auth). After an error the node must also report itself unclustered,
+	// the only evidence the teardown happened; after a 200 any answer will do, since
+	// requiring "unclustered" would stall the full budget when a 200 leaves a cluster
+	// behind (the endpoint only removes the last node).
+	budget := deleteVerifyRetries
+	if err == nil {
```
ashwani573, line 355: Bug — Budget exhaustion with no API error silently clears publicAddress in cluster_node.go Read()
                                                                                                                                                                                                            
  When memberPublicAddress() runs out its budget with no API error (the CM node is reachable but hasn't published the address yet), it returns ("", nil). In the caller:
                                                                                                                                                                                                            
  if waitErr != nil {
      waited = state.PublicAddress.ValueString() // preserve                                                                                                                                                
  }                                                                                                                                                                                                         
  publicAddress = waited
                                                                                                                                                                                                            
  When waitErr == nil and waited == "", publicAddress is set to "" and state gets types.StringValue(""). If state held "192.168.1.100" before, Terraform now sees a drift and will plan to clear            
  public_address.
                                                                                                                                                                                                            
  The same pattern has the same flaw in cluster.go Read(), where budget exhaustion with no error transitions state to types.StringNull().                                                                   
  
  Suggested fix — treat ("", nil) the same as an error when state had a value:                                                                                                                              
                  
  if waited == "" {                                                                                                                                                                                         
      // Budget exhausted but API is reachable; ambiguous — keep prior value.                                                                                                                               
      publicAddress = state.PublicAddress.ValueString()                                                                                                                                                     
  } else {                                                                                                                                                                                                  
      publicAddress = waited                                                                                                                                                                                
  }

zubeensyed: This is intentional based on how public_address is defined as Required, so config drives the plan. An empty state value produces a diff that restores the configured address, rather than one that clears it. This is also observable in the pre-fix failures: public_address = "" -> "10.171.96.224".

zubeensyed: Recording the empty value after the budget is deliberate, since it is the only way a genuine out-of-band clear (PATCH publicAddress:"") can be observed. Falling back to the prior state for "" would make that case permanently undetectable and leave the state silently disagreeing with CM.

internal/provider/cm/resource_cm_cluster.go
zubeensyed: This can't happen: a `cluster_node` with no `public_address` cannot exist. The attribute is Required, and `Create()` runs it through `extractHost()`, which rejects `""` with "host must not be empty", the apply fails before any state is written. There's no `ImportState` either, so state only comes from `Create()`. 

On a normal plan, the address is already in the member-check response above, so the condition short-circuits on the first term with no extra call and no wait.

The guard would only matter after `Read()` records a genuinely cleared address (CM empty past the budget), where state is `""` and later plans re-enter the wait. That state already has a pending diff, so it's an inefficiency and IMO not worth widening the diff on a branch that's verified green on 2.21.3 and 2.26.

PR #530, opened 2026-08-13
[TFIN-625] Fix cte_client_group op_type=ldt-pause missing field-change guard

manishkumar-thales:
The ldt-pause branch in Update() was the only one of the 7 op_type branches with no "no other field changed" validation. Every sibling branch (update, auth-binaries, update-password, reset-password, remove-client, add-client) begins with an exhaustive set of plan-vs-state checks rejecting any field change that op_type doesn't apply to; ldt-pause went straight to posting the paused flag to CM with zero such checks.

Impact: setting op_type="ldt-pause" while simultaneously changing an unrelated field (e.g. description) in the same apply succeeded with no error. CM only received the pause-state change, but Terraform's state recorded the unrelated field's new value as if it had been applied -- a false value for a live resource, only surfaced on the next refresh as a revert diff.

Fix: add the same guard block used by the sibling branches to ldt-pause, checking every field other than the mutable "paused" field itself. Verified on live CM that the same repro (ldt-pause + simultaneous description change) now fails cleanly at apply time instead of silently corrupting state, while plain ldt-pause-only usage (toggling paused, no other field changed) continues to work with no plan loop.

### DIFF
--- a/internal/provider/cte/resource_cte_clientgroup.go
+++ b/internal/provider/cte/resource_cte_clientgroup.go
@@ -1008,6 +1008,68 @@ func (r *resourceCTEClientGroup) Update(ctx context.Context, req resource.Update
 			return
 		}
 	} else if opType == "ldt-pause" {
+		// Add error checks for fields we cant change in op_type = ldt-pause
+		if !stringSlicesEqual(plan.ClientList, state.ClientList) {
+			resp.Diagnostics.AddError("Invalid data input: CTE Client Group LDT pause", "client_list cannot be changed with op_type 'ldt-pause'")
+			return
+		}
+		if plan.InheritAttributes != state.InheritAttributes {
+			resp.Diagnostics.AddError("Invalid data input: CTE Client Group LDT pause", "inherit_attributes cannot be changed with op_type 'ldt-pause'")
+			return
+		}
+		if plan.AuthBinaries != state.AuthBinaries {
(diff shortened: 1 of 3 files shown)


internal/provider/cte/resource_cte_clientgroup.go
```diff
@@ -1008,6 +1008,68 @@ func (r *resourceCTEClientGroup) Update(ctx context.Context, req resource.Update
 			return
 		}
 	} else if opType == "ldt-pause" {
+		// Add error checks for fields we cant change in op_type = ldt-pause
+		if !stringSlicesEqual(plan.ClientList, state.ClientList) {
```
ashwani573, line 1012: One pre-existing bug to note (not introduced here):                                                                                                                                                       
  In the ldt-pause payload block (already existed before this PR):
  if plan.Paused.ValueBool() != types.BoolNull().ValueBool() {                                                                                                                                              
  types.BoolNull().ValueBool() always returns false (zero value), so this condition is plan.Paused.ValueBool() == true. Setting paused=false (un-pause) never gets sent to CM. Worth a follow-up ticket, but
   out of scope for this PR.

manishkumar-thales: Fixed this issue along with other similar occurrence.

ashwani573:
approved

PR #531, opened 2026-08-13
[TFIN-634] Fix cte_client_guardpoint: adding a guard point no longer forces a whole-resource replace with no rollback.

manishkumar-thales:
guard_point_type carried stringplanmodifier.RequiresReplace() at the schema level, inside a MapNestedAttribute. That plan modifier cannot distinguish "an existing guard point's type actually changed" from "a brand-new guard_path was added" -- both look like PlanValue != StateValue, since StateValue is null for a map key that never existed in prior state. So adding ANY new guard point forced a whole-resource -/+ replace, destroying every existing, untouched guard point too. Terraform's destroy-before-create ordering means the destroy had already completed by the time Create() ran; if Create() then failed for any reason (invalid guard_point_type on the new entry, or a 409 from CM), there was no rollback -- confirmed live: recovery required a full terraform apply from scratch, and a partial multi-batch Create() failure could even leave a real, untracked, guarded path on CM with zero Terraform record.

Fix: added modifiers.RequiresReplaceUnlessNewMapEntry(), a String plan modifier that only requires replace when the map element already existed in prior state (a genuine change to an existing, converged guard point). A brand-new map key is created in place by Update()'s existing "Phase 1: create new paths" logic instead, so it can never trigger a whole-resource replace. Changing an EXISTING guard point's type still forces a replace exactly as before (verified live) -- Update()'s own immutable-field check for guard_point_type on existing entries remains in place as a defense-in-depth backstop.

Added TestCTEClientGuardPointResource_guardPointTypeReplaceScope to resource_cte_client_guardpoints_test.go, asserting via plancheck that adding a new guard point plans as ResourceActionUpdate (never DestroyBeforeCreate) and that changing an existing entry's type still plans as ResourceActionDestroyBeforeCreate.

### DIFF
--- a/internal/provider/cte/resource_cte_client_guardpoints.go
+++ b/internal/provider/cte/resource_cte_client_guardpoints.go
@@ -18,6 +18,7 @@ import (
 	"github.com/tidwall/gjson"
 
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
+	"github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/modifiers"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


ashwani573:
approved

PR #533, opened 2026-08-13
[TFIN-630] Fix cte_clientgroup_designatedprimaryset immutable field plan visibility.

manishkumar-thales:
client_group_id, name, and ldt_comm_group_service_id were declared as plain Required string attributes with no PlanModifiers, even though Update() already hard-errors on any change to them. This meant terraform plan showed a normal in-place update for a change to any of these fields, and the error only surfaced at apply time ("name is an immutable field" / "LDT comm group service id is an immutable field").

Add stringplanmodifier.RequiresReplace() to all three attributes so Terraform surfaces the required destroy+recreate at plan time instead.

Verified on live CM: before the fix, changing name or ldt_comm_group_service_id on an existing designatedprimaryset planned as an in-place update and failed at apply with the immutable-field error. After the fix, the same changes (plus client_group_id) plan as destroy-then-create replacements, and a real name/ldt_comm_group_service_id change applies cleanly as a replace with no drift afterward.

### DIFF
--- a/internal/provider/cte/resource_cte_clientgroup_dps.go
+++ b/internal/provider/cte/resource_cte_clientgroup_dps.go
@@ -16,6 +16,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 
 	common "github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/common"
+	"github.com/ThalesGroup/terraform-provider-ciphertrust/internal/provider/modifiers"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


zubeensyed:
I’d suggest using the custom modifier ImmutableString() here instead of stringplanmodifier.RequiresReplace(). 

ImmutableString() preserves the current Update() behavior, where changes to these fields are hard-blocked without destroy/recreate, while moving the validation to plan time. 

RequiresReplace() would introduce a new behavior where Terraform silently destroys and recreates the DPS on a plan change, which could leave the resource unavailable if the recreate fails after the destroy.

manishkumar-thales:
Made changes as suggested, used custom ImmutableString() instead of RequiresReplace().
Also changed the description of required fields.

zubeensyed (approved):

PR #534, opened 2026-08-13
[TFIN-628][TFIN-544][TFIN-632][TFIN-633] Fix cte guardpoint batching and immutability gaps.

manishkumar-thales:
[TFIN-628]: resource_cte_clientgroup_guardpoints.go's batchKey struct (used to group new guard points into batched POST /guardpoints calls in Create() and Update()'s "create new paths" phase) omitted guard_enabled, data_classification_enabled, data_lineage_enabled, and intelligent_protection from its comparison fields. Two guard points added in the same apply that differ ONLY in one of these fields hashed to the same batchKey and collapsed into a single batched API call, so only one guard point's requested value for that field actually reached CM ("first writer wins"). Confirmed live: creating two guard points differing only in intelligent_protection produced a single POST to the guardpoints endpoint instead of two. Fix: added the 4 missing fields to the batchKey struct (and its construction) in both Create() and Update().

[TFIN-544]: same batchKey gap, same root cause, reproduces identically on the sibling resource_cte_client_guardpoints.go (confirmed live: 1 POST for 2 guard points differing only in data_lineage_enabled instead of 2). Applied the identical batchKey fix there. This was previously partially fixed (id instability via UseNonNullStateForUnknown, PR #457) but reopened because the guard_enabled/field-collision half of the symptom remained -- that half is what this batchKey fix addresses.

[TFIN-632]: policy_id on resource_cte_clientgroup_guardpoints.go was Required with no PlanModifiers, so changing it on an EXISTING guard_path was planned as a plain in-place update and only rejected at apply time by Update()'s manual AddError check. Fix: apply modifiers.RequiresReplaceUnlessNewMapEntry() (not plain RequiresReplace(), which would reintroduce TFIN-634's destructive-replace-on-add bug) so a genuine change to an existing entry's policy_id is now visible as a plan-time "-/+ replace", while adding a brand-new guard_path with any policy_id still does not force a replace.

[TFIN-633]: automount_enabled, cifs_enabled, disk_name, diskgroup_name, early_access, intelligent_protection, is_idt_capable_device, preserve_sparse_regions, data_classification_enabled, and data_lineage_enabled on resource_cte_clientgroup_guardpoints.go had no PlanModifiers and no runtime check at all. CM silently no-ops an update to these fields via PATCH (the PATCH payload built by UpdateCTEGuardPointJSON never includes them), so apply reported success while state permanently diverged from CM with no way to self-correct. Confirmed live for automount_enabled. Fix: added a new Bool plan-modifier variant, modifiers.BoolRequiresReplaceUnlessNewMapEntry() (Bool sibling of the existing String RequiresReplaceUnlessNewMapEntry() from TFIN-634), and applied it (or the String variant, for disk_name/diskgroup_name) to all 10 fields, so a genuine change to an existing entry now correctly replaces (which actually applies the new value via a fresh Create()) instead of silently no-opping. data_classification_enabled, data_lineage_enabled, and intelligent_protection needed both this plan-modifier fix and the TFIN-628 batchKey fix -- they are complementary, not redundant.

Scope: resource_cte_clientgroup_guardpoints.go received all three fixes (batchKey + policy_id modifier + 10-field modifiers). resource_cte_client_guardpoints.go received only the batchKey fix, per scope.

Added two new top-level acceptance tests in
resource_cte_clientgroup_guardpoints_test.go:
TestCTEClientGroupGuardPointResource_policyIDRequiresReplace (TFIN-632) and TestCTEClientGroupGuardPointResource_noOpFieldsRequireReplace (TFIN-633), both passing. The TFIN-628/TFIN-544 batchKey fix is not covered by a new acceptance test: CM's GET /guardpoints response does not return intelligent_protection, data_classification_enabled, or data_lineage_enabled at all (confirmed live), so no state-based TestCheckResourceAttr assertion can distinguish correct batching from a collapsed batch -- verified instead via live-CM POST-count evidence (1 call before the fix, 2 after) captured in /opt/tfin_bugs/TFIN-628-544-632-633-cte-guardpoint/.

### DIFF
--- a/internal/provider/cte/resource_cte_client_guardpoints.go
+++ b/internal/provider/cte/resource_cte_client_guardpoints.go
@@ -133,7 +133,11 @@ func (r *resourceCTEClientGP) Schema(_ context.Context, _ resource.SchemaRequest
 								},
 								"early_access": schema.BoolAttribute{
 									Optional:    true,
-									Description: "Whether secure start is turned on.",
+									Description: "Whether secure start is turned on. Changing this value for an EXISTING guard_path is applied in place via a dedicated CM endpoint -- it does not force a replace.",
+									// CM has a dedicated early-access endpoint that supports both
+									// true->false and false->true in place for an existing
+									// GuardPoint, so no plan modifier is needed. Update() calls
+									// that dedicated endpoint directly on a genuine change.
 								},
 								"intelligent_protection": schema.BoolAttribute{
 									Optional:    true,
@@ -145,6 +149,8 @@ func (r *resourceCTEClientGP) Schema(_ context.Context, _ resource.SchemaRequest
 								},
 								"mfa_enabled": schema.BoolAttribute{
 									Optional:    true,
+									Computed:    true,
+									Default:     booldefault.StaticBool(false),
 									Description: "Whether MFA is enabled.",
 								},
 								"network_share_credentials_id": schema.StringAttribute{
@@ -153,13 +159,30 @@ func (r *resourceCTEClientGP) Schema(_ context.Context, _ resource.SchemaRequest
 								},
 								"preserve_sparse_regions": schema.BoolAttribute{
 									Optional:    true,
-									Description: "Whether to preserve sparse file regions.",
+									Description: "Whether to preserve sparse file regions. CM has a dedicated endpoint that turns this off in place for an EXISTING guard_path (true -> false does not force a replace), but once turned off it can never be turned back on for that same GuardPoint via any API call -- changing it from false to true for an EXISTING guard_path forces a whole-resource replace.",
+									PlanModifiers: []planmodifier.Bool{
+										// See resource_cte_clientgroup_guardpoints.go's
+										// preserve_sparse_regions for the full rationale -- same
+										// treatment applied here for consistency between the two
+										// resources.
+										modifiers.BoolRequiresReplaceOnFalseToTrueUnlessNewMapEntry(),
+									},
 								},
 								"guard_enabled": schema.BoolAttribute{
 									Optional:    true,
 									Computed:    true,
 									Default:     booldefault.StaticBool(true),
-									Description: "Whether the GuardPoint is enabled.",
+									Description: "Whether the GuardPoint is enabled. Changing this value for an EXISTING guard_path is applied in place via a dedicated CM endpoint -- it does not force a replace.",
+									// PR review follow-up: guard_enabled is create-time-inert on CM (a
+									// GuardPoint created with guard_enabled = false still comes
+									// up enabled) and the generic PATCH .../guardpoints/{id}
+									// also silently no-ops it. Both Create() and Update() now
+									// call CM's dedicated .../guardpoints/enable endpoint
+									// instead, confirmed live to work in both directions
+									// (true->false and false->true) for an existing GuardPoint.
+									// Since Update() actually applies the change either way (same
+									// reasoning as early_access above), no plan modifier is
+									// needed here.
 								},
 							},
 						},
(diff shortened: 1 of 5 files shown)


internal/provider/cte/resource_cte_clientgroup_guardpoints.go
zubeensyed, line 101: This is still plain `RequiresReplace()` — the same bug TFIN-634 fixed in the sibling resource (83aa15d). Terraform can't distinguish a new map key from an existing entry changing, so adding a guard path still triggers a full destroy/recreate.

That means:

1. TFIN-634's destructive replace-on-add issue remains, including its no-rollback risk.
2. This likely drives TFIN-628: add a path → full replace → `Create()` recreates both paths → the sibling's converged `guard_enabled = false` is lost. It also means this PR's `Update()` batchKey changes are never reached for a pure add.

Please switch to `modifiers.RequiresReplaceUnlessNewMapEntry()`, consistent with the reasoning in this PR's new `policy_id` comment. As-is, the new descriptions claiming that adding a new `guard_path` doesn't force replacement are incorrect and will also be reflected in `docs/`.

internal/provider/cte/resource_cte_clientgroup_guardpoints.go
```diff
@@ -153,7 +212,11 @@ func (r *resourceCTEClientGroupGP) Schema(_ context.Context, _ resource.SchemaRe
 								},
 								"preserve_sparse_regions": schema.BoolAttribute{
 									Optional:    true,
-									Description: "Whether to preserve sparse file regions.",
+									Description: "Whether to preserve sparse file regions. Changing this value for an EXISTING guard_path forces the whole resource to be replaced, since CM silently no-ops an update to this field via PATCH.",
+									PlanModifiers: []planmodifier.Bool{
+										// TFIN-633: see automount_enabled above.
+										modifiers.BoolRequiresReplaceUnlessNewMapEntry(),
+									},
```
zubeensyed, line 243: These ten attributes are `Optional` only — no `Computed`, no `Default` so on an existing guard_path where the user never set the field, `StateValue` is null, and the modifier's `req.StateValue.IsNull()` early-return fires. 
No replace is planned, the in-place update goes out, CM no-ops it, and TFIN-633 reproduces unchanged.

internal/provider/cte/resource_cte_clientgroup_guardpoints.go
```diff
@@ -153,7 +212,11 @@ func (r *resourceCTEClientGroupGP) Schema(_ context.Context, _ resource.SchemaRe
 								},
 								"preserve_sparse_regions": schema.BoolAttribute{
```
zubeensyed, line 229: CM supports changing both of these in place, so RequiresReplace here destroys every guard point in the resource for a change that doesn't need it:

- `PATCH .../guardpoints/{id}/early-access` (swagger-cte.yaml:11049) enables *and* disables early access.
- `PATCH .../guardpoints/{id}/preserve-sparse-regions-off` (:11156) handles true → false. false → true is genuinely impossible per CM's docs ("if you disable the setting, it cannot be enabled again", :8752), so a replace is right for that direction only.

TFIN-633's "CM silently no-ops it" finding came from the generic PATCH, which simply isn't the endpoint for these two fields. Suggest calling the dedicated endpoints from Update(), and keeping RequiresReplace only for preserve_sparse_regions false → true.

internal/provider/cte/resource_cte_client_guardpoints.go
zubeensyed, line 764: TFIN-633's follow-up comment about `mfa_enabled` on this resource isn't addressed here: this payload sets only `IsGuardEnabled` and `NWShareCredentialsID`, so `mfa_enabled` is never sent, unlike the clientgroup version of the same function, which does send it.

CM's generic PATCH does accept `mfa_enabled` (swagger-cte.yaml:9339-9350), so the fix here is to send the field, not to force a replace.

**Suggestion**: `mfa_enabled` is Optional-only on this resource but Optional+Computed+Default(false) on the clientgroup sibling is worth aligning.

internal/provider/cte/resource_cte_clientgroup_guardpoints.go
manishkumar-thales: Added the fix.

internal/provider/cte/resource_cte_clientgroup_guardpoints.go
manishkumar-thales: Added the fix.

internal/provider/cte/resource_cte_client_guardpoints.go
manishkumar-thales: Added the fix.

internal/provider/cte/resource_cte_clientgroup_guardpoints.go
manishkumar-thales: Added the fix.

zubeensyed (approved):

PR #537, opened 2026-08-15
docs: correct and restructure README

zubeensyed:
The README documented provider configuration that no longer matches the code, in ways that break users who follow it.

Corrections:
- Environment variables are CIPHERTRUST_*, not CM_*. The documented CM_ADDRESS / CM_USERNAME / CM_PASSWORD / CM_DOMAIN / CM_AUTH_DOMAIN are read nowhere in the provider, so the documented export block resolved no configuration at all.
- no_ssl_verify was documented inverted and with the wrong default. Since 1.0.1 it defaults to false, meaning TLS verification is on; setting it true disables verification and warns at plan time.
- rest_api_timeout defaults to 180, not 60.
- All provider attributes are Optional in the schema; address, username and password are enforced at runtime, and only outside bootstrap mode.
- Cloud key management covers AWS and OCI. Azure and GCP have connection resources only, so listing them as supported CCKM clouds overstated what ships.

Additions:
- ca_cert, log_file and log_level, none of which were documented. ca_cert is the migration path for the secure-by-default TLS change in 1.0.1.
- Defaults for aws_operation_timeout, oci_operation_timeout and replication_delay_ms.
- The remaining environment variables: BOOTSTRAP, NO_SSL_VERIFY, CIPHERTRUST_CA_CERT, REST_API_TIMEOUT, CIPHERTRUST_REPLICATION_DELAY.
- The ~/.ciphertrust/config constraints the provider enforces: the path must not be a symlink and must not be group- or world-accessible, either of which is a hard error.
- The resources unavailable on CDSPaaS.
- Requirements, install, and developing sections, and links to docs/, examples/, sample-scripts/, changelog.md, SECURITY.md and LICENSE.

Also drops the tfplugindocs frontmatter and duplicated schema dump. The README is hand-maintained and was drifting from the generated docs/index.md; the schema section now points there instead.


README.md
```diff
@@ -1,97 +1,155 @@
...
-# generated by https://github.com/hashicorp/terraform-plugin-docs
-page_title: "CipherTrust Provider"
-subcategory: ""
-description: |-
-  The CipherTrust provider can be used configure a CipherTrust instance or cluster or a CipherTrust Data Security Platform as a Service (CDSPaaS) and subsequently manage cloud resources.
----
+# Terraform Provider for Thales CipherTrust
 
-# CipherTrust Provider
+The CipherTrust provider configures a CipherTrust Manager (CM) instance or cluster, or a
```
SarahThompson: This sentence just reads a little awkwardly - maybe 2 sentences or 
The CipherTrust provider configures a CipherTrust Manager (CM) instance or cluster, or a CipherTrust Data Security Platform as a Service (CDSPaaS) tenant, and manages cloud key resources protected by these platforms.

README.md
```diff
@@ -1,97 +1,155 @@
...
+CipherTrust Data Security Platform as a Service (CDSPaaS) tenant, and manages the cloud key
+resources those platforms protect.
 
-### AWS Deployment
-To deploy a Virtual CipherTrust Manager from AWS, you must supply the Amazon Machine Image (AMI), available on the AWS Marketplace or through the Thales Cloud Provisioning System.
+- **Registry documentation:** https://registry.terraform.io/providers/ThalesGroup/CipherTrust/latest/docs
+- **Resource and data source reference:** [`docs/`](docs/) — [resources](docs/resources/), [data sources](docs/data-sources/)
+- **Runnable examples:** [`examples/`](examples/) and [`sample-scripts/`](sample-scripts/)
```
SarahThompson, line 11: Sample scripts have not had a look at in years . they used to work but wouldn't know now.

README.md
```diff
@@ -1,97 +1,155 @@
...
+- **Release notes:** [`changelog.md`](changelog.md)
 
-### Azure Deployment
-1. List the available versions with `Get-AzVMImage -location eastus2 -PublisherName thalesdiscplusainc1596561677238 -Offer cm_k170v -sku ciphertrust_manager`.
+## Requirements
 
-2. Obtain image information for a particular version with `az vm image show --location eastus2 --urn thalesdiscplusainc1596561677238:cm_k170v:ciphertrust_manager:<desired-version>`. Under `plan`, obtain the required values for `name`, `product` and `publisher`.
+| | |
+|:--|:--|
+| Terraform | >= 1.0 (provider protocol 6.0) |
```
SarahThompson: When I was trying to find the minimum value of the provider that worked for all our rsources it wasn't 1.0

README.md
```diff
@@ -102,92 +160,163 @@ provider "ciphertrust" {
...
-## Configuration File
+`tenant` accepts a tenant name (`"acme"`) or a tenant path (`"acme/eng/team"`).
 
-All provider parameters can be read from the configuration file.
+`auth_domain` is still accepted for on-prem domain routing, but it is ignored when `tenant` is set —
+the CDSPaaS path derives `auth_domain_path` from `tenant`, which supersedes it. Configuring both
+produces a warning at plan time.
 
-The configuration file is ~/.ciphertrust/config. For example:
+Some resources manage infrastructure that CDSPaaS operates on your behalf and are unavailable there.
```
SarahThompson: I don't know what this sentences means after reading 5 times!  Oh I see it's related to below list - still doesn't read well.  Maybe start with "The following" .. probably looks better when it's not a diff.

README.md
```diff
@@ -102,92 +160,163 @@ provider "ciphertrust" {
...
 
 **Example:**
-- Created resource with: `max_connections = 100`
-- Edited config to: `max_connections = "invalid"`
+
+- Created resource with `max_connections = 100`
+- Edited config to `max_connections = "invalid"`
 - Run `terraform destroy` → validation fails
-- Must revert config to valid value, then destroy works
+- Must revert config to a valid value, then destroy works
```
SarahThompson: "destroy works" ? maybe "destroy will succeed"

README.md
```diff
@@ -102,92 +160,163 @@ provider "ciphertrust" {
...
+
+**Workaround:** before destroying, ensure all attributes in your configuration have valid values,
+even if they differ from the actual resource state. This is common across Terraform providers that
+define attribute validators, so it is worth keeping in mind generally.
```
SarahThompson: Not sure why we wouldn't mention terraform -refresh-false?

README.md
```diff
@@ -102,92 +160,163 @@ provider "ciphertrust" {
...
+
+## Developing the provider
+
+```bash
+make build      # compile ./terraform-provider-ciphertrust
+make install    # go install ./...
+make fmt        # gofmt -s -w
+make lint       # golangci-lint run
```
SarahThompson: Isn't this going to throw up 1000's of issues?
I just tried to run again and now got unsupported config file error again. 
I only ran last week successfully so not sure what's changed but it no longer used by later versions of golangci-lint

README.md
```diff
@@ -102,92 +160,163 @@ provider "ciphertrust" {
...
+```
+
+Acceptance tests create real resources on a real CipherTrust Manager. They run with `TF_ACC=1` and
+read the target from the same `CIPHERTRUST_*` environment variables listed above.
+
+To try a locally built provider, point Terraform at the binary with a
+[development override](https://developer.hashicorp.com/terraform/cli/config/config-file#development-overrides-for-provider-developers)
+in your `.terraformrc`.
+
+Documentation under `docs/` is generated from the provider schema plus the files in `examples/` and
```
SarahThompson: imp-templates - it was renamed from 'templates' to prevent tfplugindocs from reading the directory.  It's not part of our docs.

README.md
zubeensyed: Updated.

README.md
zubeensyed: Updated.

README.md
zubeensyed: Removed for now.

README.md
zubeensyed: Removed for now.

